PR #107, opened 2019-12-18
feat(actions): bind gh to mousehover

kwonoj:
It's quite frequent usecases to use mousehover to see compiler error messages / or other information provided like below:

image

This PR attempts to mimic vscodevim's keybinding, bringing `gh` to server as mouse hover in vscode.

Honestly not sure if this is desired changes (and as vim beginner unsure if there's predefined bindings), open to suggestions.

p.s: bumd q, is it possible user can configure this actions by own instead of upstream?

### DIFF
--- a/vim/vscode-code-actions.vim
+++ b/vim/vscode-code-actions.vim
@@ -34,6 +34,11 @@ function! s:vscodeGoToDefinition()
     endif
 endfunction
 
+function! s:hover()
+  normal! gv
+  call VSCodeNotify('editor.action.showHover')
+endfunction
+
 command! -range -bar VSCodeCommentary call s:vscodeCommentary(<line1>, <line2>)
 
 xnoremap <expr> <Plug>VSCodeCommentary <SID>vscodeCommentary()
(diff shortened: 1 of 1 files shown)


asvetliakov:
Thank you for the PR! ```gh``` is related to the select mode but i think we can sacrifice it for now (it doesn't work now anyway and i don't intent to do it unless someone will ask specially)
I'd also add ```gH``` binding to ```editor.action.triggerParameterHints```

>p.s: bumd q, is it possible user can configure this actions by own instead of upstream?

Yes, put the same into your init.vim

vim/vscode-code-actions.vim
```diff
@@ -48,12 +48,14 @@ nnoremap <expr> == <SID>vscodeFormat() . '_'
...
 nnoremap <silent> gf :<C-u>call <SID>vscodeGoToDefinition()<CR>
 nnoremap <silent> gd :<C-u>call <SID>vscodeGoToDefinition()<CR>
+nnoremap <silent> gh :<C-u>call VSCodeNotify('editor.action.showHover')<CR>
 nnoremap <silent> <C-]> :<C-u>call <SID>vscodeGoToDefinition()<CR>
 nnoremap <silent> gF :<C-u>call VSCodeNotify('editor.action.peekDefinition')<CR>
 nnoremap <silent> gD :<C-u>call VSCodeNotify('editor.action.peekDefinition')<CR>
 
 xnoremap <silent> gf :<C-u>call <SID>vscodeGoToDefinition()<CR>
 xnoremap <silent> gd :<C-u>call <SID>vscodeGoToDefinition()<CR>
+xnoremap <silent> gh :<C-u>call VSCodeNotify('editor.action.showHover')<CR>
```
asvetliakov: This is not correct mapping - entering command line will exit visual mode which we probably don't want to do.
The correct one is to call `normal! gv` following by `VSCodeNotify()`, i.e.

```
function! s:hover()
  normal! gv
  call VSCodeNotify('editor.action.showHover)
endfunction

xnoremap <silent> gh :<C-u>call <SID>hover()<CR>
```

justinmk:
why not use `K` for this?

asvetliakov:
yes, good idea

PR #160, opened 2020-01-26
[WIP] Use different method to send edits in insert mode

### LINKED ISSUE #76: [vim-repeat] insert mode
Hello. 

Given the following **invalid** JSON (it's missing the EOL commas `,`):

```json
{
  "line1": "1"
  "line2": "2"
  "line3": "3"
  "line4": "4"
}
```

In Vim with the [vim-repeat](https://github.com/tpope/vim-repeat) plugin, if I type the following sequence on `"line1"` to insert the comma at the end of the line: 

`A,ESC` 

Then I move down to `"line2"` and type the `.` and then down to `"line3"` and type the `.` again it will repeat and insert the comma at the end of each of those lines. It doesn't seem like this behavior is currently working in vscode-neovim. 

Other vim-repeat behaviors including those coupled with [vim-surround](https://github.com/tpope/vim-surround) like `cs"'` work fine. I would assume the issue is related to switching over to insert mode and VSCode taking back control?

I've got 🤞 the solution doesn't require listening to events in insert mode as from a few other issues that sounds like a non-starter due to performance reasons.

Thanks for the hard work on this project!

asvetliakov:
*Not ready for daily use yet*

Experimental workaround for dot-repeat operation. Instead `nvim_buf_set_lines` it tries to set cursor position and call `nvim_input` with delete & text keystokes

- Doesn't work with multi-byte characters now
- Text deletion is performed only by `<BS>` keys. This matters for dot-repeat operator (i.e. delete left or right) but no way to workaround this.

Will provide soon vsix file for testing

Fixes #76

### DIFF
--- a/package.json
+++ b/package.json
@@ -400,37 +400,37 @@
 			{
 				"command": "vscode-neovim.send",
 				"key": "backspace",
-				"when": "editorTextFocus && neovim.init && neovim.mode != insert || neovim.recording",
+				"when": "editorTextFocus && neovim.init && neovim.mode != insert",
 				"args": "<BS>"
 			},
 			{
 				"command": "vscode-neovim.send",
 				"key": "shift+backspace",
-				"when": "editorTextFocus && neovim.init && neovim.mode != insert || neovim.recording",
+				"when": "editorTextFocus && neovim.init && neovim.mode != insert",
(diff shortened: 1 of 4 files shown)


asvetliakov:
Too clunky, need another way

PR #173, opened 2020-02-10
Workaround dot-repeat by replaying last changes in separate buffer

### LINKED ISSUE #76: [vim-repeat] insert mode

asvetliakov:
This one is better than PR in #160 

This is utilizing the fact that `.` command repeats last stuff globally across all buffers & windows. So we tricking neovim by creating separate window & buffer and replaying last subsequent changes here by using `nvim_input` after exiting insert mode. After that this temporary buffer is being destroyed

Fixes #76

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -60,10 +60,6 @@ interface RequestResponse {
     send(resp: unknown, isError?: boolean): void;
 }
 
-interface ActiveChange {
-    delLength: number;
-}
-
 // to not deal with screenrow positioning, we set height to high value and scrolloff to value / 2. so screenrow will be always constant
 // big scrolloff is needed to make sure that editor visible space will be always within virtual vim boundaries, regardless of current
 // cursor positioning
@@ -2306,7 +2302,6 @@ export class NVIMPluginController implements vscode.Disposable {
         if (!this.isInit) {
             return;
         }
-        // need <Esc> first otherwise it won't allow to switch the buffer
         if (this.isInsertMode) {
             this.leaveMultipleCursorsForVisualMode = false;
             await this.uploadDocumentChangesToNeovim();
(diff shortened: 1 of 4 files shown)


asvetliakov:
Test version for anyone willing to try (rename `zip` to `vsix` and install it through `Install from VSIX` in vscode):

[vscode-neovim-0.0.51.zip](https://github.com/asvetliakov/vscode-neovim/files/4178625/vscode-neovim-0.0.51.zip)

fsoft72:
the unreleased vscode-neovim-0.0.51 zip works on my system

ngharrison:
I tried the version posted in this thread. It works in some cases (changing a word, appending to end of line), but not in others.
Repeating the edit of `oword` acts like a paste of a new line and the word wherever your cursor is.
Repeating the edit of `inew word` would sometimes work and sometimes only insert the last part of it. I think it was the space and the second word, but I couldn't get it consistent.
Repeating the edit of `A (2)` replaces the last character in the line with ')'. I believe this has to do with vs code auto-completing the close parenthesis.

asvetliakov:
@ngharrison These are good and valid concerns, thank you

PR #184, opened 2020-02-12
Convert to AsciiDoc in order to add TOC to README

trkoch:
The current `README.md` is comprehensive, but hard to navigate because it misses a table of contents. As discussed [here](https://github.com/isaacs/github/issues/215#issue-36227252), GitHub does not allow automatic table of contents generation for Markdown formatted files. One suggested workaround is to convert to [AsciiDoc](http://asciidoc.org/userguide.html#_introduction).

I'm not sure if its worth the effort (learning yet another syntax) but IMHO the TOC looks quite nice and is pretty useful.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -1,0 +1,557 @@
+= Neo Vim (VS Code Neovim)
+
+Neovim integration for Visual Studio Code
+
+For those who don’t know https://neovim.io/[Neovim] is the fork of VIM
+to allow greater VIM extensibility and embeddability. The extension is
+using full embedded neovim instance as backend (with the exception of
+the insert mode and window/buffer/file management), no more
+half-complete VIM emulation
+
+VSCodeVim has neovim integration but it doesn’t use it for anything but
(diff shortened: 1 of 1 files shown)


trkoch:
See https://github.com/trkoch/vscode-neovim/tree/asciidoc-readme for a preview.

davidovich:
if a change in format is necessary, Github also supports ReStructuredText `.rst` which also has toc support. An example: https://github.com/mozilla/sops

davidovich:
Alternate conversions can be explored with [pandoc](https://pandoc.org/).

For example
```
pandoc --from gfm --to rst
```

Will take the existing markdown and convert it to rst (automatically).

trkoch:
Comparing `rst` and `asciidoc` [here](http://hyperpolyglot.org/lightweight-markup) I can see why you chose `rst`. Also, directly converting from `gfm` (as opposed to `markdown`) may solve some issues I had with automatic conversion. In any case, a TOC will be very much appreciated.

PR #219, opened 2020-03-10
Fix tabs being inserted after replacement

### LINKED ISSUE #155: s/foo/bar inserts tab after replacement

jasonrhansen:
Fixes #155


asvetliakov:
Thank you for the PR ! Superseded by #334 which is taking another method - instead using `<Tab>` we're sending `<C-e>` character which doesn't mess global or substitute command anymore

PR #221, opened 2020-03-10
Fix lag when cycling through command history

jasonrhansen:
When in command line mode, pressing up and down to go through history
would lag enough to make it annoying when pressing up or down multiple
times quickly. Calling show() on the QuickPick after setting the new value
fixes this.

### DIFF
--- a/src/command_line.ts
+++ b/src/command_line.ts
@@ -214,6 +214,7 @@ export class CommandLineController implements Disposable {
         const res = await this.neovimClient.callFunction("getcmdline", []);
         if (res) {
             this.input.value = res;
+            this.input.show();
         }
     };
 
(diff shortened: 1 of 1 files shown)


jrolfs:
Command history breaks altogether for me in the latest vscode update (1.45). Since you're presumably familiar with this code/feature... any idea what might be going on there?

jasonrhansen:
@jrolfs Sorry, I was only trying out vscode for a short time to see if it's something I wanted to use long term. There were too many little issues with the various "vim" plugins that I decided that, at least for the time being, it wasn't going to work for me. I'm just using straight up neovim now.

asvetliakov:
Thank you for you PR!

PR #237, opened 2020-04-15
Dot-repeat and u are messed up with inInsertMode

azizghuloum:
Greetings,

I was just experimenting with disabling all the things that the extension does to handle insert mode, and pass everything directly to neovim.  `c` commands, `i`, etc.,  now all work much better and dot-repeat seems to be more civilized.  

I'll be test running these changes for a few days to see what breaks in my day-to-day usage.  No need to merge as of yet.  Thanks!


azizghuloum:
Auto complete is broken :-(

PR #238, opened 2020-04-15
fix jk config example

jedwards1211:
key should be j in this case, not k right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ to enable `jk` add also:
 ```json
     {
         "command": "vscode-neovim.compositeEscape2",
-        "key": "k",
+        "key": "j",
         "when": "neovim.mode == insert",
         "args": "k"
     }


shawntax:
No, that's incorrect. Have you tried changing it to a `j` and seeing what happens?

The `firstKey` and `secondKey` naming is a bit confusing. If you peek at the [handleCompositeEscape](https://github.com/asvetliakov/vscode-neovim/blob/master/src/controller.ts#L2425) functions, you'll notice that `handleCompositeEscapeSecondKey` depends on a timestamp that is only set by `handleCompositeEscapeFirstKey`. In other words, the key for `compositeEscape2` only registers if the key for `compositeEscape1` was pressed first. That's why it's noted you cannot bind `jk` without binding `jj` first. 

Also this is why if you bind `compositeEscape2` to `j`, you get some weird behavior with VSCode running both handler functions. The timestamp condition never gets met, and so VSCode prints "k" (your arg) when you press "j" (the key)!

jedwards1211:
okay huh.  I just naively assumed this was a typo

PR #280, opened 2020-06-07
Add option to draw easymotion on top

### LINKED ISSUE #276: EasyMotion shifting text
My favorite things about easy motion is that you don't need to move your gaze away from where you want to go. The text decorations seem to break that though since they get placed behind the text and shift everything on the line over. When there are a lot of matches on a given line, it can move the characters significantly.

Is there any way to avoid replacing the text in the buffer like the original easymotion did while also keeping the text in it's original positions?

jhgarner:
The original easymotion extension drew text directly on top of the
target characters. The fork drew text behind the target characters.
This commit adds an option to pick whichever you want.
Even if you pick on top of the target, you won't end up with broken
linting.

closes #276


jhgarner:
I think there are some inconsistencies with motions at the ends of lines. I'll try to fix those and give a comment when that's done.

jhgarner:
I think this is finally working with almost perfect accuracy. The code is a bit more complicated, but it makes using the plugin significantly nicer. If I don't catch any problems in the next few days, I'll remove the WIP tag.

theol0403:
It seems to work great. One thing I noticed is that the text overlay is *slightly* too far to the left. 

image
image

This causes the shadow of the gutter to occasionally overlap the text:
image

jhgarner:
I wonder if that has to do with the magic number of 8 that's used for the margin and width. Maybe there's some way to dynamically pick the right thing? I'll look into that.

jhgarner:
@theol0403 I've switched the units from px to ch. I didn't see the problem on my machine so let me know if that fixed it for you.

theol0403:
Yep, that fixed it. It now works beautifully on my machine :+1:

jhgarner:
I've removed the WIP from the title. It seems to work pretty well.

Shatur95:
The feature I missing. Offsetting lines makes easymotion very awkward to use. But I see that @asvetliakov is busy now. Maybe it would be better to let some right to the contributors, such as @jhgarner?

asvetliakov:
Will look/review the PR within next week

Shatur95:
@jhgarner, could you resolve merge conflicts?

package.json
```diff
@@ -220,6 +220,11 @@
 					"title": "Unknown Highlight Groups",
 					"description": "Highlight Configuration for non defined group. Can accept 'vim' to use colors coming from vim or text decoration properties object",
 					"default": "vim"
+				},
+				"vscode-neovim.easymotionOnTop": {
```
asvetliakov, line 224: let's rename to `textDecorationsAtTop`

package.json
```diff
@@ -1020,6 +1025,7 @@
 		"prettier": "^2.0.5",
 		"source-map-support": "^0.5.16",
 		"typescript": "^3.7.4",
+		"vsce": "^1.79.3",
```
asvetliakov, line 1028: not needed here

src/controller.ts
```diff
@@ -1910,26 +1910,65 @@ export class NVIMPluginController implements vscode.Disposable {
         }
```
asvetliakov, line 1910: This should be in `highlight_manager.ts`

src/controller.ts
```diff
@@ -1910,26 +1910,65 @@ export class NVIMPluginController implements vscode.Disposable {
...
-                    // const col = colNum - 1;
-                    const col = Utils.convertByteNumToCharNum(line, colNum - 1);
-                    const opt: vscode.DecorationOptions = {
-                        range: new vscode.Range(lineNum, col, lineNum, col),
-                        renderOptions: {
-                            before: {
-                                ...conf,
-                                ...conf.before,
-                                contentText: text,
+                    if (Utils.getEasymotionOnTop()) {
```
asvetliakov, line 1921: Pass an option to HL manager constructor. Getting workspace configuration every time is costly

src/controller.ts
```diff
@@ -1910,26 +1910,65 @@ export class NVIMPluginController implements vscode.Disposable {
...
+                        // 8 seems to be a magic number. It works for me, but I'm not sure why...
+                        const width = text.length;
+                        // vim sends column in bytes, need to convert to characters
+                        const col = Utils.convertByteNumToCharNumOnTop(line, colNum);
```
asvetliakov, line 1925: No need for another function - `convertByteNumToCharNum()` is used now only for decorations. Also check how does it work with tabs

src/controller.ts
```diff
@@ -1910,26 +1910,65 @@ export class NVIMPluginController implements vscode.Disposable {
...
-                    const col = Utils.convertByteNumToCharNum(line, colNum - 1);
-                    const opt: vscode.DecorationOptions = {
-                        range: new vscode.Range(lineNum, col, lineNum, col),
-                        renderOptions: {
-                            before: {
-                                ...conf,
-                                ...conf.before,
-                                contentText: text,
+                    if (Utils.getEasymotionOnTop()) {
+                        // 8 seems to be a magic number. It works for me, but I'm not sure why...
```
asvetliakov, line 1922: Not sure i understand this comment

src/controller.ts
```diff
@@ -1910,26 +1910,65 @@ export class NVIMPluginController implements vscode.Disposable {
...
+                            );
+                        } else {
+                            const opt: vscode.DecorationOptions = {
+                                range: new vscode.Range(lineNum, col + text.length - 1, lineNum, col + text.length - 1),
+                                renderOptions: {
+                                    // Inspired by https://github.com/VSCodeVim/Vim/blob/badecf1b7ecd239e3ed58720245b6f4a74e439b7/src/actions/plugins/easymotion/easymotion.ts#L64
+                                    after: {
+                                        // If we try to draw off the end of the screen, VSCode will move the tet to the left.
+                                        // Each move VSCode makes to the left for us is one more we don't want to do.
+                                        margin: `0 0 0 -${Math.min(
```
asvetliakov, line 1944: What will happen if we won't use a negative right margin? Mind a share a screenshot?

src/utils.ts
```diff
@@ -277,6 +277,19 @@ export function convertByteNumToCharNum(line: string, col: number): number {
     return currCharNum;
 }
 
+export function convertByteNumToCharNumOnTop(line: string, col: number): number {
+    let totalBytes = 0;
+    let currCharNum = 0;
+    while (totalBytes < col) {
+        if (currCharNum >= line.length) {
+            return currCharNum + (col - totalBytes);
```
asvetliakov, line 285: Why this is needed? Anyway, better to modify the existing function rather than creating a copy with slighly changed behavior

asvetliakov (changes requested):

Shatur95:
Maybe it would be easier to just open new PR since master have a lot of changes?

asvetliakov:
>Maybe it would be easier to just open new PR since master have a lot of changes?

Agree

jhgarner:
Thanks for the review. I'll close this issue, refactor it for the new master, and fix the comments you brought up.

jrolfs:
↓ Can't wait to get rid of this! ↓
vscode-neovim--absolute-text-decorations


If anyone's looking for a workaround in the meantime though...

```json
  "customizeUI.stylesheet": {
    ".monaco-editor span[class*=' ced-1-TextEditorDecorationType']:not([class*=' ced-1-TextEditorDecorationType1'])::before": "position: absolute;"
  }
```

PR #281, opened 2020-06-07
Change highlight groups configuration schema to accept "vim" string

ian-h-chamberlain:
The default configuration is considered to be invalid based on the configuration schema. This PR updates the schema to make the default config valid and adds some extra validation to disallow incorrect string values.

This appears to align with the configuration values declared in the configuration's TS type, and the documentation. Also slightly reworded description to make it more clear.

Before:
Screen Shot 2020-06-07 at 4 21 43 PM

After:
Screen Shot 2020-06-07 at 4 14 08 PM


asvetliakov:
Thank you for your PR!

PR #292, opened 2020-06-30
Escape keys should only work in editorTextFocus mode

David-Else:
In reference to https://github.com/asvetliakov/vscode-neovim/issues/290


asvetliakov:
Thank you!

Issue #236, opened 2020-04-11
<c-e> and <c-y> don't work


kuator:
<c-e> and <c-y> don't work properly. Cursor is left behind, and because of this the view jumps back to it.

PR #305, opened 2020-07-31
fix: cursor is left behind

kuator:
https://stackoverflow.com/questions/49997089/vscode-how-to-set-that-when-i-press-ctrl-up-or-down-arrow-the-cursor-move-wi
Vs Code, by default, doesn't update cursor position after scrolling, which is not the case in vim. This pr makes experience more similar to scrolling in vim.

### DIFF
--- a/src/commands_controller.ts
+++ b/src/commands_controller.ts
@@ -2,6 +2,7 @@ import vscode, { Disposable } from "vscode";
 import { NeovimClient } from "neovim";
 
 import { NeovimExtensionRequestProcessable } from "./neovim_events_processable";
+import { EXT_NAME } from "./utils";
 
 export class CommandsController implements Disposable, NeovimExtensionRequestProcessable {
     private client: NeovimClient;
@@ -108,7 +109,9 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
     };
 
     private scrollLine = (to: "up" | "down"): void => {
-        vscode.commands.executeCommand("editorScroll", { to, by: "line", revealCursor: false });
+        const settings = vscode.workspace.getConfiguration(EXT_NAME);
+        const revealCursorScrollLine = settings.get("revealCursorScrollLine");
+        vscode.commands.executeCommand("editorScroll", { to, by: "line", revealCursor: revealCursorScrollLine });
     };
 
     private goToLine = (to: "top" | "middle" | "bottom"): void => {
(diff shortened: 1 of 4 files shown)


asvetliakov:
probably need a setting here

kuator:
Kind of like this?

src/commands_controller.ts
```diff
@@ -108,7 +109,9 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
     };
 
     private scrollLine = (to: "up" | "down"): void => {
-        vscode.commands.executeCommand("editorScroll", { to, by: "line", revealCursor: false });
+        const settings = vscode.workspace.getConfiguration(EXT_NAME);
```
asvetliakov, line 112: Pass the option as constructor param from the main_controller, don't query it every time for perf reasons

asvetliakov (changes requested):

kuator:
Is it better?

src/main_controller.ts
```diff
@@ -173,7 +174,22 @@ export class MainController implements vscode.Disposable {
         const channel = await this.client.channelId;
         await this.client.setVar("vscode_channel", channel);
 
-        this.commandsController = new CommandsController(this.client);
+        this.logger.debug(`${LOG_PREFIX}: UIAttach`);
+        await this.client.uiAttach(this.NEOVIM_WIN_WIDTH, this.NEOVIM_WIN_HEIGHT, {
```
asvetliakov, line 178: Broken merge ? remove

asvetliakov (changes requested):
Almost 😄

asvetliakov (approved):

asvetliakov:
Fixes #236

asvetliakov:
Thank you for your PR!

d4h0:
@kuator: Thanks a million! This problem was one of the main reason why I stopped using this extension in the past.

@asvetliakov: Is there a reason why this can not be activated by default?

And a million thanks to you as well for creating the extension – without the now fixed cursor problem, it's fantastic!

kuator:
> @kuator: Thanks a million! This problem was one of the main reason why I stopped using this extension in the past.
> 
> @asvetliakov: Is there a reason why this can not be activated by default?
> 
> And a million thanks to you as well for creating the extension – without the now fixed cursor problem, it's fantastic!

😢thanks

Issue #264, opened 2020-05-12
tabs indentation


anothergituser:
There's an issue with the cursor width when the file has tabs for indentation, where the cursor is as wide as the tab stop, no problems when using spaces for indentation. Is this a known issue or someone can corroborate?

Thank you.

legion151:
Same for me - if you have tabs and jump one word to right - cursor position is completely broken and editing happens in wrong places.

PR #307, opened 2020-08-04
Fixing tabs

### LINKED ISSUE #239: cursor is in the wrong place in tab-indent file
1. 4-space-indent file,ujhk move cursor right.
2. after change all space to tab,ujhk movement is chaos,can't reach the head of line with ^ or h,can't search the right words under the cursor, etc...

### LINKED ISSUE #275: tab width counted as 1 in visual block mode
```
abcdefghijklmn
    123456
```
The second line starts with a tab of width 4.
Now move the cursor to `e`, press `ctrl-v`, press `j`.
The cursor jumps to '4' instead of '1'.
Screenshot_2020-05-28_00-30-13

jhgarner:
This fixes #275 by matching the tabwidth that vscode and vim use. I think this also fixes #239 and #264 and any other issues about cursor movement in tabbed files.


jhgarner:
It looks like I accidentally included the easymotion changes here as well. If that's a problem, I'll look into what I need to do to fix the history.

theol0403:
I haven't verified for sure that this PR caused it, but after installing this PR, highlighting seems really messed up. I don't have a screenshot to show simply because the highlighting is barely visible (in visual-line for example). 

If you aren't having the same problem, I can try to investigate and provide more information.

jhgarner:
Thanks for pointing that out. I tested highlighting with visual block mode but managed to miss that the other visual modes had gotten messed up. I'll investigate why and update this code.

theol0403:
Cool, glad it's not just me =).
It seems to not just be the other visual modes, it's anything that requires highlighting of a whole line or more. For example, I have the vim-highlightedyank plugin installed, and it does not work when yanking anything over one line.

jhgarner:
Any highlight that tried to go beyond the end of a line was broken because I thought indexing an array past the end would return undefined, not null. Hopefully that's fixed now.

theol0403:
Sweet! Just tried it out and it works great. This whole PR has the added benefit that it highlights regions more like nvim, where it won't highlight empty lines and only highlights to the end of the line.

Thanks a ton for making these PRs, it really makes this plugin even more enjoyable to use :slightly_smiling_face: .

Shatur95:
Thank you, switched to this branch currently.

Shatur95:
I'm not sure if this is related to the PR, but do you faced with line desync? Sometimes I use `dd` and this removes previous line instead of current.

jhgarner:
I haven't seen any line desyncs. Are there any files and/or particular actions that lead to it?

Shatur95:
I would like to provide steps to reproduce, but this happens randomly :(
Maybe this because of plugins, but I use [only basic](https://github.com/Shatur95/neovim-config/tree/master/pack/plugins/start).
So, the plugin works perfectly for you?

jhgarner:
No, I haven't seen that issue although there's a lot of internal state and it's very likely I've missed some edge cases. Do the desyncs only happen in files with hard tabs?

Shatur95:
Sorry for a long reply. This was hard to reproduce, but now I have 100% reproducible steps:

```vim
dd
<Up>
u
<Up>
dd
u
<Down>
dd
```

On last step you will remove upper line instead of current. But I seems to not related to the PR, the issue reproducible on latest master too for me. Can you reproduce it on your side?

jhgarner:
Thanks! I just checked and I can also reproduce the highlighting issue following your steps. As for the desync, It doesn't seem to happen for me. I even followed the install instructions for the link you shared above with your config and it still worked as expected. Any chance you can share the file you tested this in? I've run into one desync randomly while editing a file so far which seemed similar to what you're describing but after refreshing, I couldn't reproduce it.

Shatur95:
> Thanks! I just checked and I can also reproduce the highlighting issue following your steps.

Thanks for the checking!

> It doesn't seem to happen for me. I even followed the install instructions for the link you shared above with your config and it still worked as expected. Any chance you can share the file you tested this in? I've run into one desync randomly while editing a file so far which seemed similar to what you're describing but after refreshing, I couldn't reproduce it.

Sorry, I just double checked and found that I provided a bit wrong steps. The following should reproduce the bug:

```vim
dd
gk
u
gk
dd
u
gj
dd
```

I discovered it because I have mapped `j` to `gj`, `k` to `jk` (for navigation between soft-wrapped lines)

jhgarner:
With those instructions I can reproduce it with a blank nvim config. It seems like it only happens on longer lines which makes me think something is going wrong with text wrapping. Since you said it's broken on master as well, it might be worth opening your own issue for this.

Shatur95:
Thanks! Just opened #312. I would try to fix this problem, but I'm not familiar with TypeScript :(
So, I hope that someone will be interested in fixing it.

asvetliakov:
Superseded by #334. Thank you for the PR!

PR #309, opened 2020-08-07
Use case-sensetive comparsion

### LINKED ISSUE #308: Wrong string comparsions
I found that `set ignorecase` causes weird behavior. This is because all string comparsions in Vimscript part done by just `==`. We need to use `==#` to always compare case-sensitive.
For example, with this settings in vimrc `Ctrl+Shift+P` makes incorrect visual mode conversion [this condition is always true](https://github.com/asvetliakov/vscode-neovim/blob/54e46eac62a9c537bad55073293763ff434982f1/vim/vscode-insert.vim#L12).

Shatur95:
Vimscript uses settings-dependent case comparsion by default. Use case-sensetive comparsion to avoid issues caused by `set ignorecase`. Closes #308.

### DIFF
--- a/vim/vscode-code-actions.vim
+++ b/vim/vscode-code-actions.vim
@@ -42,7 +42,7 @@ endfunction
 function! s:openVSCodeCommandsInVisualMode()
     normal! gv
     let visualmode = visualmode()
-    if visualmode == "V"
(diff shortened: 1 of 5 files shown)


asvetliakov:
Thank you for PR

Issue #291, opened 2020-06-29
Remove file path from status bar


SpyTec:
Having the file path in the status bar is a bit much, is it possible to remove this somehow?

iMarv:
Right click the status bar and uncheck Neo Vim.
You won't have a statusbar indicator for the mode your are in though.

SpyTec:
Yep, know about that part. Though as you say, will miss out on everything else

Shatur95:
> set shortmess+=F

Setting will not works since extension hardcodes [shortmess](https://github.com/asvetliakov/vscode-neovim/blob/8b78a7edd7dcee928db9fb036553852c238e8c1a/vim/vscode-options.vim#L3).

The option already exist in `shortmess`, this just not works because you do not need to set it as a string.

asvetliakov:
@Shatur95 You seems better than me in vimscript 😄  care to send a PR ?

Shatur95:
Sure :)

PR #318, opened 2020-08-23
Fix shortmess

### LINKED ISSUE #230: How to Hide Statusbar Messages Execpt Mode Name
Statusbar messages is long and noise, only mode name is useful, can I hide other messages

WeChatf3309b982f9703959d2642b70ae699bd

Shatur95:
References #291 and closes #230.

### DIFF
--- a/vim/vscode-options.vim
+++ b/vim/vscode-options.vim
@@ -1,6 +1,6 @@
 " This file used to force set neovim options which may break the extension. Loaded after user config
 
-set shortmess="filnxtToOFI"
+set shortmess=filnxtToOFI
 set nowrap
 set mouse=a
 set cmdheight=1


asvetliakov:
Awesome! Thank you

kar9222:
Hi @Shatur95, thanks for this PR. This PR changed `set shortmess="filnxtToOFI"` to  `set shortmess=filnxtToOFI`

Personally, I do use `set shortmess="filnxtToOFI"`. There is no way I can set it back even with my init.vim because 
> This file used to force set neovim options which may break the extension. Loaded after user config. 

 Is there any way not to hardcode this in __vim/vscode-options.vim__ ? I am happy to do a PR to make it configurable if you can provide some hints on how to make it configurable?

Shatur95:
You could try to override it in `after`. Many options just do not work correctly with the extension. That's why they were hard-coded.

PR #322, opened 2020-08-24
Set default Linux path

Shatur95:
In this PR I set default Neovim path to `/usr/bin/nvim` for Linux. I think this can be reasonable default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,9 +57,9 @@
 				},
 				"vscode-neovim.neovimExecutablePaths.linux": {
 					"type": "string",
-					"default": "",
+					"default": "/usr/bin/nvim",
 					"title": "Neovim executable path on Linux (and in WSL)",
-					"markdownDescription": "Full path to Neovim executable that should be used by the extension if running VS Code on Linux or WSL. If `useWSL` setting is checked, vscode-neovim will look for this path in WSL  \nExample:  \n/usr/bin/nvim"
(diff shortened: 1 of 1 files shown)


Shatur95:
I also noticed that just `nvim` as a path works for me. Is it expected?

Shatur95:
Oh, I think I got why you not set this path as default. If the path is wrong, then the user will not receive a message that they need to specify the path to Neovim, right?

asvetliakov:
>I also noticed that just nvim as a path works for me. Is it expected?

Looks like `spawn()` uses system path variable. But extension host may override path, not load the one from a user profile so it's easier to just tell `specify full path`

>Oh, I think I got why you not set this path as default. If the path is wrong, then the user will not receive a message that they need to specify the path to Neovim, right?

Correct. However this can be fixed, move https://github.com/asvetliakov/vscode-neovim/blob/f55e2374864b6b62f890a170ad3cddc481661639/src/controller.ts#L304 to top of the constructor (so we won't bind anything if there is an error), check spawn result and throw error if not spawned. Catch in `extension.ts` and display error message

Shatur95:
> Correct. However this can be fixed

Sorry, but I do not have any knowledge in TS :(
But I like the idea. This will notify user if Neovim executable does not exist.

> Looks like spawn() uses system path variable.

What if try to execute nvim from `PATH` if the previous check if failed?

Shatur95:
I will close this PR and open a new one.

Issue #325, opened 2020-08-25
Proposal: Logo


ngscheurich:
As I [mentioned](https://github.com/asvetliakov/vscode-neovim/issues/319#issuecomment-680289048) in #319, I propose changing the name of this extension to something that stands alone, i.e. not _Neo Vim_ or _Neovim_, and developing a logo for the project. I designed a logo for my suggested name—_NeoCode_—and I thought I'd share it to get some feedback. The idea was to re-create the VS Code logo while borrowing some design elements from the Neovim logo.

I understand that this name and/or design may not be something that the community is into, but I wanted to share in case it is helpful.

Thoughts?

![neocode-logo](https://user-images.githubusercontent.com/423798/91239009-f2a46680-e703-11ea-901c-f3d650bc2381.png)

asvetliakov:
Awesome work, love this one. Probably need a square one for marketplace

ngscheurich:
Glad you like it! I submitted a PR, #327,  with a slightly revised version in case you want to go ahead and use it.

David-Else:
This looks amazing! The only problem I can see is that the word 'vim' is no longer in the title. Many people who are looking for Vim shortcuts may not know about Neovim and not get the reference. Maybe if there was a strong tag line used everywhere the logo appeared this could reduce the problem? How about this?
```
NeoCode - Code at the speed of thought | Native Neovim integration for Visual Studio Code.
```

ngscheurich:
That's a really good point, @David-Else. I think using "Native Neovim integration for Visual Studio Code." as the `description` property in `package.json` is a great idea, regardless of what the name ends up being.

ngscheurich:
If we do end up using the moniker I suggested, we should keep the "c" lower case—_Neocode_—to better match up with _Neovim._ Just a thought.

ian-h-chamberlain:
Found this issue and just wanted to add another perspective to the conversation:

If you Google "Neocode" the first result appears to be a software company by that name: https://www.neocode.com/

I have no legal expertise so don't know if there is any problem with using the same name. The company seems different enough from this project that there would be little opportunity for confusion – but I wouldn't want to see this project to suffer because of a name collision like that.

That said, I love the logo and styling and think it complements the VSCode and Neovim designs quite nicely. So if there is no issue with the name collision I think it should still be used.

theol0403:
[...]

PR #327, opened 2020-08-26
Add extension icon

ngscheurich:
This adds the extension icon proposed in #325, with a few small modifications.

icon
> The icon as submitted in this PR.


ian-h-chamberlain:
Slightly related since this is an image PR... I noticed when I checked out the repo recently: the `.gitattributes` in this repo has 
```
* text eol=lf
```
Which I believe will attempt to convert line endings in files (in my case I saw a GIF getting changed before commit). It might be prudent to add something like
```
*.png binary
*.gif binary
```
or even 
```
images/** binary
```
based on [this github example](https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings), to avoid different contributors from accidentally changing line endings in image files.

asvetliakov:
Thank you for your awesome work!

Issue #108, opened 2019-12-19
bind K and gO to vscode actions


asvetliakov:

aminroosta:
@asvetliakov This plugin is awesome 🚀 

I found out that i could call any vscode command from my vimrc.
Even with arguments!
I think the above should be mentioned in the readme.

```vim
nnoremap K  :<C-u>call VSCodeCall('editor.action.showHover')<CR>
nnoremap ? :<C-u>call VSCodeCall('workbench.action.findInFiles', { 'query': expand('<cword>')})<CR>
```

asvetliakov:
@aminroosta 
From readme 😄 

>It's possible to call vscode commands from neovim. See VSCodeCall/VSCodeNotify vim functions in vscode-neovim.vim file. VSCodeCall is blocking request, while VSCodeNotify is not

Probably need to move it to own section for better visibility.

Also there are also `VSCodeCallRange`/`VSCodeNotifyRange` in case if you need to specify some visual range for vscode command

Generally use `VSCodeNotify` unless you're really need blocking call

PR #339, opened 2020-09-06
Improve code actions

theol0403:
Having both gd and gf mapped to "revealDefinition" seemed redundant, so this PR maps gf to "revealDeclaration". In some languages like C++ this makes a difference, and the mnemonic "go to forward declaration" makes sense for gf. gd remains "go to definition".

I refactored the `vscodeGoToDefinition` function to support these two modes without needing to duplicate the function.

I also provided a binding for "revealReferences", and since "hover" is gh, I figured gH is a decent choice.

Finally, I'm not sure why `<C-w>gF` was mapped to the same thing as `<C-w>gf`, I know it was meant to mirror neovim but there are so many other missing keybindings that I don't think this little one is worth the space it takes.

With neovim now working inside peek, this functionality has suddenly become significantly more useful :grin:

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,12 @@ Neovim 0.4.2 or greater
 -   It's possible to call vscode commands from neovim. See `VSCodeCall/VSCodeNotify` vim functions in `vscode-neovim.vim` file. `VSCodeCall` is blocking request, while `VSCodeNotify` is not (see below)
 -   Scrolling is done by VSCode side. `<C-d>/<C-u>/etc...` are slighly different
 -   File management commands such as `e` / `w` / `q` etc are mapped to corresponding vscode commands and behavior may be different (see below)
--   `gf`/`gd`/`<C-]` are mapped to `editor.action.revealDefinition` (Shortcut `F12`). also `<C-]>` works in vim helps files
--   `gF`/`gD` are mapped to `editor.action.peekDefinition` (opens definition in peek)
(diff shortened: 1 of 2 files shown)


asvetliakov (approved):

asvetliakov:
Thank you for your PR!

Shatur95:
> since "hover" is gh,

BTW, why the default is `gh`? Usually this is `K`.

asvetliakov:
This is coming from #107 PR. Primary for a users coming from vscodevim. 

There is also old issue #108 to add `K` and `gO` (can be mapped to focus outline view / `goToSymbol`, also create jumppoint before sending request to vscode). PRs are welcome here! 😄

Shatur95:
I will send a PR if you agree :)
Should I change `gh` to `K` or add it?

asvetliakov:
Add second `K` for hover, don't remove `gh`

Shatur95:
@theol0403, is `gH` works for you?

theol0403:
Yes it does. Make sure that "go to references" is something that your LSP server for your chosen language supports.

Shatur95:
theol0403, Strange, in my case it just does the same that gH should do...

Shatur95:
Never mind, it was configuration problem. I deleted my configuration and it work. Sorry for bothering.

Shatur95:
The problem was in [vim-xkbswitch](https://github.com/lyokha/vim-xkbswitch). The following solved my issue:

```vim
let g:XkbSwitchSkipGhKeys = ['gh', 'gH']
```

Maybe someone will be interested.

PR #356, opened 2020-09-08
Fix marks jumping

### LINKED ISSUE #228: Uppercase marks not working
_Example to reproduce:_
Set an uppercase mark, e.g. **mD**.
Open another editor.
Jump back to the mark by pressing **`D**

_What happens vs what is expected:_
Cursor jumps to the correct position (i.e. marked line number and column) but in the current file.
It should jump to that position the original file instead.

Shatur95:
Closes #228.

### DIFF
--- a/vim/vscode-jumplist.vim
+++ b/vim/vscode-jumplist.vim
@@ -29,7 +29,7 @@ augroup VscodeJumplist
 augroup END
 
 
-onoremap <C-o> <Cmd>call <SID>jumpToMark()<CR>
+onoremap ` <Cmd>call <SID>jumpToMark()<CR>
 nnoremap <silent> <C-o> :<C-u>call <SID>jump(0)<CR>
 nnoremap <silent> <C-i> :<C-u>call <SID>jump(1)<CR>
 nnoremap <silent> <Tab> :<C-u>call <SID>jump(1)<CR>
--- a/vim/vscode-jumplist.vim
+++ b/vim/vscode-jumplist.vim
@@ -4,13 +4,19 @@
 " nnoremap <silent> <C-i> :<C-u>call VSCodeNotify("workbench.action.navigateForward")<CR>
 " nnoremap <silent> <Tab> :<C-u>call VSCodeNotify("workbench.action.navigateForward")<CR>
 
-function s:jump(forward)
+function! s:jumpToMark() abort
+    let g:isJumping = 1
+    normal! `
+    let g:isJumping = 0
+endfunction
+
+function! s:jump(forward) abort
     let g:isJumping = 1
     exe a:forward ? "normal! 1\<C-i>" : "normal! \<C-o>"
     let g:isJumping = 0
 endfunction
 
-function! s:clearJumplist()
+function! s:clearJumplist() abort
     if exists("w:vscode_clearjumps") && w:vscode_clearjumps
         let w:vscode_clearjumps = 0
         clearjumps
@@ -23,6 +29,7 @@ augroup VscodeJumplist
 augroup END
 
 
+onoremap <C-o> <Cmd>call <SID>jumpToMark()<CR>
 nnoremap <silent> <C-o> :<C-u>call <SID>jump(0)<CR>
 nnoremap <silent> <C-i> :<C-u>call <SID>jump(1)<CR>
 nnoremap <silent> <Tab> :<C-u>call <SID>jump(1)<CR>


Shatur95:
BTW, I noticed that you use something like

```vim
nnoremap <silent> <C-o> :<C-u>call <SID>jump(0)<CR>
```

This can be done in more convenient syntax:

```vim
nnoremap <C-o> <Cmd>call <SID>jump(0)<CR>
```

Let me know if you like the idea, I will send a PR.

asvetliakov:
>`nnoremap <C-o> <Cmd>call <SID>jump(0)<CR>`

I like, but `<Cmd>` has slightly different semantics. Be careful when changing it

vim/vscode-jumplist.vim
```diff
@@ -23,6 +29,7 @@ augroup VscodeJumplist
 augroup END
 
 
+onoremap <C-o> <Cmd>call <SID>jumpToMark()<CR>
```
asvetliakov, line 32: Why `<C-o>` here? What about jumping with ``` `A ``` and `'A` ?

vim/vscode-jumplist.vim
```diff
@@ -4,13 +4,19 @@
 " nnoremap <silent> <C-i> :<C-u>call VSCodeNotify("workbench.action.navigateForward")<CR>
 " nnoremap <silent> <Tab> :<C-u>call VSCodeNotify("workbench.action.navigateForward")<CR>
 
-function s:jump(forward)
+function! s:jumpToMark() abort
+    let g:isJumping = 1
+    normal! `
```
asvetliakov, line 9: what about `'A` ?

asvetliakov (changes requested):

vim/vscode-jumplist.vim
Shatur95: Omg, sorry, just a stupid typo after merge. Fixed.

Shatur95:
Hm... The shortcut is working, but `let g:isJumping = 1` and `let g:isJumping = 0` does not have effect.

Shatur95:
Do <C-o> and <C-i> work for you for jumping between files?
All this shortcut do not jump between files for me.

asvetliakov:
>Do `<C-o>` and `<C-i>` work for you for jumping between files?

Yes

I just tried PR, it doesn't work. But changing these to:

```
function! s:jumpToMark() abort
    let g:isJumping = 1
    normal! `A
    let g:isJumping = 0
endfunction

nnoremap `A <Cmd>call <SID>jumpToMark()<CR>
```

then `mA` in first file, then ``` `A``` in other file correctly jumps. So i guess we shouldn't do operator remap here and use ```nnoremap``` in a loop

Shatur95:
> Yes

This is strange. What I have tried:

0. Remove all my Neovim settings an plugins.
1. Open this branch in VSCode.
2. Launch the extension (another VSCode instance opened)
3. Open any file. Then open another file in split.
4. Jump to the end, then jump to another side of split, then jump to the beggining.
5. Press `<C-o>` - jumping occurs only inside single file.

> I just tried PR, it doesn't work.

Of course. But for me any jumping do not works between two files.
Do I understand correctly that ` from this PR works, it just does not move between files (e.g. it does not fixes the problem)?
I guess you have `<C-i>` and `<C-o>` reassigned to `workbench.action.navigateBack` and `workbench.action.navigateForward` in VSCode settings. Could you check it?

asvetliakov:
>Open any file. Then open another file in split.
>Jump to the end, then jump to another side of split, then jump to the beggining.

Jumplist is each own separate por editor column and not being inherited by split commands (see readme) 😄

Shatur95:
> Jumplist is each own separate por editor column and not being inherited by split commands (see readme) smile

Still not works for me. What I tried

1. Opened any file.
2. Open another file (now I have two tabs)
3. `<C-o>` - nothing happened.
4. `G` - jumped to the end
5. `<C-o>` - jumped to the previous position
6. `<C-o>` - nothing happened.
7. Swith to the previous tab.
6. `<C-o>` - nothing happened.

I also updated this branch to the latest master just to be sure. Also this was done without any settings.

Shatur95:
Oh, got it. The problem in `workbench.editor.enablePreview` set to `true`. Because of it `<C-o>` and `<C-i>` not works even if I keeping tabs in editor.

If `workbench.editor.enablePreview` is set to `true` it works. But it also have issues. Try the following:

0. Close all tabs.
1. Open first file.
2. Open second file.
3. `<C-o>`. It jumps to first file.
4. Open third file
3. `<C-o>`. It jumps to first file.
3. `<C-o>`. **It jumps to first file.**
This hotkey shouldn't have done that.

Another issue I noticed:

1. Open a lot of files.
2. Close them quickly except one.
3. Quickly press `<C-o>` a lot of time.
You will have several Untitled buffers with content insted of opened files. It happens sometimes, but often. Try 1-3 steps two-three times and you should reproduce it.

Is it possible to improve jumping system? I think this should be done before fixing marks.

asvetliakov:
>You will have several Untitled buffers with content insted of opened files. It happens sometimes, but often. Try 1-3 steps two-three times and you should reproduce it.

I noticed this problem too, for some reason neovim sends `/file/path` instead `file:///` uri when jumping and vscode ignores it. Will try to solve it by either fixing neovim or by accepting paths too

>6. <C-o>. It jumps to first file.
>This hotkey shouldn't have done that.

There are limitations with jumping. Opening a buffer doesn't create a jumppoint. Not easy to solve

Shatur95:
> There are limitations with jumping. Opening a buffer doesn't create a jumppoint. Not easy to solve

Maybe use VSCode jumping for `<C-o>` / `<C-i>`? Yes, it works a bit differently, but do not have issues.

asvetliakov:
>Maybe use VSCode jumping for <C-o> / <C-i>? Yes, it works a bit differently, but do not have issues.

Unfortunately vscode jumping is completely broken because setting vscode cursor is creating vscode jumppoint for each cursor move

Shatur95:
> Unfortunately vscode jumping is completely broken because setting vscode cursor is creating vscode jumppoint for each cursor move

Yes :(
But I use this in my vimrc for now to avoid current jumping issues.

asvetliakov:
Probably can be workarounded by using vscode's `cursorMove` command. But need to count character shifting then between old & new cursor pos

asvetliakov:
>`cursorMove`

Nope. `cursorMove` vscode command stops when reaching a new line and ignoring given `value` count. Possible workaround here too but too clunky. Need to fix vim jumplist

Shatur95:
Closed after #461.

PR #358, opened 2020-09-09
Text decorations at top

jhgarner:
This pull requests adds an option to draw text decorations from things like EasyMotion on top of the text they're decorating instead of behind the text. It's nearly identical to the old one but based on the rewrite.


Shatur95:
I tested, it works fine, but the setting applied only after restart. Is it okay? Usually VSCode extension apply their setting on change.

asvetliakov (approved):

asvetliakov:
>I tested, it works fine, but the setting applied only after restart. Is it okay? I understand that this is because the setting is passed to the constructor, but usually VSCode extension apply their setting on change.

Yep, that's ok. We don't want to risk performance by querying settings each time (They're probably cached in the extension host, but better to be safe than not)

asvetliakov:
@jhgarner Thank you for your PR!

PR #362, opened 2020-09-12
Refactor Vimscript

theol0403:
I believe this was suggested by @Shatur95 in another issue.

A command exists called `<Cmd>` that replaces the need for `:<C-U>` and `gv` in mappings that call commands.

In brief: (`help :map-cmd`),
> The <Cmd> pseudokey begins a "command mapping", which executes the command
directly (without changing modes).
This is more flexible than `:<C-U>` in visual and operator-pending mode, or
`<C-O>:` in insert-mode, because the commands are executed directly in the
current mode (instead of always going to normal-mode).  Visual-mode is
preserved, so tricks with |gv| are not needed.  Commands can be invoked
directly in cmdline-mode (which would otherwise require timer hacks).

I did a search and replace for this over the whole project, as well as removed all the unnecessary `gv` commands and even entirely removed some helper functions.

Finally, there is this note in the help txt:
> Because <Cmd> avoids mode-changes (unlike ":") it does not trigger CmdlineEnter and CmdlineLeave events. This helps performance.

When testing out this PR, I've noticed an improvement in editor responsiveness and speed!

### DIFF
--- a/vim/vscode-code-actions.vim
+++ b/vim/vscode-code-actions.vim
@@ -8,7 +8,7 @@ function! s:vscodeFormat(...) abort
         let [line1, line2] = [line("'["), line("']")]
     endif
 
-    call VSCodeCallRange("editor.action.formatSelection", line1, line2, 0)
+    call VSCodeCallRange('editor.action.formatSelection', line1, line2, 0)
 endfunction
 
 function! s:vscodeCommentary(...) abort
@@ -21,13 +21,13 @@ function! s:vscodeCommentary(...) abort
         let [line1, line2] = [line("'["), line("']")]
     endif
 
-    call VSCodeCallRange("editor.action.commentLine", line1, line2, 0)
+    call VSCodeCallRange('editor.action.commentLine', line1, line2, 0)
 endfunction
 
 function! s:vscodeGoToDefinition(str)
     if exists('b:vscode_controlled') && b:vscode_controlled
         exe "normal! m'"
-        call VSCodeNotify("editor.action.reveal" . a:str)
+        call VSCodeNotify('editor.action.reveal' . a:str)
     else
         " Allow to funcionar in help files
         exe "normal! \<C-]>"
@@ -41,14 +41,14 @@ endfunction
 
 function! s:openVSCodeCommandsInVisualMode()
     let mode = mode()
-    if mode ==# "V"
-        let startLine = line("v")
-        let endLine = line(".")
-        call VSCodeNotifyRange("workbench.action.showCommands", startLine, endLine, 1)
+    if mode ==# 'V'
+        let startLine = line('v')
+        let endLine = line('.')
+        call VSCodeNotifyRange('workbench.action.showCommands', startLine, endLine, 1)
     else
-        let startPos = getpos("v")
-        let endPos = getpos(".")
-        call VSCodeNotifyRangePos("workbench.action.showCommands", startPos[1], endPos[1], startPos[2], endPos[2] + 1, 1)
+        let startPos = getpos('v')
+        let endPos = getpos('.')
+        call VSCodeNotifyRangePos('workbench.action.showCommands', startPos[1], endPos[1], startPos[2], endPos[2] + 1, 1)
     endif
 endfunction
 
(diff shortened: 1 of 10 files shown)


vim/vscode-insert.vim
```diff
@@ -25,7 +25,7 @@ augroup MultipleCursors
...
 
 " Multiple cursors support for visual line/block modes
-xnoremap <silent> ma :<C-u>call <SID>vscodePrepareMultipleCursors(1, 1)<CR>
-xnoremap <silent> mi :<C-u>call <SID>vscodePrepareMultipleCursors(0, 1)<CR>
-xnoremap <silent> mA :<C-u>call <SID>vscodePrepareMultipleCursors(1, 0)<CR>
-xnoremap <silent> mI :<C-u>call <SID>vscodePrepareMultipleCursors(0, 0)<CR>
+xnoremap <silent> ma <Cmd>call <SID>vscodePrepareMultipleCursors(1, 1)<CR>
+xnoremap <silent> mi <Cmd>call <SID>vscodePrepareMultipleCursors(0, 1)<CR>
+xnoremap <silent> mA <Cmd>call <SID>vscodePrepareMultipleCursors(1, 0)<CR>
+xnoremap <silent> mI <Cmd>call <SID>vscodePrepareMultipleCursors(0, 0)<CR>
```
Shatur95, line 31: Tests fails for multiply cursors. You should probably edit `vscodePrepareMultipleCursors` fuction.

Shatur95 (changes requested):
Yes, thanks!

I tested, works correctly for me.

vim/vscode-insert.vim
theol0403: Humm, so it looks like staying in visual mode causes the problem. I am not very experienced with vimscript, do you have any suggestions on how to exit visual mode, other than keeping the `:<C-U>`?

Shatur95: This is because `s:vscodePrepareMultipleCursors` relies on `call feedkeys('i', 'n')`. But sending `i` will not produce insert mode from visual, you need to send `<Esc>` first. So, try to change `call feedkeys('i', 'n')` to `call feedkeys("\<Esc>i", 'n')` (you should use double quotes in this case).

theol0403: Yeah, you are right about the cause of the issue. However, the change you suggested didn't work. I tried a few combinations but I was not able to get it working, any other suggestions?

Shatur95: Okay, let's go into insert mode first. Just call `i<Esc>` inside this mapping first and remove `feedkeys` from the function.

Shatur95: Like `xnoremap <silent> ma <Esc>i<Cmd>call <SID>vscodePrepareMultipleCursors(1, 1)<CR>.

theol0403: Humm, so I got it to kind of work by leaving `call feedkeys('i', 'n')` (just `('i')` does the same thing too), and doing
```
xnoremap <silent> ma <Esc><Cmd>call <SID>vscodePrepareMultipleCursors(1, 1)<CR>
```
However, it is only working the first time and rarely after. I think there is timing issues going on with the cursors being spawned and going into insert mode.

Shatur95: Could you try `xnoremap <silent> ma <Esc>i<Cmd>call <SID>vscodePrepareMultipleCursors(1, 1)<CR>` without `call feedkeys('i', 'n')`?

theol0403: I did, but that won't work, because `vscodePrepareMultipleCursors` needs to set some global flags *before* the insert mode autocommand is triggered.

Shatur95: Oh, yes, you are right.
What if call it via `normal!`?

theol0403: I've tried a lot of things and none of them work all the time, even when I revert this PR.
I think this was caused by refactor, so I think we need to call on @asvetliakov . I think the cursor gets moved by neovim while entering insert mode, which steps on the vscode cursor spawning which happens before.

Shatur95: Have you tried deleting the config? Multiply cursors work for me on master.

Shatur95: Just tried solution with `feedkeys` and it also works. Let's check it on CI.

vim/vscode-code-actions.vim
```diff
@@ -70,39 +64,39 @@ nnoremap <expr> = <SID>vscodeFormat()
...
 
 " Workaround for gk/gj
-nnoremap gk :<C-u>call VSCodeCall('cursorMove', { 'to': 'up', 'by': 'wrappedLine', 'value': v:count ? v:count : 1 })<CR>
-nnoremap gj :<C-u>call VSCodeCall('cursorMove', { 'to': 'down', 'by': 'wrappedLine', 'value': v:count ? v:count : 1 })<CR>
+nnoremap gk <Cmd>call VSCodeCall('cursorMove', { 'to': 'up', 'by': 'wrappedLine', 'value': v:count ? v:count : 1 })<CR>
+nnoremap gj <Cmd>call VSCodeCall('cursorMove', { 'to': 'down', 'by': 'wrappedLine', 'value': v:count ? v:count : 1 })<CR>
 
 " workaround for calling command picker in visual mode
-xnoremap <silent> <C-P> :<C-u>call <SID>openVSCodeCommandsInVisualMode()<CR>
+xnoremap <silent> <C-P> <Cmd>call <SID>openVSCodeCommandsInVisualMode()<CR>
```
Shatur95, line 102: Now this command fail tests.

Shatur95 (changes requested):

vim/vscode-code-actions.vim
theol0403: I have no idea what causes it. Either there is something wrong with the test or it just randomly failed due to timing changes. Maybe putting `wait()` commands in the test will fix it.

Shatur95: Yes, I don't know either. The command works correctly for me.

Shatur95: I found solution! The problem was in `visualmode()`.

Shatur95:
Maybe also remove `<silent>` from mappings? It no longer necessary.

vim/vscode-insert.vim
Shatur95: I figured it out. Fixed :)

Shatur95:
Now LGTM.

Shatur95 (approved):

asvetliakov (approved):
Awesome!

Shatur95:
I will solve merge conflicts, give me a minute.

asvetliakov:
Thank you @Shatur95 @theol0403

PR #382, opened 2020-09-21
Check for insert mode before dispose

Shatur95:


src/typing_manager.ts
```diff
@@ -72,7 +72,7 @@ export class TypingManager implements Disposable {
                 );
                 this.changeManager.getDocumentChangeCompletionLock(editor.document)?.then(() => {
                     this.isEnteringInsertMode = false;
-                    if (this.typeHandlerDisposable) {
+                    if (this.typeHandlerDisposable && this.modeManager.isInsertMode) {
                         this.logger.debug(`${LOG_PREFIX}: Waiting done, disposing type handler`);
                         this.typeHandlerDisposable.dispose();
```
asvetliakov, line 77: better to wrap these lines into a check instead because you might lose the pressed keys otherwise

Or move `if(this.pendingKeysAfterEnter)` check outside of this block

asvetliakov (changes requested):
Need a test also

src/typing_manager.ts
Shatur95: Thank you, I just wanted to ask about this :)

Shatur95: Done.

src/typing_manager.ts
```diff
@@ -72,14 +72,14 @@ export class TypingManager implements Disposable {
...
                         this.typeHandlerDisposable = undefined;
-                        if (this.pendingKeysAfterEnter) {
-                            commands.executeCommand("default:type", { text: this.pendingKeysAfterEnter });
-                            this.pendingKeysAfterEnter = "";
-                        }
+                    }
+                    if (this.pendingKeysAfterEnter) {
+                        commands.executeCommand("default:type", { text: this.pendingKeysAfterEnter });
```
asvetliakov, line 81: I forgot - need one more thing here, executing `default:type` may insert text in normal mode which is wrong,

Change to:
`commands.executeCommand(this.modeManager.isInsertMode ? "default:type" : "type", ....)`

asvetliakov (changes requested):
I'll write a test later

src/typing_manager.ts
Shatur95: Thanks!

Shatur95:
@asvetliakov, the issue was reproduced by plugin, how test would look to check this?

asvetliakov:
Go without test, i'm lack of the time again last months

Shatur95:
Okay, could you merge it? Can't merge without approve and can't approve my own PR :)

asvetliakov (approved):

PR #396, opened 2020-10-05
Map :x to Wq

cornedor:
I often use :x to quickly save and close a file, but I noticed it was not mapped and caused issues where the extensions stopped working when a another file was opened.

xa is already implemented, so this is just for the single file save/close

### DIFF
--- a/vim/vscode-file-commands.vim
+++ b/vim/vscode-file-commands.vim
@@ -38,6 +38,7 @@ command! -bang Wall call VSCodeNotify('workbench.action.files.saveAll')
 command! -bang Quit if <q-bang> ==# '!' | call VSCodeNotify('workbench.action.revertAndCloseActiveEditor') | else | call VSCodeNotify('workbench.action.closeActiveEditor') | endif
 
 command! -bang Wq call <SID>saveAndClose()
+command! -bang Xit call <SID>saveAndClose()
 
 command! -bang Qall call VSCodeNotify('workbench.action.closeAllEditors')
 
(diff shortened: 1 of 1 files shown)


vim/vscode-file-commands.vim
```diff
@@ -53,6 +53,7 @@ AlterCommand sav[eas] Saveas
 AlterCommand wa[ll] Wall
 AlterCommand q[uit] Quit
 AlterCommand wq Wq
+AlterCommand x Wq
```
Shatur95, line 56: But you mapped to `Wq`. Maybe it should be mapped to `Xit`? You should create this mapping.

Shatur95 (changes requested):

vim/vscode-file-commands.vim
cornedor: Good suggestion

vim/vscode-file-commands.vim
```diff
@@ -53,7 +54,7 @@ AlterCommand sav[eas] Saveas
 AlterCommand wa[ll] Wall
 AlterCommand q[uit] Quit
 AlterCommand wq Wq
-AlterCommand x Wq
+AlterCommand x Xit
```
Shatur95, line 57:
```suggestion
AlterCommand x[it] Xit
```

Shatur95 (changes requested):
Last minor change :)

Shatur95 (approved):

PR #397, opened 2020-10-07
Disable shada

Shatur95:
This PR uncomments `set shada=` option. [Similar](https://github.com/VSCodeVim/Vim/pull/3288) approach used in VSCodeVim but with command-line arguments. Without this `v:oldfiles` will be filled with useless `file://...`.


Shatur95:
Sorry, I forgot that this is also responsible not only for the file history. And there is no way to disable only file history (I only can hardcode the whole `shada` option). Need to think about it.

PR #398, opened 2020-10-09
Ignore WSL option on non windows OS

### LINKED ISSUE #381: error message re WSL - on linux
I got this error message but I am not sure how to fix it in a way that does not break my windows installation. I need to untick the "Use WSL" setting (vscode-neovim.useWSL) for it to work on Linux, but I need to tick it for it to work on Windows. This is a bit of a headache for me because I use cloud synchronization of my vscode settings.

    2020-09-21T21:27:28.189Z MainController: Neovim spawn error: spawn C:\Windows\system32\wsl.exe ENOENT

ObliviousJamie:
Using the node.js variable process.platform the platform node is running on can be found. If it isn't windows based then the wsl setting is set to false. 
Manually tested on windows 10 and a debian based OS. 

closes #381

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -21,7 +21,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     const mouseVisualSelection = settings.get("mouseSelectionStartVisualMode", false);
     const useCtrlKeysNormalMode = settings.get("useCtrlKeysForNormalMode", true);
     const useCtrlKeysInsertMode = settings.get("useCtrlKeysForInsertMode", true);
-    const useWsl = settings.get("useWSL", false) && isWindows;
+    const useWsl = isWindows && settings.get("useWSL", false);
     const revealCursorScrollLine = settings.get("revealCursorScrollLine", false);
     const neovimWidth = settings.get("neovimWidth", 1000);
     const customInit = getNeovimInitPath() ?? "";
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -13,13 +13,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
         vscode.window.showErrorMessage("Neovim: configure the path to neovim and restart the editor");
         return;
     }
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -13,13 +13,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+    const isWindows = process.platform == "win32";
+
     const highlightConfIgnore = settings.get("highlightGroups.ignoreHighlights");
     const highlightConfHighlights = settings.get("highlightGroups.highlights");
     const highlightConfUnknown = settings.get("highlightGroups.unknownHighlight");
     const mouseVisualSelection = settings.get("mouseSelectionStartVisualMode", false);
     const useCtrlKeysNormalMode = settings.get("useCtrlKeysForNormalMode", true);
     const useCtrlKeysInsertMode = settings.get("useCtrlKeysForInsertMode", true);
-    const useWsl = settings.get("useWSL", false);
+    const useWsl = settings.get("useWSL", false) && isWindows;
```
Shatur95, line 24: I'm not very good at JavaScript, but shouldn't we put in a cheaper check first?

Shatur95 (changes requested):

src/extension.ts
ObliviousJamie: Good point. I've swapped these now.

Shatur95:
Thanks! Could anyone test it?

Shatur95:
@hholst80, could you test?

hholst80:
Sure, I will do that.

Shatur95:
Just get it from CI in the [artifacts](https://github.com/asvetliakov/vscode-neovim/actions/runs/362055492) section :)

willnorris:
Just tested on Windows and Mac, and works as expected as far as I can tell... Windows is using neovim from WSL and Mac is using the locally installed version, even with `useWSL: true`.

Shatur95 (approved):

Shatur95:
Thanks!

PR #402, opened 2020-10-10
Use column position when using Ctrl-A in insert

### LINKED ISSUE #283: i_Ctrl-A not inserting the correct text
Amazing work on this plugin; really promising!

There seems to be a bug with i_Ctrl-A currently; I'm not sure exactly what's going on.

---

Peek 2020-06-09 09-37

In this first example I type

```
iText 1<Esc>oText 2<Esc>o<Ctrl-A><Esc>
```

I'd expect it to insert `Text 2` but it inserts `Text 1`

I then continued, with `oText 3<Esc>` etc, but `<Ctrl-A>` always inserts `Text 1`

---
Peek 2020-06-09 09-38

In this example I type

```
i1<Esc>a2<Esc>a3<Esc>a<Ctrl-A><Esc>
```
I would expect the result to be `1233` but what gets output is `123123`.

I continue with `a4<Esc>a<Ctrl-A><Esc>` etc and instead of appending `4` twice, it adds the 4 and then a duplicate of the entire buffer to that point: `12312341231234`

---

Theoretically this is with a blank init.vim

ObliviousJamie:
With this change using Ctrl-A in insert mode will take into account the column position as well. This means that only the newest insert gets pasted.

closes #283

### DIFF
--- a/src/test/suite/insert-mode.test.ts
+++ b/src/test/suite/insert-mode.test.ts
@@ -487,14 +487,14 @@ describe("Insert mode and buffer syncronization", () => {
         await vscode.window.showTextDocument(doc, vscode.ViewColumn.One);
         await wait();
 
-        await sendVSCodeKeys("i1"); 
+        await sendVSCodeKeys("i1");
         await sendEscapeKey(1000);
-        await sendVSCodeKeys("a2"); 
+        await sendVSCodeKeys("a2");
         await sendEscapeKey(1000);
-        await sendVSCodeKeys("a3"); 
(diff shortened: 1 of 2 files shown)


Shatur95:
@benjie, could you check if this fixes issue for you?

benjie:
@Shatur95 Afraid I no longer use VSCode (back to Vim + tmux for me!), but the test looks like it covers what I would do to test it if I did :+1:

Shatur95:
@benjie, good choice, BTW :)

Shatur95 (approved):

PR #428, opened 2020-11-07
Launch nvim in WSL in a shell-agnostic way

### LINKED ISSUE #147: Launching in WSL
Ok, I know what's the problem. The command for launching in `nvim` is *bash-specific* and therefore it fails to create a process. Or, more precisely, the process dies almost immediately after being created, because it doesn't actually spawn `nvim`.

https://github.com/asvetliakov/vscode-neovim/blob/21ba0eb0ac483f45d73fb1333e46ae3178e4448d/src/controller.ts#L267

This line does a `$(wslpath 'c:\blabla')`. But that code doesn't work in *fish*, which is my default shell. So we should probably either start `nvim` inside of a *bash*, or somehow make the path resolution shell-agnostic.

Sorry to cause trouble.

_Originally posted by @DanielFabian in https://github.com/asvetliakov/vscode-neovim/issues/145#issuecomment-573396921_

longhungn:
Fixes #147 

I tested with bash and fish in VSCode 1.51.0

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -34,7 +34,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     try {
         const plugin = new MainController({
             customInitFile: customInit,
-            extensionPath: context.extensionPath,
+            extensionPath: context.extensionPath.replace(/\\/g, "\\\\"),
             highlightsConfiguration: {
                 highlights: highlightConfHighlights,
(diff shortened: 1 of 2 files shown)


Shatur95:
I tested it on Linux, it not breaks anything, so I think that we can merge it.

Shatur95 (approved):

PR #429, opened 2020-11-08
Set vim cwd as vscode workspace

Orpidus:
When the first file is opened, set vim cwd to the
current workspace folder root if one is open
else the home directory of the user(determined by '~').

The workspace folder is not set properly before the first file is opened, so the neovim 'c' flag does not work and 'cmd' is used.

Since extensions are reloaded when a new workspace is opened, the correct directory is set even when a new workspace is opened.

This change will make 'vim-fugitive' work.


Shatur95:
Could you resolve conflicts?

theol0403 (approved):
It works great for me!

Orpidus:
> 
> 
> Could you resolve conflicts?

Resolved!

PR #431, opened 2020-11-10
fix(visiblerange): try to restore visiblerange after set cursorpos

### LINKED ISSUE #427: Focus jumps to latest position when switching tabs
Please see at: ( sorry couldn't convert it to gif)
https://webmshare.com/play/PxBdq

As you can see, I have the first file opened, I move the cursor to the line where it says "test", scroll back to the top of that file without moving the cursor. Then I change to a different file in a different tab and when I change back to the previous file, the file focus is change (there is a "jump") to the cursor.

This really bugs me and distracts me when I'm trying to write code. Is this intended? Is there a way to disable this behavior?

Thanks.

kwonoj:
- closes #427

While this is not elegant solution, this PR attempt to address #427 as much. Before setting cursor position it try to preserve last known visible range of active document, then force set once after cursor position is set. It still slightly moves scroll position since revealing range based on line number - if scroll position is in between two lines, slight scroll can be observed.

I don't believe this is best approach - but this is one of daily workflow blockers for me, and having this seems better than nothing.

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -91,6 +91,8 @@ export class CursorManager
     }
 
     public handleRedrawBatch(batch: [string, ...unknown[]][]): void {
+        const activeEditorVisibleRanges = window.activeTextEditor?.visibleRanges;
+
         const winCursorsUpdates: Map<number, { line: number; col: number }> = new Map();
         for (const [name, ...args] of batch) {
             const firstArg = args[0] || [];
@@ -179,6 +181,17 @@ export class CursorManager
             }
         }
         winCursorsUpdates.clear();
+
+        if (!activeEditorVisibleRanges?.[0]) {
+            return;
+        }
+
+        this.logger.debug(
+            `${LOG_PREFIX}: found preserved visibleRange ${JSON.stringify(
+                activeEditorVisibleRanges,
+            )}, trying to restore`,
+        );
+        window.activeTextEditor?.revealRange(activeEditorVisibleRanges[0], TextEditorRevealType.AtTop);
     }
 
     /**


Shatur95:
@amitlevy21, could you test it? Just download artifact from [here](https://github.com/asvetliakov/vscode-neovim/actions/runs/355404579).

amitlevy21:
@Shatur95 works for me! :tada:

Shatur95:
I tested myself, jittering is very noticeable.

kwonoj:
I'm totally ok to close PR if there's way to fix other way around. I couldn't find possible api surface allows preserve range other than line numbers unfortunately. I am not fully satisfied either.

Shatur95:
Just wait, I think that @asvetliakov could help with this.

asvetliakov:
Implementing `cursor-follows-scroll` here should if not solve the problem, but at least make it 90% better. The problem is not the setting the cursor itself, but rather revealing it. If the cursor would be within visible ranges (and it will be if it would follow scroll), it will be handled by [`TextEditorRevealType.Default`](https://github.com/asvetliakov/vscode-neovim/blob/ff0082fc0234f069c8cf0983dab9d5820ad9869a/src/cursor_manager.ts#L378) reveal type, which does little scrolling as possible (usually max 1 line depending on your viewport height).

[see these lines for example](https://github.com/asvetliakov/vscode-neovim/blob/ff0082fc0234f069c8cf0983dab9d5820ad9869a/src/cursor_manager.ts#L449) , they're little obsoleted though but should bring the idea.

amitlevy21:
> I tested myself, jittering is very noticeable.

Interesting, as I did not encounter any jitter at all.

However, now when I search with `/` the matching search term is not auto focused (I need to scroll it to find where the cursor is)

205g0:
> [see these lines for example](https://github.com/asvetliakov/vscode-neovim/blob/ff0082fc0234f069c8cf0983dab9d5820ad9869a/src/cursor_manager.ts#L449) , they're little obsoleted though but should bring the idea.

Where do we add those lines?

asvetliakov:
I think the original problem was solved, so closing.

PR #452, opened 2020-12-16
H M L keys: Fix bugs, add jumplists.

kar9222:
* Fix H M L keys first moving to first character of the line before it's triggered. Closes https://github.com/asvetliakov/vscode-neovim/issues/367
* Internally, H M L keys uses native VSCode commands, which don't register jumplist. Fix by registering jumplist in Vim e.g. for subsequent use of `<C-o>`

If you look at commit messages https://github.com/asvetliakov/vscode-neovim/commit/e67f5203ed8ec0bb958982a06c464e6758f4d54e, initially I added hacky fixes

```ts
case "move-cursor": {
    const [to] = args as ["top" | "middle" | "bottom"];

    // Fix bug: `goToLine` First moving to first character of the line before `goToLine` is triggered.
    // Temporary solution: Move cursor to first character of the line and delay by 5 milliseconds before triggering `goToLine`. 1 millisecond delay failed manual tests.
    vscode.commands.executeCommand("cursorHome");
    await new Promise((resolve) => setTimeout(resolve, 5));

    this.goToLine(to);
    break;
}
```

But after adding jumplists using Vim commands at https://github.com/asvetliakov/vscode-neovim/commit/c83dacd19f86cd7c572e28b9fbff1012958c21a2, which inherently has some milliseconds delays and it somehow magically fixes the bugs lol. So I remove the hacky fixes in the last commit https://github.com/asvetliakov/vscode-neovim/commit/39eb8a212cd4c9f61d21a571d80905e7b520c350. 

If possible, when merging this PR, please squash the commit messages for future references in case the bugs somehow come back. The bugs are still there but somehow adding jumplist made the bugs go away.

# Test

fix_h_l_m_keys

### DIFF
--- a/src/commands_controller.ts
+++ b/src/commands_controller.ts
@@ -51,6 +51,12 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
             }
             case "move-cursor": {
                 const [to] = args as ["top" | "middle" | "bottom"];
+
+                // Fix bug: `goToLine` First moving to first character of the line before `goToLine` is triggered.
+                // Temporary solution: Move cursor to first character of the line and delay by 5 milliseconds before triggering `goToLine`. 1 millisecond delay failed manual tests.
+                vscode.commands.executeCommand("cursorHome");
+                await new Promise((resolve) => setTimeout(resolve, 5));
+
                 this.goToLine(to);
                 break;
             }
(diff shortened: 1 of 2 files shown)


theol0403 (approved):
Thank you for the PR!
Timing issues are the worst.

I agree a squash merge would be best.

PR #455, opened 2020-12-17
Add g0 g$ for normal mode

kar9222:
Add `g0` and `g$` for normal mode. (Partially?) Closes https://github.com/asvetliakov/vscode-neovim/issues/433

For other modes e.g. `vg0` `cg0`, they are not as straightforward due to e.g. 
* visual mode not producing the real VSCode selection
* delay between VSCode command and Vim command

I tried adding `vg0` `vg$`, playing around a bit but gave up. Kinda buggy. Might revisit in future. `g0` and `g$` covers some use cases for now.

# Test 

Start/end of screen line motion in normal mode
`g0`
`g$`
<br>

add_g0_g$__v2

### DIFF
--- a/vim/vscode-motion.vim
+++ b/vim/vscode-motion.vim
@@ -1,0 +1,10 @@
+function! s:toLastCharOfScreenLine()
+    call VSCodeNotify('cursorEnd')
+    " Optmized delay is required after calling VSCode command
+    sleep 85m
+    " Offset cursor moving to the right by 1 column caused by calling `VSCodeNotify('cursorEnd')` in Vim mode
+    normal! h
+endfunction
+
+nnoremap g0 <Cmd>call VSCodeNotify('cursorHome')<CR>
+nnoremap g$ <Cmd>call <SID>toLastCharOfScreenLine()<CR>
--- a/vim/vscode-motion.vim
+++ b/vim/vscode-motion.vim
@@ -1,10 +1,12 @@
-function! s:toLastCharOfScreenLine()
-    call VSCodeNotify('cursorEnd')
-    " Optmized delay is required after calling VSCode command
-    sleep 85m
-    " Offset cursor moving to the right by 1 column caused by calling `VSCodeNotify('cursorEnd')` in Vim mode
-    normal! h
+function! s:toFirstCharOfScreenLine()
+    call VSCodeExtensionNotify('cursor-move', 'wrappedLineFirstNonWhitespaceCharacter')
 endfunction
 
-nnoremap g0 <Cmd>call VSCodeNotify('cursorHome')<CR>
+function! s:toLastCharOfScreenLine()
+    call VSCodeExtensionNotify('cursor-move', 'wrappedLineLastNonWhitespaceCharacter')
+    " Offfset cursor moving to the right caused by calling VSCode command in Vim mode
+    call VSCodeNotify('cursorLeft')
+endfunction
+
+nnoremap g0 <Cmd>call <SID>toFirstCharOfScreenLine()<CR>
 nnoremap g$ <Cmd>call <SID>toLastCharOfScreenLine()<CR>
--- a/vim/vscode-motion.vim
+++ b/vim/vscode-motion.vim
@@ -1,9 +1,9 @@
 function! s:toFirstCharOfScreenLine()
-    call VSCodeExtensionNotify('cursor-move', 'wrappedLineFirstNonWhitespaceCharacter')
+    call VSCodeNotify('cursorMove', { 'to': 'wrappedLineFirstNonWhitespaceCharacter' })
 endfunction
 
 function! s:toLastCharOfScreenLine()
-    call VSCodeExtensionNotify('cursor-move', 'wrappedLineLastNonWhitespaceCharacter')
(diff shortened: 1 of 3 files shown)


vim/vscode-motion.vim
```diff
@@ -0,0 +1,10 @@
+function! s:toLastCharOfScreenLine()
+    call VSCodeNotify('cursorEnd')
+    " Optmized delay is required after calling VSCode command
```
Shatur95, line 3: This is a very dirty hack. Have you tried to use `VSCodeCall` instead?

kar9222: I refactor with VSCode commands and removing the `sleep` in https://github.com/asvetliakov/vscode-neovim/pull/455/commits/670718fcddec938acd6b9d43de771621d55f61cf

kar9222:
Additionally, the VSCode built-in `vscode.commands.executeCommand("cursorMove", ...);` is lower level and can be used to simplify previous and future implementations and avoid hacky solutions.

kar9222:
The GIF in my first message is updated. `g$` doesn't have delay now.

Shatur95:
How is `VSCodeExtensionNotify` different from `VSCodeCall`?

kar9222:
Sorry I forgot to mentioned that I also tested `VSCodeCall` which  also has the same behaviour.

Also, the new `cursorMove` added for this PR is a more complete command exported by VSCode https://github.com/microsoft/vscode/blob/0adc15c4bb876a9537a5b25a8fb64786e6396e36/src/vs/editor/common/controller/cursorMoveCommands.ts#L581-L834


* can be re-used in another PR for resolving the bugs the hidden bugs in https://github.com/asvetliakov/vscode-neovim/pull/452. 
* is simpler and has more options than `goToLine` used in https://github.com/asvetliakov/vscode-neovim/pull/452 and can be exported to for users e.g. `call VSCodeExtensionNotify('cursor-move', 'wrappedLineStart')` if they want more control in the future, or we might use it internally in the future.

My local branch has a more complete version of `cursorMove`. I intend to submit another PR for resolving the 'hidden' bugs in #452 to avoid confusion.

```ts
case "cursor-move": {
    const [to, by] = args as [
        (
            | "wrappedLineStart"
            | "wrappedLineEnd"
            | "wrappedLineFirstNonWhitespaceCharacter"
            | "wrappedLineLastNonWhitespaceCharacter"
            | "viewPortTop"
            | "viewPortCenter"
            | "viewPortBottom"
        ),
        "line" | "wrappedLine" | "character" | "halfLine",
    ];
    this.cursorMove(to, by);
    break;
}

  private cursorMove = (
      to:
          | "wrappedLineStart"
          | "wrappedLineEnd"
          | "wrappedLineFirstNonWhitespaceCharacter"
          | "wrappedLineLastNonWhitespaceCharacter"
          | "viewPortTop"
          | "viewPortCenter"
          | "viewPortBottom",
      by: "line" | "wrappedLine" | "character" | "halfLine",
  ): void => {
      vscode.commands.executeCommand("cursorMove", { to: to, by: by, value: 1, select: false });
  };
```

```vim
function s:cursorMove(to)
    let to = a:to
    normal! m'
    call VSCodeExtensionNotify('cursor-move', to)
endfunction

nnoremap H <Cmd>call <SID>cursorMove('viewPortTop')<CR>
xnoremap H <Cmd>call <SID>cursorMove('viewPortTop')<CR>
nnoremap M <Cmd>call <SID>cursorMove('viewPortCenter')<CR>
xnoremap M <Cmd>call <SID>cursorMove('viewPortCenter')<CR>
nnoremap L <Cmd>call <SID>cursorMove('viewPortBottom')<CR>
xnoremap L <Cmd>call <SID>cursorMove('viewPortBottom')<CR>
```

and hence below can be completely removed...

https://github.com/asvetliakov/vscode-neovim/blob/670718fcddec938acd6b9d43de771621d55f61cf/src/commands_controller.ts#L125-L147

Shatur95:
If `VSCodeExtensionNotify` behaves the same as `VSCodeCall`, why do we need an additional command?

kar9222:
For
* `wrappedLineFirstNonWhitespaceCharacter` and `wrappedLineFirstNonWhitespaceCharacter` used in this PR
* Other reasons mentioned in the above comment, in summary
    * `viewPortTop` `viewPortCenter` `viewPortBottom` for resolving the hidden bug in #452 
    * exported for user who wants more control `wrappedLineStart`, `wrappedLineEnd`
    * simpler and more controls than the current internal function `goToLine`, that we might use them in future

Shatur95:
We need to review from @asvetliakov for it.

asvetliakov:
>How is VSCodeExtensionNotify different from VSCodeCall?

`VSCodeExtensionCall/VSCodeExtensionNotify` are used to send custom non non-vscode command to the extension, `VSCodeCall/Notify` pass it directly to vscode.

I think you can use `VSCodeNotify` directly here without making any changes to ext code:

```vim
function! s:toFirstCharOfScreenLine()
    call VSCodeNotify('cursorMove',  { 'to': 'wrappedLineFirstNonWhitespaceCharacter' })
endfunction
```

kar9222:
Should be good to go now. Manually tested as per the 1st comment above.

Shatur95 (approved):

asvetliakov (approved):

PR #461, opened 2020-12-28
Use vscode's cursorMove and jumplist and other cursor fixes

### LINKED ISSUE #386: Unable to type into comment boxes when using GitHub PR extension
When attempting to leave a pull request comment using the GitHub PR extension, the cursor jumps to the top of the file and prevents typing in the comment box.

## Steps to reproduce:
- Open a file from the GitHub PR extension.
- Click on one of the changed lines
- Click inside the comment box
- Start typing

## Expected Behavior
- Text is typed into the comment box

## Actual Behavior
- Cursor jumps to line 1 of the file

## Affected Versions
VSCode version 1.49.1
NeoVim extension version 0.0.63 (although this has been in the last several releases)

### LINKED ISSUE #392: Jump motion CTRL-O opens unsaved buffer instead of previously opened file
**Steps:**

1. Have Vetur installed (it also reproduces with other files, but with Vetur I could reproduce it constantly)
2. Make changes in `.vue` a file and don't save those.
3. Jump to any other file using VSCode file search.
4. Go back using default vim hotkey `CTRL-o`

**Expected:**
Previously opened `vue` file is focused again

**Actual:**
Buffer with no association with previously opened file is focused

vscode-neovim-vetur

It doesn't happen with all `.vue` files, though. Only several are affected. I will try to dig into the code later and see why this is happening, but would also appreciate some guidance on where to look. 

Env:
```
VSCode: Version: 1.49.1
NVIM v0.5.0-dev
OS Fedora 32
```

init.vim:
```vimrc
call plug#begin('~/.vim/plugged')
Plug 'tpope/vim-surround'
Plug 'tpope/vim-repeat'
Plug 'tpope/vim-unimpaired'
Plug 'posva/vim-vue'
call plug#end()

xmap gc  <Plug>VSCodeCommentary
nmap gc  <Plug>VSCodeCommentary
omap gc  <Plug>VSCodeCommentary
nmap gcc <Plug>VSCodeCommentaryLine

nnoremap <C-h> <Cmd>call VSCodeNotify('workbench.action.focusLeftGroup')<CR>
xnoremap <C-h> <Cmd>call VSCodeNotify('workbench.action.focusLeftGroup')<CR>
nnoremap <C-j> <Cmd>call VSCodeNotify('workbench.action.focusBelowGroup')<CR>
xnoremap <C-j> <Cmd>call VSCodeNotify('workbench.action.focusBelowGroup')<CR>
nnoremap <C-k> <Cmd>call VSCodeNotify('workbench.action.focusAboveGroup')<CR>
xnoremap <C-k> <Cmd>call VSCodeNotify('workbench.action.focusAboveGroup')<CR>
nnoremap <C-l> <Cmd>call VSCodeNotify('workbench.action.focusRightGroup')<CR>
xnoremap <C-l> <Cmd>call VSCodeNotify('workbench.action.focusRightGroup')<CR>
```

### LINKED ISSUE #404: Split window cursor retains its original position
when using vscode with split editors, a mouse click on the text in an inactive editor pane will cause the cursor to jump to the last position of the cursor for that file. How do I get the cursor to remain exactly where I clicked

asvetliakov:
This switches to `cursorMove` vscode command for some editor movement and switches jumplist to vscode one. Also changes some logic regarding unloading buffers - since we don't need to use vim jumplist, we can delete them when hidden.

Also fixes uppercase marks ( #228 ) 

Fixes #404 
Fixes #392 
Fixes #386


theol0403:
This looks awesome! Does this mean that all the jumps in the other vim files can be removed?
Like https://github.com/asvetliakov/vscode-neovim/blob/d191c89e5e2d81af45205fe87561cbd0c97cdab9/vim/vscode-scrolling.vim#L19-L22

Also, does this mean that moving the cursor by clicking with the mouse, and other jumps that we don't catch (like  vscode outline links), will now properly work with the jumplist?

asvetliakov:
Yes. It will use native vscode jumplist so no more jump hacks. Vim jumplist can be still used manually within same buffer though (won't work for jumping to other buffer)

Yes, this can be removed too, i'll clean after fixing few other problems

asvetliakov:
Ok, despite some random test failing, latest commits should be stable:

Test version
https://github.com/asvetliakov/vscode-neovim/suites/1736049460/artifacts/32913741

theol0403:
Sorry @asvetliakov if I should be creating a new issue for this, but while this PR really improves things, I noticed a bug. 

If you have a vscode tab open, you make some changes, you switch to a different tab, and then you switch back to the original tab, you are no longer able to undo those changes.

asvetliakov:
@theol0403 #466

theol0403:
Fixed! Thank you so much.

asvetliakov:
released 0.0.72

Issue #477, opened 2020-12-30
scrolling with cursor makes page jump


trusktr:
When I scroll with the cursor in VS Code (f.e. holding `j`), the page jumps like this:

https://user-images.githubusercontent.com/297678/103330141-127eca80-4a15-11eb-8a46-fec21f5f8eaf.mp4

But I would like it to be like in neovim with no jumping:

https://user-images.githubusercontent.com/297678/103330258-81f4ba00-4a15-11eb-97c8-6de79cfa482d.mp4

Is there any way to prevent the jumping?

asvetliakov:
Your cursor is just too 🐎  🐎  🐎  🚗  

Here the behaviour at normal speed:

https://user-images.githubusercontent.com/8881674/103330427-bf7f3480-4a61-11eb-9c31-b73fbff06fa7.mp4

Changing this:
https://github.com/asvetliakov/vscode-neovim/blob/d366766f397954078e5b9eeb01532913fb90b48f/src/cursor_manager.ts#L422

to `TextEditorRevealType.Default` can help but will possibly degrade UX of over jumps

asvetliakov:
I just compiled PR in #479 . Please try attached build to GHA check once them completed.

trusktr:
> Your cursor is just too :racehorse: :racehorse:  :racehorse:  :car:

Oh, I see. :smile: 

> I just compiled PR in #479 .

I think it might be a good change. I also have mappings like this:

```js
nmap <c-j> 5j
nmap <c-k> 5k
```

and those trigger it on a single move, which makes sense based on what you described, but also not like Neovim:

VS Code:

![5-at-a-time-vscode](https://user-images.githubusercontent.com/297678/103330761-b4071b80-4a17-11eb-93ea-826d4bd2265f.gif)

Neovim:

![5-at-a-time-neovim](https://user-images.githubusercontent.com/297678/103330764-b8cbcf80-4a17-11eb-9aa2-c4f787dc46af.gif)

trusktr:
> I just compiled PR in #479 . Please try attached build to GHA checks once they are completed.

Cool, it works exactly as expected! 

Also I tried VS Code's `Go to Definition` command on some things, and I don't experience any issues, it seems to work fine like before. What other other jumps should I test (and how specifically)?

asvetliakov:
>Also I tried VS Code's Go to Definition command on some things, and I don't experience any issues

Is the screen centered after go to def?

[...]

PR #479, opened 2020-12-30
Exploration - Try default revealing

asvetliakov:
related to #477 . This probably will lead to some consequences for go-to-def/outline navigation and other code jumps - they won't be centered anymore.


trusktr (approved):
Works great for me!

asvetliakov:
Latest commit is trying to improve things:
if jump is outside of top/bottom visible line + 10 line margin - the viewport would be centered. Otherwise it will use default strategy with little scrolling as possible

PR #482, opened 2020-12-30
Refactor hl attrs to not override default colors

theol0403:
If the foreground or background colour keys are not specified, the plugin forcefully sets the colours to be the theme's default foreground/background colours. This means that if the text colour was changed by syntax hl, then it would become white. 

This is useful for underlining something without changing the colour - notably for use with `quick-scope` or `vim-current-word`:
```vim
  highlight QuickScopePrimary guifg=NONE guisp='#afff5f' gui=underline ctermfg=NONE cterm=underline
  highlight QuickScopeSecondary guifg=NONE guisp='#5fffff' gui=underline ctermfg=NONE cterm=underline
```

image

Hopefully this does not cause any problems! I'm lucky to have figured this out :slightly_smiling_face: 

Do you think the above snippet should be added to the readme?

Finally, do you think the default underline "vimSpecialColor" should be changed to something more neutral, like the foreground colour?

Another quick thought, further down for the "reverse" attr, should the colours be read then inverted from the colour keys instead of just default foreground/background colour?

### DIFF
--- a/src/highlight_provider.ts
+++ b/src/highlight_provider.ts
@@ -42,11 +42,11 @@ function vimHighlightToVSCodeOptions(
     vimSpecialColor: string,
 ): ThemableDecorationRenderOptions {
     const options: ThemableDecorationRenderOptions = {};
-    // for absent color keys default color should be used
-    options.backgroundColor = uiAttrs.background
-        ? "#" + uiAttrs.background.toString(16)
-        : new ThemeColor("editor.background");
-    if(uiAttrs.foreground) {    
+    // for absent color keys color should not be changed
+    if (uiAttrs.background) {
+        options.backgroundColor = "#" + uiAttrs.background.toString(16);
+    }
+    if (uiAttrs.foreground) {
         options.color = "#" + uiAttrs.foreground.toString(16);
     }
     const specialColor = uiAttrs.special ? "#" + uiAttrs.special.toString(16) : vimSpecialColor;
--- a/src/highlight_provider.ts
+++ b/src/highlight_provider.ts
@@ -46,7 +46,9 @@ function vimHighlightToVSCodeOptions(
     options.backgroundColor = uiAttrs.background
         ? "#" + uiAttrs.background.toString(16)
         : new ThemeColor("editor.background");
-    options.color = uiAttrs.foreground ? "#" + uiAttrs.foreground.toString(16) : new ThemeColor("editor.foreground");
+    if(uiAttrs.foreground) {    
+        options.color = "#" + uiAttrs.foreground.toString(16);
+    }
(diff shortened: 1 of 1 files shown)


asvetliakov:
I just followed API instructions here i believe:
```
["hl_attr_define", id, rgb_attr, cterm_attr, info]
	Add a highlight with `id`  to the highlight table, with the
	attributes specified by the `rgb_attr` and `cterm_attr` dicts, with the
	following (all optional) keys.

	`foreground`:	foreground color.
	`background`:	background color.
	`special`:	color to use for underline and undercurl, when present.
	`reverse`:	reverse video. Foreground and background colors are
			switched.
	`italic`:	italic text.
	`bold`:		bold text.
	`strikethrough`:  struckthrough text.
	`underline`:	underlined text. The line has `special` color.
	`undercurl`:	undercurled text. The curl has `special` color.
	`blend`:	Blend level (0-100). Could be used by UIs to support
			blending floating windows to the background or to
			signal a transparent cursor.

	For absent color keys the default color should be used. Don't store
	the default value in the table, rather a sentinel value, so that
	a changed default color will take effect.
	All boolean keys default to false, and will only be sent when they
	are true.

	Highlights are always transmitted both for both the RGB format and as
	terminal 256-color codes, as the `rgb_attr` and `cterm_attr` parameters
	respectively. The |ui-rgb| option has no effect effect anymore.
	Most external UIs will only need to store and use the `rgb_attr`
	attributes.

	`id` 0 will always be used for the default highlight with colors defined
	by `default_colors_set` and no styles applied.

	Note: Nvim may reuse `id` values if its internal highlight table is full.
	In that case Nvim will always issue redraws of screen cells that are
	affected by redefined ids, so UIs do not need to keep track of this
	themselves.

	`info` is an empty array by default, and will be used by the
	|ui-hlstate| extension explained below.
```

asvetliakov:
But since vscode is not neovim, i don't think it would break something. So... probably safe to merge?

asvetliakov (approved):

theol0403:
Interesting! From how I interpret the "sentinel" section however, it seems to say that the default (as in the current colour of the text) should not be overridden.

I'm pretty sure vim's behaviour aligns with this PR, as NONE does not mean select foreground/background, it means don't change.

theol0403:
Before this is merged, I have a change to `vimSpecialColor` that will make it match vim's behaviour.

theol0403:
I tested with neovim, and if some text has an underline *without* a special color, the color of the text is used, not the foreground color, nor some random fixed color. 

If you are alright with this change I am ready to merge.

asvetliakov:
go

theol0403:
Is this error normal?

asvetliakov:
tests are far from ideal, so sometimes this happens

PR #489, opened 2020-12-30
Fix word HL after setting cursor directly

### LINKED ISSUE #159: Highlight occurrences doesn't work in normal mode
So the highlight occurrences feature in vs code doesn't work while vim is in normal mode, I understand that this might be an implication of vscode not knowing where the cursor is located while vim is in normal mode but is there a way to get around that?
I am taking about this 
image
Notice how instance of `anchor_start` are highlighted, but that's only in insert mode because vim isn't engaged.

I also wonder if the extension might report to vscode the current cursor location which i think it does because other things like bracket highlighting seems to work, then why doesn't it work with word occurrences?

### LINKED ISSUE #488: sometimes VS Code cursor is not updated
This time I got the correct title!

In the following video, when I press `%`, the cursor _does_ move to the matching closing tag in Neovim, but this is not reflected in VS Code. After a moment, when I press `h` or `l` to move the cursor by one character, then VS Code will finally update to the correct output.

no-cursor-update

In the video,

- I moved the cursor initially with `h` and `l`, while on the opening `<mesh>` tag
- I pressed `%` after I placed the cursor at the beginning of the `mesh` word (after a small delay).
- VS Code should have immediately rendered the cursor on the closing `</mesh>` tag, on top of the `/` character.
- After a moment, I finally pressed `l`
- Now VS Code correctly rendered the cursor on top of the `m` character that is one to the right of the `/` character.

asvetliakov:
Fixes #159

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -432,6 +432,7 @@ export class CursorManager
                             : TextEditorRevealType.Default
                         : TextEditorRevealType.Default;
                 editor.revealRange(newPos, type);
+                commands.executeCommand("editor.action.wordHighlight.trigger");
             }
         }
     };


asvetliakov:
Big thanks to @huejim for pointing to the correct comamnd

PR #494, opened 2021-01-01
Also close the window on ctrl+w ctrl+c as expected

### LINKED ISSUE #483: <c-w><c-c> not working.
According to neovim docs,

```
|CTRL-W_CTRL-C|	CTRL-W CTRL-C	   same as "CTRL-W c"
```

so it should also close the window, but currently doesn't. It's a good thing to have because I can press `wcwcwc` a lot faster while my pinky holds `ctrl`. Pressing `ctrl+w c ctrl+w c ctrl+w c` is slower and error prone (easy to accidentally do `<c-w><c-c>`, which is probably why that map exists in neovim).

trusktr:
Fixes #483

### DIFF
--- a/vim/vscode-window-commands.vim
+++ b/vim/vscode-window-commands.vim
@@ -63,6 +63,8 @@ nnoremap <C-w>q <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 xnoremap <C-w>q <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 nnoremap <C-w>c <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 xnoremap <C-w>c <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
+nnoremap <C-w><C-c> <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
+xnoremap <C-w><C-c> <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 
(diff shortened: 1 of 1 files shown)


vim/vscode-window-commands.vim
```diff
@@ -59,6 +59,8 @@ nnoremap <C-w>q <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 xnoremap <C-w>q <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 nnoremap <C-w>c <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
 xnoremap <C-w>c <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
+nnoremap <C-w><C-c> <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
+xnoremap <C-w><C-c> <Cmd>call VSCodeNotify('workbench.action.closeActiveEditor')<CR>
```
trusktr, line 63: Do you prefer `<C-w><C-c>` capitalization format instead of `<c-w><c-c>` with all lowercase?

asvetliakov:
Pulling latest master must fix test failures

vim/vscode-window-commands.vim
samvv: Looking at the other code, `<C-w><C-c>` seems like the way to go.

samvv:
I rebased your branch on `master`. It seems like the tests were just canceled. Let's see if they pass with the latest version.

samvv:
LGTM Unless @asvetliakov has something against this, I'll merge this in tomorrow.

asvetliakov (approved):

PR #499, opened 2021-01-02
Cancel current mode when switching to other editor

### LINKED ISSUE #156: Cursor frozen if you switch tab while in visual mode
cursor freeze after switching tab in visual

1. Create two untitled files
2. Write some lines in the first one, go into visual mode and select several lines
3. Go to the other untitled and try to go into insert mode and write something. Nothing will happen

Actually the cursor unfreezes if you press `s` (substitute) after moving to the new tab

vscode-neovim 0.50
vscode 1.41.1

asvetliakov:
Fix #156

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -179,7 +179,7 @@ export class MainController implements vscode.Disposable {
         this.commandsController = new CommandsController(this.client, this.settings.revealCursorScrollLine);
         this.disposables.push(this.commandsController);
 
-        this.modeManager = new ModeManager(this.logger);
+        this.modeManager = new ModeManager(this.logger, this.client);
         this.disposables.push(this.modeManager);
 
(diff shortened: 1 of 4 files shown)


asvetliakov:
Actually vim preserves visual mode when switching to the same buffer in different split. But personally i found this very frustrating and not useful to bother to replicate this behaviour.

asvetliakov:
@3nuc Can you try this build?

https://github.com/asvetliakov/vscode-neovim/suites/1753779528/artifacts/33363352

3nuc:
> 
> 
> @3nuc Can you try this build?
> 
> https://github.com/asvetliakov/vscode-neovim/suites/1753779528/artifacts/33363352

I tried this and it indeed fixes 153 - the repro steps I posted in the issue no longer prevent insert mode from starting (and the mode is reset when changing editors).

PR #500, opened 2021-01-02
Use latest cursor position from neovim rather than vscode to calculate cursor movement diff

### LINKED ISSUE #495: My cursor is on speed, often leaves VS Code behind.
As you may have noticed in some of my recordings, my cursor is :horse_racing: :horse_racing: :horse_racing: :horse_racing: :horse_racing: :shopping_cart:.

I some times move too fast, and VS Code rendering will not be up to date. Or sometimes, VS Code's cursor will have moved further than mine. Maybe this happens once it has been slammed with all the previous movement, and then VS Code handles the input instead of neovim? I think that may be the case, because sometimes in insert mode I'll type some stuff, then ctrl+c and start moving with `j`, but instead I'll see `jjjjjj` inputted into the text.

I think this is also more likely to happen when other things are taking up CPU, so that it slows both VS Code and Neovim down.

I noticed the `Vim` plugin also has the same issue.

asvetliakov:
Fixes #495 

Ops, totally broken

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -50,7 +50,7 @@ export class CursorManager
      * ! Note: we should track this because setting cursor as consequence of neovim event will trigger onDidChangeTextEditorSelection with Command kind
      * ! And we should skip it and don't try to send cursor update into neovim again, otherwise few things may break, especially jumplist
      */
-    private neovimCursorPosition: WeakMap<TextEditor, { line: number; col: number }> = new WeakMap();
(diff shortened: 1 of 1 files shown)


jvbianchi:
I just tried this version, the cursor does not move.

asvetliakov:
Yep, let me adjust the PR

jvbianchi:
After this fix, it seems to be working fine.

asvetliakov:
Unfortunately there are many difficulties with this approach. I need to find different way

PR #501, opened 2021-01-02
Cursor updates

### LINKED ISSUE #495: My cursor is on speed, often leaves VS Code behind.

asvetliakov:
~1. Bind `j/k` to VSCode's `cursorMove` command~ This breaks macros unfortunately
2. When cursor movement is within same line - set it directly rather than call `cursorMove` command. This has much better speed and doesn't create jumppoint if within same line

Solves same line movement in #495

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -458,15 +458,18 @@ export class CursorManager
         if (Math.abs(deltaLine) <= 1) {
             this.logger.debug(`${LOG_PREFIX}: Editor: ${editorName} using cursorMove command`);
             if (Math.abs(deltaLine) > 0) {
-                this.logger.debug(`${LOG_PREFIX}: Editor: ${editorName} Moving cursor by line: ${deltaLine}, char: 0`);
-                commands.executeCommand("cursorLineStart");
+                if (newCol !== currCursor.character) {
+                    deltaChar = newCol;
+                    commands.executeCommand("cursorLineStart");
+                } else {
+                    deltaChar = 0;
+                }
                 commands.executeCommand("cursorMove", {
                     to: deltaLine > 0 ? "down" : "up",
                     by: "line",
                     value: Math.abs(deltaLine),
                     select: false,
                 });
-                deltaChar = newCol;
             }
             if (Math.abs(deltaChar) > 0) {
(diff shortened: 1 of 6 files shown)


jvbianchi:
It works, but sometimes the cursor goes back and forth (in the video in the end of the line).


https://user-images.githubusercontent.com/13983663/103462197-5a963980-4d02-11eb-9052-2c0742022289.mp4

asvetliakov:
How is the movement latency compared to 0.0.63?

jvbianchi:
There is a high latency, sometimes when I stop pressing `llllllllllllll` the cursor continues moving  for a couple of seconds.

asvetliakov:
Is it workable?
Are you using it through WSL or native win nvim?

jvbianchi:
I'm on linux (ubuntu 20.04).

I would say that the latency is comparable to https://github.com/VSCodeVim/Vim

asvetliakov:
Ah, sorry, confused with `wsl.exe` from your video.

jvbianchi:
Oh its a shared config file.

jvbianchi:
Using the mapping that you passed before, its like theres no latency when a compare to this PR (high latency).

asvetliakov:
Ok, let me try one more trick

asvetliakov:
@jvbianchi Try with latest commit again

jvbianchi:
In this latest one, there is no latency and the cursor is working as expected.

asvetliakov:
awesome, thank you

PR #510, opened 2021-01-04
Closes #509

### LINKED ISSUE #509: extensionPath needs to be escaped or wrapped in quotes
https://github.com/asvetliakov/vscode-neovim/blob/9bf0183ac5c96fc8ee78cbec09e639d475f443c0/src/main_controller.ts#L103

If the extension path is not quoted (or escaped), paths with space in them cause the above to throw the error:

```
Unable to init vscode-neovim: Command failed:
C:\Windows\system32\wsl.exe wslpath c:\\Users\\My Name\\.vscode\\extensions\\asvetliakov.vscode-neovim-0.0.76 
wslpath: Invalid argument
```

This is due to the space in the path in 'My Name'.

Adding single quotes around `${extensionPath}` fixed the issue.

kyleflan:
Wrap `${extensionPath}` in single quotes to alleviate issues with spaces existing in `extensionPath` var.

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -100,7 +100,7 @@ export class MainController implements vscode.Disposable {
         let extensionPath = settings.extensionPath;
         if (settings.useWsl) {
             // execSync returns a newline character at the end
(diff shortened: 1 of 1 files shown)


asvetliakov (approved):

asvetliakov:
Thank you!

PR #517, opened 2021-01-06
Fix tab indents

### LINKED ISSUE #515: TAB-indentation causes cursor jumps when appending
When tabs are used for indentation trying to append from normal-mode by pressing **a** will cause the cursor to jump to a false position. Please see the screencast which demonstrates the problem: Pressing **a** moves the cursor far into the line.

https://streamable.com/jz0c7q

The jump distance seems to be related to the number of leading tabs: the more tabs, the farther the cursor jumps. This does not occur with leading whitespaces. The problem persists even with a completely empty *init.vim*, as well as a completely empty *settings.json* (except for the nvim path) and is independent of the file-type (plain text, golang, c).

I have never had problems with tabs but after coming back to vscode+nvim after a week of absence this problem appeared. nvim, vscode and the plugin are up-to-date.

Any ideas?

### LINKED ISSUE #516: shift+i and a are offset
Hi everyone,

Love the project and I use it everyday !

I've got a problem that occured the first time at least two weeks ago (I've been on holidays so I don't really know when was the update that broke everything).

# Bugs

- When using `a` in some lines the cursor moves forward
- When using `shift+i` in some lines the cursor doesn't go to the beginning of the line (first character that isn't a white space, as before the bug occured)

I've made a screencap to illustrate:

https://user-images.githubusercontent.com/22097902/103771059-554d2d80-5027-11eb-9752-aefd0bd13f27.mov

# Clues

I've not been able de to reproduce the bug accurately on every kind of line.

- It happens everytime I'm on a JSdoc line (maybe something with comments is broken)
- It doesn't happen on a line inside a JSdoc block if it isn't starting with a `*` (maybe JSdoc is handled with specific movements idk)
- It sometimes happens in code, but I've not been able to reproduce it for the screencap as it is intermittent.

# Details

- I'm on macOS 10.15.7
- I never see any error popping up
- I've tried more than one version of vscode-neovim
  - 0.0.76
  - 0.0.73
  - 0.0.65
- I've installed every version of neovim linked in issues of this repository having relation to `shift+i`, `shift+a`, `a`, and `i`.
  - neovim --HEAD (v0.4)
  - neovim nightly from tarball of the official repo
  - neovim nightly from `jason0x43/homebrew-neovim-nightly`
- I've killed VSCode everytime
- I've disabled every extension but vscode-neovim

asvetliakov:
Fixes #516 
Fixes #515

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -218,7 +218,7 @@ export class CursorManager
                             editor.document.lineAt(cursorPos.line).text,
                             cursorPos.col,
                             // !For cursor updates tab is always counted as 1 col
-                            1,
+                            cursorPos.isByteCol ? 1 : (editor.options.tabSize as number),
                             cursorPos.isByteCol,
                         );
                         this.neovimCursorPosition.set(editor, { line: cursorPos.line, col: finalCol });
@@ -236,7 +236,7 @@ export class CursorManager
                     const finalCol = calculateEditorColFromVimScreenCol(
                         editor.document.lineAt(cursorPos.line).text,
                         cursorPos.col,
-                        1,
+                        cursorPos.isByteCol ? 1 : (editor.options.tabSize as number),
                         cursorPos.isByteCol,
                     );
                     this.neovimCursorPosition.set(editor, { line: cursorPos.line, col: finalCol });


asvetliakov:
@lowgos try this PR build

asvetliakov:
@slavistan also

slavistan:
@asvetliako It works. Thanks for the fast reply!

PR #528, opened 2021-01-18
Fix CTRL+U/D behavior to move cursor

### LINKED ISSUE #293: Page movements don't affect cursor
`<C-d>`, `<C-u>`, `<C-f>`, and `<C-b>` all move the view but not the cursor, which is inconsistent with native Neovim behavior, which is to move the cursor a half/full screen and only adjust the view if necessary.

wcauchois:
Fixes #293.

Note: calculation of a "page" is based on [this source from VSCodeVim](https://github.com/VSCodeVim/Vim/blob/d9517680785b926b69d8d65bb1b334ec71cc1f3c/src/configuration/configuration.ts#L399).

### DIFF
--- a/src/commands_controller.ts
+++ b/src/commands_controller.ts
@@ -107,7 +107,15 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
 
     /// SCROLL COMMANDS ///
     private scrollPage = (by: "page" | "halfPage", to: "up" | "down"): void => {
-        vscode.commands.executeCommand("editorScroll", { to, by, revealCursor: true });
+        const editor = vscode.window.activeTextEditor;
+        if (!editor) {
+            return;
+        }
+        const visibleRanges = editor.visibleRanges;
+        const scrollLines = Math.ceil(
+            (visibleRanges[0].end.line - visibleRanges[0].start.line) / (by === "halfPage" ? 2 : 1),
+        );
+        vscode.commands.executeCommand("cursorMove", { to, by: "line", value: scrollLines, revealCursor: true });
     };
 
     private scrollLine = (to: "up" | "down"): void => {


src/commands_controller.ts
```diff
@@ -107,7 +107,15 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
...
     /// SCROLL COMMANDS ///
     private scrollPage = (by: "page" | "halfPage", to: "up" | "down"): void => {
-        vscode.commands.executeCommand("editorScroll", { to, by, revealCursor: true });
+        const editor = vscode.window.activeTextEditor;
+        if (!editor) {
+            return;
+        }
+        const visibleRanges = editor.visibleRanges;
+        const scrollLines = Math.ceil(
+            (visibleRanges[0].end.line - visibleRanges[0].start.line) / (by === "halfPage" ? 2 : 1),
```
asvetliakov, line 116: I think you should take both first and last visible ranges, since for folds there can be few entries here

asvetliakov (changes requested):

src/commands_controller.ts
asvetliakov: You can use this function:
https://github.com/asvetliakov/vscode-neovim/blob/266817e84d9e0f6bf63579f5c0798844b47fdf5f/src/utils.ts#L534-L543

and modify it to return `Range | undefined` rather than boolean

chaeing:
Any progress?

farnoy:
Hey @Georift , were you able to implement it the way you described?

Georift:
> Hey @Georift , were you able to implement it the way you described?

No sorry I didn't end up getting a chance. I might have a progress branch lying around somewhere, although I won't have a chance to look for a few days.

harrisonmg:
FWIW, I went ahead and tried this change out since I don't use folds; I found that it doesn't move the cursor while in visual mode.

theol0403:
Superseded by #885

PR #530, opened 2021-01-18
add <c-w><c-v>, <c-w><c-s>, and <c-w><c-o> commands

Yong-L:
with vim, you could execute window commands without letting go of
of the ctrl key which is much more convenient and fluid than having
to let go with your pinky.


asvetliakov (approved):

asvetliakov:
Thank you!

PR #536, opened 2021-01-24
Update README to include vim-sandwich color fix

### LINKED ISSUE #533: vim-sandwich underlining not displaying
In Neovim, [vim-sandwich](https://github.com/machakann/vim-sandwich) has a feature in which it underlines the text that you are about to surround. It also will highlight the background of the surrounding characters if you are attempting to change or delete them.

In VSCode, currently all that happens is the character your cursor is over gets underlined but nothing else will appear.

MaximoMachado:
Fixes #533

I encountered some issues with the highlighting that vim-sandwich does when using its operators. As such I thought it'd be a good idea to include the fix for the issue in the README as it is a similar solution to the one presented for the quick-scope plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,19 @@ highlight QuickScopeSecondary guifg='#5fffff' gui=underline ctermfg=81 cterm=und
 
 to your init.vim
 
+## VIM sandwich
+
+[vim-sandwich](https://github.com/machakann/vim-sandwich) in much the same way as quick-scope uses highlight groups that are ignored. To fix add
+
+```vim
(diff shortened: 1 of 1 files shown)


samvv:
I'm thinking this might be a better fit for an entry in the wiki? The README is getting quite large, and I'm not sure people can easily find the information they're looking for. I'll see if I can create some wiki boilerplate so that the information can go there.

Thanks for taking the time to look into this by the way!

samvv:
I've created an entry in the wiki: https://github.com/asvetliakov/vscode-neovim/wiki/sandwich.vim

I hope nobody is against this decision. If so, feel free to re-open this.

Thanks again for reporting!

PR #557, opened 2021-02-19
Fixes :e! to call `workbench.action.files.revert`

### LINKED ISSUE #556: Behavior of :e! (colon edit bang) inconsistent with with vim
Documentation:

https://github.com/asvetliakov/vscode-neovim/blob/6469c216f30c4bf29d5718f6d3fe99d517a33db0/README.md#L202

Source:

https://github.com/asvetliakov/vscode-neovim/blob/6469c216f30c4bf29d5718f6d3fe99d517a33db0/vim/vscode-file-commands.vim#L8

Why not revert current file to last saved version (`workbench.action.files.revert`) instead? That would be more consistent with normal vim behavior.

ogupte:
Closes #556.

`:e!` now calls `workbench.action.files.revert` instead of `workbench.action.files.openFile`.


theol0403 (approved):
I believe this is better. Besides, usually the DE's open file picker will not have vim bindings, so you are using the mouse anyways when you try to find a file.

theol0403:
However, I'd like to also update the readme with the changes. I'm not sure what the best way to do this is. Is there a way to edit a PR? Should I merge and then make a new commit? Should I merge this into my own PR? Or can you edit your PR?

PR #588, opened 2021-03-23
Add <C-s> as a keybinding to send to neovim

Yong-L:
Was running into issues of why <C-w><C-s> wasn't working and
realized that we have a set of keybindings we wish to send
to neovim

### DIFF
--- a/package.json
+++ b/package.json
@@ -667,6 +667,12 @@
             },
             {
                 "command": "vscode-neovim.send",
+                "key": "ctrl+s",
+                "when": "editorTextFocus && neovim.init && neovim.mode != insert && neovim.ctrlKeysNormal",
+                "args": "<C-s>"
+            },
+            {
+                "command": "vscode-neovim.send",
                 "key": "ctrl+x",
                 "when": "editorTextFocus && neovim.init && neovim.mode != insert && neovim.ctrlKeysNormal",
                 "args": "<C-x>"


Yong-L:
@samvv could you take a look when you have a chance? Thanks!

samvv:
All right I'll try tomorrow.
-------- Original Message --------

samvv:
LGTM

@asvetliakov what's our policy regarding merging? Do I wait for you to approve or is it good enough that the tests pass and that at least one contributor (being me in this case) approves? Also, I have a tendency to rebase everything ... Is that OK for you or do you prefer regular merges?

PS: I've just realised I already rebased two other pull requests without thinking, sorry.

asvetliakov:
@samvv I'm fine with merging when any contributor has approved

trkoch:
Doesn't this break native save on Windows desktops?

Yong-L:
yeah i think it would. same could be said about `ctrl+w` since that breaks closing windows as well

trkoch:
Granted. Also, there is the `useCtrlKeysForInsertMode` setting if users want to opt-out.

> same could be said about ctrl+w since that breaks closing windows as well

theol0403:
I don't think this should be merged - this plugin tries to not step too much on vscode, and C-S has no meaning in Vim. If you want to send this key to vim (and break the default saving shortcut), I think it should be done by the user. There are instructions in the readme.

theol0403:
I'm going to close this, because it is different from the other ctrl key bindings. The current objective is to only map the bindings that are used with vim by default, and not touch the rest. C-S has no use in vanilla vim, and so adding it to the plugin would have no effect other than to break vscode save. If people want to map it, they have to send it through.

Anyways, thanks for the PR 👍

Issue #605, opened 2021-04-13
% is not working after 100 lines.


unisons:
"%" is used to go to the next occurrence of matching pair under the cursor in normal mode.
It seems to work before 100 lines but after that it strangely stop to work.

My environment :
 - Ubuntu 18.04
 - Vscode 1.55.1
 - vscode-neovim 0.0.78
 - neovim v0.5.0-dev, neovim/bionic,now 0.5.0+ubuntu2+git202104112336-2e156a3b7-d569569c9-df97c72c7~ubuntu18.04.1 amd64 [installed]

You can reproduce the situation as below : 
1. Make new filer
2. Copy and paste the below pattern to become over 100 lines
```
{
// put empty line more than 25
}
```
3. Put the cursor on "{" and press "%".
4. If it moves to the matching pair of "}" then go to the next line and put the cursor on "{"
5. Repeat step 3 ~ 4.
6. Around over 100 line, it stop to move the cursor to matching pair.

But when it stops to work and put arrow down key, you might see the screen is refreshed and the cursor was moved to the next line of the matching pair as if it was moved to the correct position without refreshing.

unisons:
Before at 100 line, it seems that there is an event of refreshing as below:
```
2021-04-13T01:16:28.089Z CursorManager: Applying changed selection, kind: 3, WinId: 1004, cursor: [37, 0], isMultiSelection: false
2021-04-13T01:16:28.089Z CursorManager: Skipping event since neovim has same cursor pos

2021-04-13T01:16:57.707Z CursorManager: Received cursor update from neovim, gridId: 7, pos: [57, 0]
2021-04-13T01:16:57.708Z CursorManager: No pending document changes, gridId: 7
2021-04-13T01:16:57.708Z CursorManager: Updating cursor in editor: file:///xxxx.cpp, viewColumn: 1, pos: [57, 0]
2021-04-13T01:16:57.708Z CursorManager: Editor: file:///xxx.cpp, viewColumn: 1 setting cursor directly
2021-04-13T01:16:57.709Z CursorManager: SelectionChanged
2021-04-13T01:16:57.709Z CursorManager: Waiting for possible layout completion operation
2021-04-13T01:16:57.709Z CursorManager: Waiting for possible document change completion operation
2021-04-13T01:16:57.709Z CursorManager: Waiting done
[...]

PR #606, opened 2021-04-17
Update cursor position without grid_cursor_goto event.

unisons:
It is a fix of #605.


asvetliakov:
Thank you for the PR. I think would be better to just to use single `win_viewport` event and drop `grid_cursor_goto` completely then. It produces cursor jitter sometimes, but less evil than numerous workarounds 😕

asvetliakov:
Fixed by #649

PR #513, opened 2021-01-05
Refactor bindings and README

theol0403:
This is my work in progress for refactoring keybindings and readme. I hope you agree with the changes, and I will explain some of the keybindings at a later time. For now, I just want to show my progress, and see if you have any suggestions or feedback.

Would it be better to split this into two PRs? I put them in one because the readme needs to be changed to reflect the keybindings anyways.

**Todo:**

- [x] Add list/tree navigation (fixes #457)
- [x] Add file explorer actions
- [x] Make C-C binding toggleable (fixes #440)
- [x] Add navigation in peek editor (also fixes #604 and fixes #534)
- [x] Send home and end keys (fixes #448)
- [x] Refactor readme design and sentences
- [x] Convert readme large binding list into table
- [x] Document new bindings
- [x] Change ext name to "VSCode Neovim"???
- [x] Add suggested vscode settings for best keyboard navigation
- [x] Document new peek mode bindings
- [x] Add table of contents
- [ ] Provide opinionated vscode UI navigation bindings?
- [ ] Or just provide link to blog posts where people make the vscode ui very keyboard-centric
- [ ] Embrace the new wiki, move tips information over, add example navigation bindings there? Maybe after the PR.

PR #627, opened 2021-05-05
Use vim syntax highlighting in README

ian-h-chamberlain:
Basically just replace 
````
```
````
with
````
```vim
````
where applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ _Examples_:
 
 Produce linewise selection and show vscode commands (default binding)
 
-```
+```vim
 function! s:showCommands()
     let startLine = line("v")
     let endLine = line(".")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -381,24 +381,25 @@ To use VSCode command 'Increase/decrease current view size'
     <details>
     <summary>Copy this into init.vim</summary>
 
-        function! s:manageEditorSize(...)
```
theol0403, line 384: This breaks the vscode markdown renderer - revert this and I can merge.

theol0403 (changes requested):
Thanks for this! Just a note, it has already been implemented in #513, which will hopefully be merged soon.
However, I think it would be fine to merge now, so we don't need to wait.

theol0403:
Implemented by #513, thanks anyways for the PR!

PR #650, opened 2021-06-01
Fix cursor in visual mode

asvetliakov:
Regression introduced by #649


asvetliakov:
original commit was reverted instead

PR #655, opened 2021-06-05
Improve window keybindings

theol0403:
- organize `vscode-window-commands.vim` and comment
- fix missing visual-mode toggleEditorWidth
- add arrow-key variants of `<C-w> hjkl` and related (I use a split keyboard with an arrow key layer, as I don't have hjkl on my home row)
- add `<C-w> S-hjkl` bindings which call `workbench.action.moveActiveEditorGroupDown` etc
- send `C-arrow` and `S-arrow` from vscode to neovim
- update readme

I think the new `S-hjkl` bindings work quite well, I'm not sure if the vscode commands did not exist when you first made the bindings, or if there is some other reason its not ideal.


theol0403:
I just need an external approval to be able to merge my own PRs 😉

asvetliakov (approved):

Issue #175, opened 2020-02-10
:number and :relativenu not working


jamochl:
Hi! Just coming over from the popular Vim extension and so far things are mostly fantastic. However, there seems to be just a couple of quality of life things that would make the package better.

First is possibly tying :nu and :rnu to the line numbers of vscode. I like to change between relative number and number sometimes in the Vim extension, and giving neovim this ability would be great.

The most pressing issue however, is the inability to view :reg, :ju :ls, and other long line outputs, which does work with the popular vim extension. I've seen that the developer is planning to work on this feature, so i'm just putting this out to again express interest :)

Great work guys!

asvetliakov:
>First is possibly tying :nu and :rnu to the line numbers of vscode. I like to change between relative number and number sometimes in the Vim extension, and giving neovim this ability would be great.

As i know there is no default built-in command to change the line numbering, but you can do the trick by installing something like [settings-toggle](https://marketplace.visualstudio.com/items?itemName=Ho-Wan.setting-toggle), then binding command from neovim to call an extension command to toggle line numbers, e.g.:

```vim
command! Nu call VSCodeNotify('extension.toggle') " This is above extension command, assuming you configured it to toggle relative/non-relative line numbers
AlterCommand nu Nu
AlterCommand rnu Nu
```

>view :reg, :ju :ls, and other long line outputs

Related #78

lubaroli:
Related to this, as a new user coming from the Vim extension as well, it would be awesome if there was an option to automatic toggle the line numbering in normal (relative) and insert (absolute).

And yes, congrats for the extension, truly awesome!

asvetliakov:
@lubaroli Using the method above you can try to use `autocmd InsertEnter` and `autocmd InsertLeave`

theol0403:
Perhaps this functionality can instead be added to the plugin? Example implementation here:
https://github.com/VSCodeVim/Vim/pull/5010/

Advantages:
- does not rely on an external settings toggle plugin
- does not get out of sync, because rather than toggle, it specifies the mode it wants.
- works between editors?
- works without changing settings file?

It would also be doable to use this functionality to implement the commands requested by OP, and leave the auto-switching to the dedicated vim plugin you would use in bare vim.

asvetliakov:
Good points. Good way then is to support separate extension command which you can trigger from neovim's `autocmd` by `VSCodeExtensionNotify()`. Although vscode settings option is viable too

theol0403:
[...]

PR #656, opened 2021-06-05
Implement smart line numbers

theol0403:
Related: #175

- allow setting the editor line numbers from neovim
- provide example in the readme for smart line numbers

I'm a little stuck on what is the best approach for the vim commands. I don't know if there is a good way to override `set number`, `set rnu`, etc. Also, `:number` and `:nu` is already used for something else. Right now, the usage is `Number on/relative/off`, with aliases for `RelativeNumber` and `NoNumber`. I don't know if abbreviations are worth it, or if those aliases are worth it at all. 

I guess no one will use this exposed functionality outside of the smart line numbers autocmd, because they will just set it in vscode and forget it. So I'm not sure what is best.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ editor commands, making the best use of both editors.
     -   [Wildmenu completion](#wildmenu-completion)
     -   [Multiple cursors](#multiple-cursors)
     -   [Keyboard Quickfix](#keyboard-quickfix)
+    -   [Smart Line Numbers](#smart-line-numbers)
     -   [Invoking VSCode actions from neovim](#invoking-vscode-actions-from-neovim)
         -   [Examples](#examples)
 -   [⌨️ Bindings](#️-bindings)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -247,6 +248,23 @@ and add to your init.vim:
...
+
+To replicate this functionality, first make relative numbers default by adding `"editor.lineNumbers": "relative"` to
+your vscode `settings.json`. Then, add the following to your `init.vim` to automatically switch between the two modes:
+
+```vim
+augroup smartnnumbers
+  autocmd!
+  autocmd BufEnter,InsertLeave * RelativeNumber
+  autocmd BufLeave,InsertEnter * Number
+augroup END
```
justinmk, line 265: Instead of adding custom commands, vscode-neovim can listen to the `OptionSet` autocmd. However we probably don't want to actually have sign/fold/relnum columns actually enabled in the nvim server, so we would want to do something like:

```
augroup vscode-neovim
  autocmd!
  autocmd OptionSet * call VscodeNvimHandleOptionSet(deepcopy(v:event))
augroup END

func! VscodeNvimHandleOptionSet(ev) abort
    " set the relevant vscode settings...
    ...

    " then UNSET the nvim-side settings, because the sign/fold/relnum columns aren't useful to us.
    " use `noau` to ensure that OptionSet isn't circularly triggered.
    noau set nonumber
    noau set norelativenumber
    noau set signcolumn=no
    noau set foldcolumn=0
endfunc
```

PR #657, opened 2021-06-06
Paste from register in cmdline

theol0403:
- add extension command for pasting into cmdline
- add keybinds to paste into cmdline
- update when clause for other cmdline bindings to support more modes
- update readme

It looks like vscode does not provide api to know where the cursor is, so it will always paste at the end of the text box. Kind of annoying, but it still works pretty well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,16 +444,17 @@ Refer to vim manual for their use.
 
 Always enabled.
 
-| Key                             | Desription                                                |
-| ------------------------------- | --------------------------------------------------------- |
-| <kbd>C-h</kbd>                  | Delete one character left.                                |
-| <kbd>C-w</kbd>                  | Delete word left.                                         |
-| <kbd>C-u</kbd>                  | Clear line.                                               |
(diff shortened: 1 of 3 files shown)


theol0403:
I'm noticing a bug with this. When a register contains a linewise yank (ends with a linebreak), the cmdline interprets that as the enter key, and runs command without the chance to edit. 

A simple solution would be to trim the whitespace at the end, or just take out the linebreaks with regex, but I don't know how it would work when there are multiple lines in the register. In native vim, it squishes all the lines into one line, but keeps the linebreaks as a printable character.

What would be the easiest fix for this?

justinmk:
> A simple solution would be to trim the whitespace at the end, or just take out the linebreaks with regex, but I don't know how it would work when there are multiple lines in the register.

just trimming or joining is perfectly fine. multiline cmdline is rare, and more often unwanted.

theol0403:
This will be replaced by #908! Reminder to myself not to forget to update readme.

PR #658, opened 2021-06-06
Send selection when finding in files (#535)

### LINKED ISSUE #535: Convert selection to VS Code selection for "find in project"
First, thanks for this awesome extension, it works great.

I often like to select a range of text and then search for it. Would it be possible to make the selection turn into a real VS Code selection when I press `cmd+shift+f`? Any selection is auto-filled into the search input, and this would be very useful.

theol0403:
Fixes #535.

- generalize `openVSCodeCommandsInVisualMode` to `VSCodeNotifyVisual` and add as api
- **change `<C-P>` to `<C-S-P>` when passing to nvim**. It does not affect the user, but it seems to make more sense. I don't know if there is a reason this is bad.
- add C-S-F for "findInFiles"
- pass `<C-S-F>` to neovim
- organize `code-actions.vim`
- update readme


theol0403:
I made it so that it would convert a visual block mode selection into a visual mode selection. It changes the shape of the selection, but I think it's better than not working at all (at least the stop/end pos are correct).

asvetliakov (approved):

PR #585, opened 2021-03-22
Implement Ctrl-O for insert mode

QAston:
Fixes #181

PR #662, opened 2021-06-09
Remove c-o note (#585)

theol0403:
When rebasing #585 I forgot to remove the outdated note (which also was in the wrong place to start with). Oops.


trkoch:
@theol0403 Thanks for fixing and sorry I added it at the wrong place.

asvetliakov (approved):

PR #672, opened 2021-06-17
fix: go-to-file was not working

alexandermckay:
I am not sure if `revealDeclaration` has been deprecated but `revealDefinition` is a better fit to replicate the 'go-to-file` feature of Neovim.


alexandermckay:
I've just realised a mistake in the commit.

PR #673, opened 2021-06-17
Fix: go to file

alexandermckay:
`revealDeclaration` doesn't do anything when the cursor is over a file path. Whilst `revealDefinition` isn't perfect it still is closer to the native `gf` functionality.

### DIFF
--- a/vim/vscode-code-actions.vim
+++ b/vim/vscode-code-actions.vim
@@ -1,5 +1,3 @@
-
-
 function! s:vscodeFormat(...) abort
     if !a:0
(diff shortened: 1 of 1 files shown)


theol0403:
Why can't you use `gd` for this? Currently, `gd` is `revealDefinition` (and goes to file), and `gf` is `revealReclaration`. It is better in my opinion to have the option to choose between the two, than to have them both do the same thing.

This was introduced in #339, this PR basically undoes that 😕 

If you prefer it your way, you can just add it to your init.vim, but I am willing to consider reasons why this change would be better for everyone.

PR #676, opened 2021-06-21
Add notebook support

### LINKED ISSUE #210: Enable navigation inside jupyter cells
I was wondering whether there was a straightforward way to allow for navigation in normal mode inside jupyter code cells as provided by https://github.com/Microsoft/vscode-python? Right now, j and k only move the selection between cells.

### LINKED ISSUE #669: Enter binding for list.select takes precedence over default Notebook: Edit Cells
I noticed today that `enter` stopped functioning to edit Jupyter cells. The shortcut troubleshooter told me it was firing list.select for vscode-neovim, introduced in 0.82. I'm not sure what use case the keybind is supposed to be handling, but I imagine all Jupyter users will quickly discover this. I've changed the When condition to `listFocus && !inputFocus && !notebookCellListFocused` to workaround the issue.

https://github.com/asvetliakov/vscode-neovim/commit/8ee0c96a0e42525b898f85c29c38639a44ecdfa5#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R417-R419

theol0403:
Fixes #210 and fixes #669

This adds basic support for notebooks (jupyter/ipython). 

It allows the enter key to focus a cell for editing in normal mode, and allows the escape key to focus the cell list while in normal mode. While the cell list, the default vscode bindings work (j/k/dd/o/y/m/etc). Also, list navigation bindings already work from this plugin (gg/G/c-u/c-d).

The vscode bindings for the cell list are not very vim-like. However, I don't think vim bindings should be introduced by this plugin, because it would have to disrupt the default y/o/etc bindings. 

Instead, I will try to add some bindings to the wiki, after I rebuild the wiki and migrate stuff from readme (on my todo list). Ideas:
- `i` works like enter but direct to insert mode
- `o/O` insert cell below/above
- `zz` center view
- fold bindings to toggle output (which reminds me, fold bindings are also on my todo)
- some binding to switch between markdown and code
- copy/cut/paste

Other things which might be nice to implement:
- navigation in normal mode moves to the next editor

About the new `escape` binding, instead of adding to the when clause of the normal escape:
I wanted to do: 
`"when": "editorTextFocus && neovim.init && !markersNavigationVisible && !parameterHintsVisible && !referenceSearchVisible && !(notebookCellFocused && neovim.mode == normal)"`
but https://github.com/microsoft/vscode/issues/91473

However, since keybindings are evaluated bottom to top, I can take advantage of that and add an interception to `notebook.cell.quitEdit` when `notebookCellFocused && neovim.mode == normal`. This is not ideal but it's the cleanest I can think of.

### DIFF
--- a/package.json
+++ b/package.json
@@ -429,7 +429,7 @@
             {
                 "key": "enter",
                 "command": "list.select",
-                "when": "listFocus && !inputFocus"
+                "when": "listFocus && !inputFocus && !notebookCellListFocused"
             },
             {
                 "key": "g g",
@@ -464,7 +464,7 @@
             {
                 "key": "escape",
                 "command": "list.toggleKeyboardNavigation",
-                "when": "listFocus && inputFocus && listSupportsKeyboardNavigation"
+                "when": "listFocus && inputFocus && listSupportsKeyboardNavigation && !notebookCellListFocused"
             },
             {
                 "key": "r",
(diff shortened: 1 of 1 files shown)


theol0403:
Closed in favor of #680

Issue #613, opened 2021-04-22
Setting to open editors in insert mode by default


pokey:
Would you be interested in a PR that would add a setting for opening editors in insert mode by default, instead of normal mode? I would think I could just add a check to [this if statement](https://github.com/asvetliakov/vscode-neovim/blob/a4735e901cb154a91bd36643970abf4d953622a3/src/mode_manager.ts#L84-L87).

asvetliakov:
`autocmd WinEnter startinsert` or something like it. Try to search, there were similar questions/requests

pokey:
I'm guessing you're referring to #579.  Unfortunately, that doesn't work for me, because the call to `commands.executeCommand("vscode-neovim.escape")` causes VSCode to lose the current cursor position, which the autocmd doesn't restore.  Is there a solution that allows me to preserve the cursor position?

asvetliakov:
Right. Then an option to disable mode changing on editor change would be appreaciated

pokey:
Great; I'll take a swing!

pokey:
And to confirm, am I on the right track with 

> I would think I could just add a check to [this if statement](https://github.com/asvetliakov/vscode-neovim/blob/a4735e901cb154a91bd36643970abf4d953622a3/src/mode_manager.ts#L84-L87).

Would it be alright if I added a command to toggle the behaviour on and off?  I switch between keyboard and voice coding, and when I'm coding by voice I am mostly in insert mode, and by keyboard mostly normal mode

QAston:
I'm using something linke this in init.vim:

trkoch:
I think there are a number of features which could be useful related to this issue. Happy to create new issues, but first wanted to discuss it here.

1. I noticed the editor stays in insert mode while switching to another application (e.g. <Cmd+Tab>). However, the editor enters normal mode when changing tabs or splits. This seems inconsistent to me. Not sure, but I think I'd prefer it if the editor stayed in insert mode in both scenarios.

2. Sometimes I'd like to disable the bindings completely (including <Esc>), for instance when pairing with a non Vim user, or when in an editing scenario where using the mouse extensively makes sense. VsCodeVim has a setting disableExtension and a command for this. I used to bind the command to a toolbar item I could toggle quickly.

trkoch:
> I'm using something linke this in init.vim:
> 
> ```
> if exists('g:vscode')
>  " start in insert mode
>    au BufEnter * start
> endif
> ```

[...]

PR #716, opened 2021-07-31
Add config options to control the editor mode

### LINKED ISSUE #642: Focus jumps to latest position when 2 tabs are open side by side
Same issue and behaviour as #427 but for editors opened in side-by-side tabs

QAston:
defaultMode - What mode to start the editor in.
resetModeOnWindowChange - "If 'true'
    will reset windows to the defaultMode on window change

Fixes #642 #613

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,20 @@
                     ],
                     "description": "Log Level"
                 },
+                "vscode-neovim.defaultMode": {
+                    "type": "string",
+                    "default": "normal",
+                    "enum": [
+                        "insert",
+                        "normal"
+                    ],
+                    "description": "What mode to start the editor in."
+                },
+                "vscode-neovim.resetModeOnActiveEditorChange": {
+                    "type": "boolean",
+                    "default": true,
+                    "markdownDescription": "If 'true' will reset the mode of the activated editor to vscode-neovim.defaultMode."
+                },
                 "vscode-neovim.logOutputToConsole": {
                     "type": "boolean",
                     "default": false,
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -17,6 +17,7 @@ import {
 } from "vscode";
 
 import { Logger } from "./logger";
+import { ModeManager } from "./mode_manager";
 import { NeovimExtensionRequestProcessable, NeovimRedrawProcessable } from "./neovim_events_processable";
 import { calculateEditorColFromVimScreenCol, callAtomic, getNeovimCursorPosFromEditor } from "./utils";
 
@@ -36,6 +37,8 @@ const BUFFER_NAME_PREFIX = "__vscode_neovim__-";
 
 /**
  * Manages neovim buffers and windows and maps them to vscode editors & documents
+ *
+ * Buffered contents must not be used while insert mode because they're not kept in sync with the contents of the editor.
  */
 export class BufferManager implements Disposable, NeovimRedrawProcessable, NeovimExtensionRequestProcessable {
     private disposables: Disposable[] = [];
@@ -75,6 +78,12 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
     private editorTabConfiguration: WeakMap<TextEditor, { tabSize: number; insertSpaces: boolean }> = new WeakMap();
 
     /**
+     * Keep track if the layout or window need syncing before exiting insert mode
+     */
+    private activeWindowChangedInInsertMode = false;
+    private layoutChangedWhileInInsertMode = false;
+
+    /**
      * Buffer event delegate
      */
     public onBufferEvent?: (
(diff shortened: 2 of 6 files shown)


trkoch:
Any ideas why the test is failing? @QAston Is this ready for review?

package.json
```diff
@@ -135,6 +135,20 @@
...
+                    "enum": [
+                        "insert",
+                        "normal"
+                    ],
+                    "description": "What mode to start the editor in."
+                },
+                "vscode-neovim.resetModeOnWindowChange": {
+                    "type": "boolean",
+                    "default": true,
+                    "markdownDescription": "If 'true' will reset windows to the vscode-neovim.defaultMode on window change."
```
trkoch, line 150: Just for clarification, window change means when you leave Vscode (e.g. switch to browser)?

package.json
```diff
@@ -1260,6 +1274,7 @@
     },
     "scripts": {
         "lint": "eslint --ext .ts src",
+        "lint-fix": "eslint --fix --ext .ts src",
```
trkoch, line 1277: Personally I'd prefer unrelated changes like this to be in a separate PR.

package.json
QAston: It means switching to a different editor in vscode (perhaps editor change is a better name)

QAston:
@trkoch the test was failing on master - now the master is fixed I'm going to look into cleaning up and updating this PR (i've been happily using the changes for a while)

package.json
trkoch: Seems like Vscode terminology is in fact _editor_ (as indicated eg. by the command _Move Editor into Right Group_).

QAston: Yeah, I might have been thinking about nvim's window, but I agree editor is better.

QAston:
I've done some cleanups, but the ci still fails because it also fails on the master branch

trkoch:
@theol0403 Can you give us a hint?

justinmk:
> Fixes #642 #613

Adding an option doesn't fix anything. Either the issue is fixed or it isn't. Making users fiddle with options is a pretend fix. Especially with the amount of code added here.

Is it possible to do something reasonable for all users?

QAston:
People have different preferences, some people like to be in normal mode all the time. This preserves the current default.

justinmk:
So you can define a `VimEnter` autocmd that sets whatever mode you want. 

Why does there need to be a special option for this? The PR does not have a clear description. I glanced at the reference issues, they are too chatty. What is a 2-sentence explanation of the problem this PR solves?

trkoch:
> VimEnter autocmd

In general I'm of course in favor to leverage default vim features where possible. However, I tried this and if I remember it correctly, it introduces a lag.

> What is a 2-sentence explanation of the problem this PR solves?

* `defaultMode` allows to start in insert mode, without the lag introduced by `VimEnter` autocmd
* `resetModeOnActiveEditorChange` controls whether mode is restored to normal when switching editors ("buffers" in vim terms). If set to `false` and currently in insert mode, the editor will stay in insert mode after you change the editor.

`resetModeOnActiveEditorChange` can be useful if you plan to stay in insert mode for an extended period. Why you would do this of course depends on your usage and preferences. Personally I found myself wanting this when clicking/selecting/pasting across editors.

trkoch:
> Making users fiddle with options is a pretend fix.

I don't think it was meant as a clear-and-cut fix, rather a feature "fixing" a feature request (as it is often rather confusingly written).

> Especially with the amount of code added here.

@QAston Perhaps splitting both options into multiple PRs makes things easier to follow.

QAston:
@justinmk I've tried doing autocmds, using them caused other issues

@trkoch I think I'm going to work on my own plugin instead 😛

justinmk:
> @justinmk I've tried doing autocmds, using them caused other issues

Those are the issues we should focus on. Adding layers of special-case options bloats documentation, codebase, and avoids fixing actual problems. Entire purpose of having embedded Nvim is to use Nvim features.



> `resetModeOnActiveEditorChange` controls whether mode is restored to normal when switching editors ("buffers" in vim terms). If set to `false` and currently in insert mode, the editor will stay in insert mode after you change the editor.

That can be done with a `WinEnter` handler. I've just confirmed that this works:

    au WinEnter * startinsert

So `resetModeOnActiveEditorChange` absolutely is an anti-feature.

QAston:
@justinmk Yes, it works but makes the cursor break with multiple editors if i recall correctly. Feel free to take over the PR if you wish though.

trkoch:
> We should spend our time fixing those mysterious "issues". Not adding layers of special-case options.

I agree  – when meant as a general guideline. However, we should not forget that this is not Vim in its natural habitat and some concessions or compromises may be worth to consider when they improve overall usability from the users point of view.

> au WinEnter * startinsert

I see how this can replace `defaultMode`. From what I observe, `resetModeOnActiveEditorChange` controls behavior specific to this extension.

`true` (default):
* Editor A: Press `i`. Editor A is in insert mode.
* Switch to Editor B (e.g. via go to file): Editor B is in normal mode.
* Switch back to Editor B: Editor B is in normal mode.

`true` (default):
* Editor A: Press `i`. Editor A is in insert mode.
* Switch to Editor B (e.g. via go to file): Editor B is in normal mode.
* Switch back to Editor B: Editor B is in normal mode.

`false`:
* Editor A: Press `i`. Editor A is in insert mode.
* Switch to Editor B (e.g. via go to file): Editor B is in insert mode.
* Switch back to Editor B: Editor B is in insert mode.

Taking a step back here, I believe the underlying issue here is that in practice (I suspect) many users spend more time in insert mode to use features not available in normal/visual mode (mouse support, multi-cursor). Which is why setting the option above to `false` may be considered helpful by some.

justinmk:
> #794, strikingly one of the most popular feature request.

+1 for #794 , it's more generalized.

> I see how this can replace `defaultMode`. From what I observe, `resetModeOnActiveEditorChange` controls behavior specific to this extension.

which parts of your table are not achieved by the `WinEnter` autocmd?

theol0403:
Replaced by https://github.com/vscode-neovim/vscode-neovim/pull/1050 and potentially https://github.com/vscode-neovim/vscode-neovim/pull/1050. 

Use `au BufEnter * stopinsert` to leave insert mode when switching editors, and use `au BufEnter * startinsert` to enter insert mode when opening vscode or switching editors.

PR #334, opened 2020-08-31
Refactor

asvetliakov:
Refactor/rewrite almost all code. Simplicities and fixes many things related to buffer/window/cursor synchronisation

fixes #53 , fixes #187 , fixes #220 , fixes #223, fixes #202, fixes #78, fixes #226 

Fixes tab related problems, fixes: #275, fixes #239, fixes #264, fixes #167, fixes #100, fixes #152 , fixes #289 

Fixes macro recording problems: fixes #207 

Fixes some command line problems: fixes #155, fixes #288

Fixes some desync issues: fixes #312 

Ohter fixes: fixes #331 , fixes #300 , fixes #296

PR #723, opened 2021-08-11
Fix vscode-neovim.mouseSelectionStartVisualMode

cspotcode:
It looks like #334 accidentally ignored `vscode-neovim.mouseSelectionStartVisualMode`.

Explained here: https://github.com/asvetliakov/vscode-neovim/issues/395#issuecomment-897005222

This PR re-enables that setting.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -41,7 +41,7 @@
                         "regexp": "Compiling.*?|Compilation .*?starting"
                     },
                     "endsPattern": {
-                        "regexp": "webpack \\d+\\.\\d+\\.\\d+ compiled|Compilation .*?finished"
+                        "regexp": "Compiled .*?successfully|Compilation .*?finished"
                     }
                 }
             }
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -41,7 +41,7 @@
                         "regexp": "Compiling.*?|Compilation .*?starting"
                     },
                     "endsPattern": {
-                        "regexp": "Compiled .*?successfully|Compilation .*?finished"
+                        "regexp": "webpack \\d+\\.\\d+\\.\\d+ compiled|Compilation .*?finished"
                     }
                 }
             }
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -205,7 +205,7 @@ export class MainController implements vscode.Disposable {
             this.bufferManager,
             this.changeManager,
             {
-                mouseSelectionEnabled: false,
+                mouseSelectionEnabled: this.settings.mouseSelection,
             },
         );
         this.disposables.push(this.cursorManager);


yogeshdhamija:
Confirmed working on my machine 👍🏽  ❤️

yogeshdhamija (approved):

trkoch:
I was also curious about that `mouseSelectionStartVisualMode` setting. Good catch!

sky0hunter:
It is working partially for me; The selection does translate to the neovim selection, but on some files the selection always starts from the end of the file, or instead of starting a new selection it uses the current cursor position as starting point.

searleser97:
> It is working partially for me; The selection does translate to the neovim selection, but on some files the selection always starts from the end of the file, or instead of starting a new selection it uses the current cursor position as starting point.

I have the same problem

trkoch:
I noticed double click selects a word and the following whitespace (including the new line character).

.vscode/tasks.json
```diff
@@ -41,7 +41,7 @@
                         "regexp": "Compiling.*?|Compilation .*?starting"
                     },
                     "endsPattern": {
-                        "regexp": "Compiled .*?successfully|Compilation .*?finished"
+                        "regexp": "webpack \\d+\\.\\d+\\.\\d+ compiled|Compilation .*?finished"
```
trkoch, line 44: Why was this change necessary? Can you elaborate on this?

auderer: this seems unrelated to me, @cspotcode can you remove this from the PR?

cspotcode: I don't remember entirely, but I think it was in response to some changes to webpack's logging.  Rather than logging "Compiled * successfully" it was logging "webpack <version #> compiled."  The regexp required an update to match.

JimmyMcBride:
> It is working partially for me; The selection does translate to the neovim selection, but on some files the selection always starts from the end of the file, or instead of starting a new selection it uses the current cursor position as starting point.

Same issue. It seems the more you highlight, the more it selects in the opposite direction. It also includes the block on the end. I could live with it including the block on the end (even though it doesn't show it highlighted) but having it select more characters from the opposite direction is kind of breaking.

auderer:
> > It is working partially for me; The selection does translate to the neovim selection, but on some files the selection always starts from the end of the file, or instead of starting a new selection it uses the current cursor position as starting point.
> 
> Same issue. It seems the more you highlight, the more it selects in the opposite direction. It also includes the block on the end. I could live with it including the block on the end (even though it doesn't show it highlighted) but having it select more characters from the opposite direction is kind of breaking.

We should make a new issue for this. This PR just fixes the problem where the user's settings for mouseSelection are ignored, let's get this merged then discuss improving the mouse behavior.

theol0403 (approved):

theol0403:
Thanks for the investigation and PR! Yeah, please create another issue for this.

Issue #689, opened 2021-07-02
VIM Command Line ctrl+p only returns last command and ctrl+n doesn't work at all


vlrevolution:
What could be causing this?

theol0403:
Humm, I am also experiencing this with the latest neovim. I feel like the problem is relatively recent?

vlrevolution:
Yes, I tried with empty neovim config and still get the same, also with most VSCode plugins disabled. I am thinking recent VSCode updates have broken something about cmdline. Anyone else experience this?

mizagog:
i've had it for a while (even before NVIM 5), but i didn't have a go with an empty config, also i'm on Windows.

vlrevolution:
Yeah, I am also on windows.

sisrfeng:
+1

theol0403:
Fixed by #908

PR #730, opened 2021-08-16
Fix command line history up/down

### LINKED ISSUE #708: No access to command/search history
In normal vim you can press `/` or `:` to start a search or a command, and then you can press the up arrow to access previous commands.

This does not work with this extension, it just shows you an empty textbox. This is quite frustrating when your regex contains a mistake and you need to undo and apply it again. Unless you have the foresight to copy it before executing, you'll have to type it again.

joshuabrink1:
Fixes #708 and #689
 command line history navigation

## Problem
The current API calls are \<Up\> and \<Down\>
These calls find commands in history that can be prefixed by the current command in the input.
This causes the effect of the history getting stuck on the last command in history.

## Solution
To fix this \<S-Up\> and \<S-Down\> calls are used instead in order to get the typical history navigation.

### DIFF
--- a/src/command_line.ts
+++ b/src/command_line.ts
@@ -223,7 +223,7 @@ export class CommandLineController implements Disposable {
     };
 
     private onHistoryUp = async (): Promise<void> => {
-        await this.neovimClient.input("<Up>");
+        await this.neovimClient.input("<S-Up>");
         const res = await this.neovimClient.callFunction("getcmdline", []);
         if (res) {
             this.input.value = res;
@@ -232,7 +232,7 @@ export class CommandLineController implements Disposable {
     };
 
     private onHistoryDown = async (): Promise<void> => {
-        await this.neovimClient.input("<Down>");
+        await this.neovimClient.input("<S-Down>");
         const res = await this.neovimClient.callFunction("getcmdline", []);
         if (res) {
             this.input.value = res;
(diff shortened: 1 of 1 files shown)


harrisonmg (approved):
Works great!

One thing of note - in terminal neovim the `currentCommand` is set anytime the user edits the search term, so editing a command for history would cause the resulting term to be the `currentCommand` that gets restored upon returning to `historyLocation` 0.

I think this behavior discrepancy is acceptable for now, since we currently have no simple way of knowing whether `onChange` is called due to a user edit, or due to something like a history recall.

theol0403:
Thanks for the PR, and sorry for the delay. I should probably merge this to get a fix out, but I was hoping if there was a more elegant way to solve this. 

Two years ago, the current code worked fine and didn't have an issue navigating history. Arrow keys even used to work (thanks vscode). However, the extra code required here to keep track of the user input and insert it when the history is scrolled down seems like a smell of another bug. 

I don't understand why 
```js
        await this.neovimClient.input("<S-Down>");
        const res = await this.neovimClient.callFunction("getcmdline", []);
```
Does not return "" when the user scrolls all the way down.

Replacing `<Up>` with `<S-Up>` is a good solution, but ideally I'd like to avoid the extra code. Not sure why the nvim api isn't working as expected.

theol0403:
Alright, this is tricky. The problem is that whenever the input is changed, vscode clears and re-types the input into nvim. This means when you scroll up, the history item then gets typed into the current input. This both breaks the normal `<Up>` and scrolling all the way down. I'm looking into a better fix https://github.com/theol0403/vscode-neovim/tree/fix_command_history but there are some bugs.

PR #731, opened 2021-08-16
Fix command line open workspace files

### LINKED ISSUE #675: `:e`, `:split`, `:vsplit`, and `:tabe` with local file argument don't open file
Using latest version of vscode, Neovim, and this extension on Windows.

When using the `:e` or other file opening commands, I'm unable to just type a file in the current directory and open it.

It goes like this:
1. I type `:e myfile.txt`
2. Nothing happens

They do work with directories like $MYVIMRC or full paths like "C:/Users/username/...".

The only way I'm able to work with local files is if I just press `:e` then press enter, THEN search for the file in the drop-down menu.

I want to just type out `:e ` with the filename in one line  to quickly edit the file like in regular Vim or Neovim (which, the suggestions of files on the local directory do pop up, but even if I click on them NOTHING happens).

joshuabrink1:
Fixes #675

## Problem
Open commands (:e, :split, :vsplit, .etc) do not open workspace files.
This is because it tries to open the file with just the filename, but the path is also needed when opening documents in vscode.

## Solution
Resolve the workspace root folder path with the input path to get the files absolute path before opening.
This fix also works for relative and absolute paths to files outside the workspace.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -192,11 +192,12 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
                 if (fileName === "__vscode_new__") {
                     doc = await workspace.openTextDocument();
                 } else {
-                    const uri = await this.getInternalUri(fileName)
+                    const normalizedName = fileName.trim();
+                    const uri = await this.getInternalUri(normalizedName);
                     if (uri) {
                         doc = await workspace.openTextDocument(uri);
                     } else {
-                        doc = await workspace.openTextDocument(fileName.trim());
+                        doc = await workspace.openTextDocument(normalizedName);
                     }
                 }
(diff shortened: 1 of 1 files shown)


theol0403:
Thank you for the PR!

theol0403 (approved):

src/buffer_manager.ts
```diff
@@ -587,6 +594,17 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
         return false;
     }
 
+    private findPathFromFileName(name: string): string {
+        const rootFolderPath = workspace.workspaceFolders![0].uri.fsPath;
```
theol0403, line 598: I should have checked the tests - this line fails when there is no workspace folder (brand new vscode window).

PR #735, opened 2021-08-21
Fix cmd line tab complete

### LINKED ISSUE #215: edit command doesn't autocomplete files
Hello,

First of all, thanks for the awesome extension.

I'm trying to edit file in my home directory (my workspace points do different directory, I wanted to quickly edit a dot file in my home directory) using **edit** command. From vscode-neovim command palette I typed **e ~/**  and pressed tab. I didn't see any autocomplete results. Is it as expected? How do I use **edit** like use it in vim?

Thank you

joshuabrink1:
Fixes #215

## Problem
Command completion uses space to delimit and replace the current command with the incoming completion item.
This causes completion with paths & command symbols to be completely replace the current command. (if space is not used)

## Solution
Use path symbols and vim command symbols to delimit the current command, then compare the last delimited item with the incoming completion item.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -1,6 +1,7 @@
 import { debounce } from "lodash-es";
 import { Buffer, NeovimClient, Window } from "neovim";
 import { ATTACH } from "neovim/lib/api/Buffer";
+import path from "path";
 import {
     commands,
     Disposable,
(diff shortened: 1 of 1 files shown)


trkoch:
@joshuabrink1 Can you rebase and remove the reverts? It add unnecessary noise to the history and makes it hard to review.

PR #737, opened 2021-08-24
Fix i_Ctrl-a

Sejdens:
Fixes #402 and solves #283.

Current implementation was correctly getting the boundaries for the inserted text, but it was removing the front text first, shifting the position of the remaining text and thus incorrectly removing the text coming after it.

By simply removing the tail first, the shift happens only after the inserted text has been trimmed, so it works correctly now.
And it needed to be substracted 2 instead to 1 for the tail range because of the `colEnd` position being actually +1 from the last letter of the inserted text and the array range getting until -1 of the specified end index. So the difference is actually 2.

### DIFF
--- a/src/test/suite/insert-mode.test.ts
+++ b/src/test/suite/insert-mode.test.ts
@@ -529,6 +529,28 @@ describe("Insert mode and buffer syncronization", () => {
         );
     });
 
+    // currently fails
+    it.skip("Handles repeating last inserted text with newline", async () => {
+        const doc = await vscode.workspace.openTextDocument({ content: ["blah1 blah3"].join("\n") });
+        await vscode.window.showTextDocument(doc, vscode.ViewColumn.One);
+        await wait();
+
+        await sendVSCodeKeys("wiblah2\n");
+        await sendEscapeKey(1000);
+        await sendVSCodeKeys("A");
+        vscode.commands.executeCommand("vscode-neovim.ctrl-a-insert");
+        await wait();
+
+        await sendEscapeKey(1000);
+
+        await assertContent(
+            {
+                content: ["blah1 blah2", "blah3blah2"],
+            },
+            client,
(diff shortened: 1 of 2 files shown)


justinmk (approved):

justinmk:
Thanks, a test would be ideal. Also in the future please include the description in the commit message.

justinmk:
looks like there is already a failing test, so that just needs to be updated perhaps

trkoch:
Even better, amend the commit.

> Also in the future please include the description in the commit message.

Sejdens:
Hi! I apologize for my absence. I included the description in the commit message as requested, rebased to check if the problem was still happening (yes), and here I includet the test for before and after this fix.

Steps: 
1. Insert some text inside a line
2. Exit insert mode
3. Enter insert mode somewhere else
4. Press CTRL-A

## PROBLEM
i_ctrl-a problem

## FIXED
i_ctrl-a fixed

theol0403:
I added a new test, but it currently fails. However, it was failing before this PR, so I don't think it should block merging. However, if you can take a shot at fixing this, it would be great!

theol0403:
Awesome, that seems to work much better! It's always nice when the solution involves less magic code :p
Is this ready for merging? Thanks!

Sejdens:
Yes! I forgot to change the draft status, sorry!
If this gets merged, this would be my first actual-code contribution, so this is very exciting! (Even if it was just one line xd)
I'm glad I can contribute somehow

Sejdens:
I just removed the "currently fails" comment in the test. If you too agree, I think it's ready

Sejdens:
I think the failing test is not related to this PR.
Is there something else needed for this to get merged?

justinmk (approved):

theol0403:
Thanks for the PR!

Issue #270, opened 2020-05-23
compositeEscape(j,j) not removing "-- INSERT --" in master branch build for Windows. 


vettukal:
Steps to reproduce:
1. I built a vsix in Ubuntu and installed it on windows VSCode from master branch.
2. Enter insert mode using `i`
3. Press `j`,`j`. This enters normal mode.
4. But string  "-- INSERT --" still remains in status bar.

Note:
Bug not present in ubuntu installation of master branch
Bug not present in official release of plugin in windows.
I also successfully reproduced the bug in clean portable VSCode

 Keybindings.json
```
[
    {
        "command": "vscode-neovim.compositeEscape1",
        "key": "j",
        "when": "neovim.mode == insert",
        "args": "j"
    }
]
```

Setting.json
```
{
    "vscode-neovim.highlightGroups.unknownHighlight": {
    
    },
    "vscode-neovim.neovimExecutablePaths.win32": "C:\\Users\\****\\Downloads\\Neovim\\bin\\nvim.exe"
}
```

tfriedel:
I also have this problem when I use the escape key in the windows version.

asvetliakov:
It exists however from the insert mode, right?

tfriedel:
yes. I can enter insert mode and the string '-- INSERT --' appears. I leave the insert mode with escape and I'm obviously in normal mode, but the INSERT string remains forever.

rafmsou:
Same kind of problem on macOS too.

nwaywood:
I'm on macOS, when I press escape it removes `--INSERT--` string from the status bar, but when I press `jk` it does not

theol0403:
For me on linux, doing `i<esc>` toggles `--INSERT--` on and then back off properly, but if I type anything during insert mode like `iz<esc>` the `--INSERT--` remains stuck on the status bar (until I do `i<esc>` again).

Shatur95:
@theol0403, yes, same here.

brentluna:
Same here on OSX. I sometimes get weird buggy behaviour when existing insert with `jk`. Wondering if it's related

adamfeldman:
I'm also experiencing this, but more similar to #422 (which was closed as duplicate). 

I am _not_ using composite escape. `-- INSERT --` remains on the status bar when I exit NORMAL mode after having hit _any_ keys while in INSERT mode. If I just enter INSERT and then hit the escape key, the status bar updates as I expect.

[...]

PR #772, opened 2021-11-20
Fix jk not removing "--INSERT--"

zbw8388:
Solves #270.

It seems like there is a tiny buffer in the communication channel between neovim and vscode. As a result, there might be multiple `msg_showmode` messages a singular neovim notification (as seen in the screenshot below). This issue is likely to occur when there is an edit during the insert mode as neovim would issue several `msg_showmode` messages. However, current implementation only uses the first notification, which is the stale data indicating the mode string should still be `-- INSERT --`. This PR fixes this issue by using the latest message instead of the old one.
image


echaya:
can someone please review and approve this, thanks!

theol0403 (approved):

theol0403:
Thank you for the PR!

Issue #773, opened 2021-11-20
Back word `b` broken across wraps in markdown


milesfrain:
Going forward with the `w` command works fine across wraps, but going back with the `b` command causes the cursor to skip to the second character of the second wrapped line.

I also tried with both `:set wrap` and `:set nowrap`, but these have zero effect on the wrapping of `.md` files (wrapping is always enabled).

https://user-images.githubusercontent.com/3578681/142744332-f2a0fbe5-2c2b-45eb-a4e8-f22f5c62433e.mov

benlieb:
I'm also struggling to unwrap markdown...

Mostly wanting to be able to use this function with I've used for 15 years to easily navigate over wrapped lines:

```vimscript
noremap <silent> <Leader>w :call ToggleWrap()<CR>
function ToggleWrap()
 if &wrap
   echo "Wrap OFF"
   setlocal nowrap
   silent! iunmap <buffer> k
   silent! iunmap <buffer> j
   silent! iunmap <buffer> ^
   silent! iunmap <buffer> $

   silent! nunmap <buffer> <Up>
   silent! nunmap <buffer> <Down>
   silent! nunmap <buffer> <Home>
   silent! nunmap <buffer> <End>
   silent! iunmap <buffer> <Up>
   silent! iunmap <buffer> <Down>
   silent! iunmap <buffer> <Home>
   silent! iunmap <buffer> <End>
 else
   echo "Wrap ON"
   setlocal wrap linebreak nolist
   setlocal display+=lastline
   noremap  <buffer> <silent> k      gk
   noremap  <buffer> <silent> j      gj
   noremap  <buffer> <silent> ^      g^
   noremap  <buffer> <silent> $      g$

   noremap  <buffer> <silent> <Up>   gk
   noremap  <buffer> <silent> <Down> gj
   noremap  <buffer> <silent> <Home> g<Home>
   noremap  <buffer> <silent> <End>  g<End>
   inoremap <buffer> <silent> <Up>   <C-o>gk
   inoremap <buffer> <silent> <Down> <C-o>gj
   inoremap <buffer> <silent> <Home> <C-o>g<Home>
   inoremap <buffer> <silent> <End>  <C-o>g<End>
 endif
endfunction
```

Issue #573, opened 2021-03-11
Support for long lines


schiegl:
The extension ceases to work correctly on longer lines (most of the time). For example motion keys such as `w` don't work after a certain number of characters. Or highlighting a paragraph with `vip` doesn't highlight the entire paragraph.

This is rarely a problem in regular programming but often a problem when writing prose e.g. in latex or markdown. I know that I could hard-wrap the text (with the rewrap extension), but hard-wrapping doesn't work as well as soft-wrapping for me. It often doesn't work when editing or inserting text into an already existing paragraph. I always end up having to do it manually.

What is it that prevents the extension from working with longer lines? Can I increase a parameter to make it work in latex/markdown documents? I tried increasing the "Neovim Width" but that didn't help.

QAston:
You can change this setting: `"vscode-neovim.neovimWidth": 1000,` to enable longer lines

schiegl:
As mentioned in the original post. Changing this setting does not solve the problem. Increasing the width from 1000 to 5000 does not let me access/modify longer lines than before.

asvetliakov:
fyi you need to restart vscode after changing this setting

Try vim command `set wrap`. The ext is explicitly make it `nowrap`, but the reasons for it might be not relevant anymore. The visual mode highlight likely would be wrong though, but it's not fixable

schiegl:
> fyi you need to restart vscode after changing this setting

I restarted vscode but it did not help.

> Try vim command set wrap. The ext is explicitly make it nowrap, but the reasons for it might be not relevant anymore. The visual mode highlight likely would be wrong though, but it's not fixable

Adding `set wrap` to my `init.vim` didn't change anything. But setting it manually (after starting) did.

However, the behavior is still buggy. Sometimes I can navigate with `w` and `b` around the whole line, sometimes it won't let me go further. Same goes for `A`, `gg` and `G`. This seems to happen more when I modified the line e.g. removed/added a word. It seems to lose track of how long the line actually is (with `A`) and what the "first" soft-wrapped line is (with `gg`).

theol0403:
Should be fixed now other than visual highlight.

PR #775, opened 2021-11-21
Fix b on long lines

zbw8388:
Solves #773 and partially addresses #573.

Somehow a sequence of cursorMove commands is causing a bug in vscode. Removing them and directly setting the cursor location solves this issue. We don't need cursorMove vscode commands in this case anyways as less than one line move does not seem to get registered in jumplist.

Since the extension is getting highlights from redraw notifications, which is limited by window size, and neovim did not make the starting column of the current viewport in buffer available in the drawing notifications, the visual mode still displays incorrectly in vscode. However, it seems like the actual editing itself is working properly. There is no good fix to this other than implementing the display part of visual mode in the extension.

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -451,63 +451,24 @@ export class CursorManager
         const currCursor = editor.selection.active;
         const deltaLine = newLine - currCursor.line;
         let deltaChar = newCol - currCursor.character;
-        if (Math.abs(deltaLine) <= 1) {
-            this.logger.debug(`${LOG_PREFIX}: Editor: ${editorName} using cursorMove command`);
-            if (Math.abs(deltaLine) > 0) {
-                if (newCol !== currCursor.character) {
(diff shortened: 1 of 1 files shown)


justinmk:
The deleted code sure looks attractive :) 

Possible to write some sort of test? At least to exercise the codepath.

Rebasing will fix CI.

theol0403:
I've been using this for a while and it seems to work well. In addition, I think the codepath is already excersized by tests because it is a core part of the plugin. I think it also fixes a bug where lots of j/k up and down on the end of the line won't return to the same col after a bit.

theol0403 (approved):
There *might* be a slight performance decrease when holding j/k, but I'm having a lot of trouble being sure. I think it's acceptable.

theol0403:
Additionally, I think this fixes a problem where some jumps don't get registered in jumplist, because they come from a sequence of cursormove commands.

PR #750, opened 2021-10-08
Add backgroundColor on textDecorations

searleser97:
# What does this PR do?

It adds background color support when using `textDecorations`

# Evidence

See: https://github.com/justinmk/vim-sneak/pull/285

PR #784, opened 2021-12-09
Add backgroundColor on textDecorations (no fix-mouse-selection)

andyljones:
This is #750 but with the mouse selection stuff removed. It's @searleser97 's work but they seem to have gone quiet and I'm pretty keen to have this merged into master.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -41,7 +41,7 @@
                         "regexp": "Compiling.*?|Compilation .*?starting"
                     },
                     "endsPattern": {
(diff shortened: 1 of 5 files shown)


tranhl:
@trkoch Any chance we could get your eyes on this? Without this change setting `vscode-neovim.textDecorationsAtTop` to `true` is infeasible as the text decorations are hard to read when laid on top of the original code's text.

Screen Shot 2022-03-01 at 11 20 52 am

trkoch:
@andyljones With all the back-and-forth it would be helpful to rebase and tidy up this PR.

Issue #776, opened 2021-11-23
Cursor jumps when switching split-screen tabs 


hffmnn:
When editing in one tab and switching to another tab (e.g. via the mouse) the cursor position is somehow reset: 
- It always jumps to the highlighted line. This only happens, when one of the files is edited before switching. 
- It also happens when switching tabs via the native OS tab switch mechanism (in this case macOS: `Cmd-Shift-[`
- This does not happen when using the window shortcuts provided by the plugin e.g. `<C-W>l`
---

https://user-images.githubusercontent.com/352753/143090454-f43436c8-4c0c-4267-bf04-bf61d1d32675.mov

---
- As you can see I don't have any other plugins enabled
- VS Code Settings are completely empty besides this line: `"vscode-neovim.neovimExecutablePaths.darwin": "/usr/local/bin/nvim"`
- No custom `nvim` configuration  at all:
```bash
open ~/.config/nvim
The file ~/.config/nvim does not exist.
```
- `nvim` is installed via `brew`:

```bash
nvim --version
NVIM v0.5.1
Build type: Release
LuaJIT 2.1.0-beta3
[...]

PR #808, opened 2022-01-24
Fix cursor jumps when switching split-screen tabs

mussinbenarbia:
Fixes https://github.com/vscode-neovim/vscode-neovim/issues/776

Took a stab at this since I've had this happen to me as well a couple of times.
I did some tests with multiple files and tabs and it seems to work fine but I would appreciate it if someone else tried it as well 🙏  

https://user-images.githubusercontent.com/77260898/150796674-3651c6fa-5d2e-48ef-8212-818c235971c5.mov

It's my first time contributing to this amazing project so pardon me if I did something wrong. Please let me know if there is anything else that I can do for this PR.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -194,7 +194,7 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
                     if (fileName === "__vscode_new__") {
                         doc = await workspace.openTextDocument();
                     } else {
-                        const normalizedName = fileName.trim()
+                        const normalizedName = fileName.trim();
                         const filePath = this.findPathFromFileName(normalizedName);
                         doc = await workspace.openTextDocument(filePath);
                     }
@@ -448,6 +448,8 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
         this.logger.debug(
             `${LOG_PREFIX}: Setting active editor - viewColumn: ${activeEditor.viewColumn}, winId: ${winId}`,
         );
+        const cursor = getNeovimCursorPosFromEditor(activeEditor);
+        await this.client.request("nvim_win_set_cursor", [winId, cursor]);
         await this.client.request("nvim_set_current_win", [winId]);
     };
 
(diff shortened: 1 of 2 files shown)


QAston:
I've had a similar issue and solved it with https://github.com/vscode-neovim/vscode-neovim/pull/716, do you still have the issue if you apply that instead? It block sending positions from insert mode which was what was causing the issue I believe.

mussinbenarbia:
I tried running the extension off of the `feature/insert-mode-by-default` branch but it seems the issue is not resolved. Just making sure I did it correctly, I wiped `dist` and re-ran the extension while checking out your branch and this is the behavior of the cursor when jumping between tabs while in insert mode, same as in `master`:

https://user-images.githubusercontent.com/77260898/154809750-ddb7e26a-6d04-4163-8bbf-1e255a902798.mov

QAston:
Cool, looks like both patches are needed then, thanks

trkoch:
I checked this out. Seems to work fine.

Jakoviz:
> I checked this out. Seems to work fine.

Would you approve this? Hoping to get this fix soon.

Jakoviz (approved):
Works for me.

theol0403 (approved):

trkoch:
> I've had a similar issue and solved it with https://github.com/vscode-neovim/vscode-neovim/pull/716

For the record, reverting #808 and applying #716 I end up with the original issue (jumps) fixed and correct cursor position after leaving insert mode (no regression). We should probably decompose #716. It does many things in just one commit, but the original author seems to have lost interest.

theol0403:
Actually, this caused a regression.

Reproduce by opening an empty file and typing `abc<Esc>`. The cursor jumps to the beginning of the line, instead of staying at the end. Then, type `A12<CR>3<Esc>`. The cursor jumps to the middle of the first line instead of staying on the `3` in the bottom line.

theol0403:
Please repoen this in a new issue and we can investigate the cause or find another solve for #776

PR #821, opened 2022-02-15
WIP: fix mouse visual mode selection issues

QAston:
Viusual mouse selection mode doesn't work properly on master, the initial click doesn't appear where it's supposed to.

This patch fixes that.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -1,7 +1,8 @@
+import path from "path";
+
 import { debounce } from "lodash-es";
 import { Buffer, NeovimClient, Window } from "neovim";
 import { ATTACH } from "neovim/lib/api/Buffer";
-import path from "path";
 import {
     commands,
     Disposable,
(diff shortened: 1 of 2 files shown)


trkoch:
I gave this a go. This only has any effect with `mouseSelectionStartVisualMode`, right? I wonder, wouldn't it make sense to allow expanding an (existing) visual selection without this setting enabled?

QAston:
Yes, the changes here are only for that setting. Yeah, I think it could be the default, it's probably behind a flag because it's pretty broken.

trkoch:
@justinmk Could we leverage nvim API to expand selection of visual mode with mouse? I'm thinking of entering visual mode with `v` and then clicking somewhere. I'm not aware of a `click` event with line/column in Vscode API though.

trkoch:
Related https://github.com/vscode-neovim/vscode-neovim/pull/1043.

PR #824, opened 2022-02-20
feat: support outputting logs to the output channel

QAston:
Add option `vscode-neovim.logOutputToChannel`


theol0403:
@xiyaowong Investigate using vscode's new logging API. It might provide the ability to log to different places.

xiyaowong:
`TelemetryLogger`? Not feeling well.

theol0403:
I think I'm thinking of LogOutputChannel

xiyaowong:
> I think I'm thinking of LogOutputChannel

This is what was added by this PR. 👀

theol0403:
No, is it not just a regular OutputChannel?

theol0403:
I don't think it is readonly. I haven't looked too much into it, but I'd imagine that the reason it exists is that the vscode devs added some nice formatting for the different log levels. I was maybe hoping to refactor the entire logger to use the vscode logger, or at least parts of it.

xiyaowong:
code

theol0403:
There must be a way to change the log level... or else the API seems useless.

xiyaowong:
> There must be a way to change the log level... or else the API seems useless.

Not possible for now: https://github.com/microsoft/vscode/blob/a8b7239d0311d4915b57c837972baf4b01394491/src/vs/workbench/api/common/extHostOutput.ts#L24

The current project log is sufficient for now. It's just a log, after all.

xiyaowong:
Forgot one crucial fact - we cannot use 'output' because it is a textEditor which would lead to infinite output. @theol0403 🤯

PR #827, opened 2022-02-24
Support <c-t> as navigateBack

kflu:
C-t is used in vim as jumping back in tag stack. With LSP, this is
normally used for "navigating back". So adding this to vscode neovim for
the same convention.

However, as I tested, it does NOT currently work - it *can* send <c-t>
to neovim alright. But I got "E73: tag stack empty", as if c-t were
still bound to neovim's default c-t behavior (jump back in tag stack).
I expected it to have been mapped to `workbench.action.navigateBack`.
Any hint?

TEST:

```
npm install vsce
export PATH=`realpath node_modules/.bin`:$PATH
vsce package

code --install-extension vscode-neovim-0.0.83.vsix

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -841,6 +841,12 @@
             },
             {
                 "command": "vscode-neovim.send",
+                "key": "ctrl+t",
+                "when": "editorTextFocus && neovim.init && neovim.mode != insert && neovim.ctrlKeysNormal",
+                "args": "<C-t>"
+            },
+            {
+                "command": "vscode-neovim.send",
                 "key": "ctrl+j",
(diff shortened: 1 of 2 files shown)


theol0403:
1. you just mapped `xnoremap`, which is visual mode, there is actually two lines that you must map
2. not sure what the vim function you made does, but its probably not necessary
3. it would be great if you could just do this in your PR: https://github.com/kflu/vscode-neovim/blob/55d64509ac962623e218053ce281c85febf27bb9/vim/vscode-jumplist.vim#L1
Thanks!

PR #830, opened 2022-02-25
Improve efficiency applying small edits

### LINKED ISSUE #829: Flash of non syntax-highlighted text when making small edits from normal mode
This is a subtle effect I noticed with e.g. `dw` or `x`. Notice `const` and `JSDOM` flash in bright white for a very short time.

https://user-images.githubusercontent.com/73974/155815593-14df1b5a-aabe-49d2-81e6-ac51ad7dc61e.mov

After digging through some code, I believe this flash happens because the whole line is replaced no matter how small the edit was. This forces VS Code to parse again in order to apply syntax highlighting, which may take a noticeable time. In other words, the issue is VS Code needs to parse text of the line that was not even changed.

https://github.com/vscode-neovim/vscode-neovim/blob/b9def15f267d0e212c989981c883741de4256993/src/document_change_manager.ts#L476

Here is the frame in question (after `dw`, but before VS Code had the change to apply colors).

flash-no-syntax

trkoch:
This is my attempt to fix #829. The idea is to use `fast-diff` (already in use by other parts of extension) to determine the minimum set of changes to a line required to sync back VS Code to Neovim (after a buffer event). This list of operations is then applied to the text builder object provided by `editor.edit()` (part of Code Extension API).

### DIFF
--- a/src/document_change_manager.ts
+++ b/src/document_change_manager.ts
@@ -447,7 +447,7 @@ export class DocumentChangeManager implements Disposable, NeovimExtensionRequest
                     }
                     this.documentSkipVersionOnChange.set(doc, doc.version + 1);
 
-                    // const cursor = editor.selection.active;
+                    const cursorBefore = editor.selection.active;
                     const success = await editor.edit(
                         (builder) => {
                             for (const range of ranges) {
@@ -521,6 +521,12 @@ export class DocumentChangeManager implements Disposable, NeovimExtensionRequest
                     if (success) {
                         if (!editor.selection.anchor.isEqual(editor.selection.active)) {
                             editor.selections = [new Selection(editor.selection.active, editor.selection.active)];
+                        } else {
+                            // Some editor operations change cursor position. This confuses cursor
+                            // sync from Vim to Code (e.g. when cursor did not change in Vim but
+                            // changed in Code). Fix by forcing cursor position to stay the same
+                            // indepent of the diff operation in question.
+                            editor.selections = [new Selection(cursorBefore, cursorBefore)];
                         }
                         this.cursorAfterTextDocumentChange.set(editor.document, {
                             line: editor.selection.active.line,
--- a/src/document_change_manager.ts
+++ b/src/document_change_manager.ts
@@ -479,20 +479,16 @@ export class DocumentChangeManager implements Disposable, NeovimExtensionRequest
                                             range.start == range.end && range.newStart == range.newEnd;
 
                                         let editorOps;
-                                        let isSupportedOperation;
 
                                         if (changeSpansOneLineOnly) {
                                             editorOps = computeEditorOperationsFromDiff(diff(oldText, newText));
-
-                                            // Insert breaks cursor position after undo. Support only delete operation for now.
-                                            isSupportedOperation = editorOps.length == 1 && editorOps[0].op === -1;
                                         }
 
                                         // If supported, efficiently modify only part of line that has changed by
                                         // generating a diff and computing editor operations from it. This prevents
                                         // flashes of non syntax-highlighted text (e.g. when `x` or `cw`, only
                                         // remove a single char/the word).
-                                        if (editorOps && changeSpansOneLineOnly && isSupportedOperation) {
+                                        if (editorOps && changeSpansOneLineOnly) {
                                             applyEditorDiffOperations(builder, { editorOps, line: range.newStart });
                                         } else {
                                             builder.replace(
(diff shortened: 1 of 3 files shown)


trkoch:
🚧 Seems to break (at least change) cursor position after undo.

trkoch:
Managed to reduce number of failing tests (4 failing tests remaining). Not sure though those failures are related to my (semantic) changes. Something wrong with CI/CD setup? I needed to apply 94fd103 and 1bad56a to even get to the point that tests were run.

@theol0403 Any ideas? In wonder if there is a way to retrigger a GitHub action. That way we can run it on `main` and see if the build started to break without code changes.

theol0403:
This looks great! If the undo issue could be fixed, this would really help with performance. CI is still broken but might be doing better than before?

theol0403:
Tests are fixed!

theol0403:
It's possible tests are still failing because of the changes in this PR.

trkoch:
Thank you so much for fixing CI and giving this draft a close look!

Undo works fine now, though it may flash. I tried to limit my changes to a very narrow set of use cases so I don’t break too much. I’m quite happy with the result since flashes happen only rarely and I have not observed and (introduced) bugs yet impeding my workflow. 

By the way, this sync code is pretty crazy. Fascinated by what @asvetliakov has come up with. Eager to learn what those experimental upstream changes will allow us to do. 

Will check remaining test failures.

trkoch:
So I've taken a closer look at the first of two failing tests, namely `Navigating with tab indents`. It turns out the cursor position is wrong after deleting a tab character. I'm puzzled by the fact that removing tab characters is treated differently by `builder.delete()`. One possible mitigation is to exclude old/new text with tab characters from this optimization.

theol0403:
> Undo works fine now, though it may flash.

Sorry, I meant finding a workaround for having more efficient undos that doesn't involve disabling insertion. For example vim-surround `csb[` performs much better with this PR (and the insertion enabled). Something like calculating and setting the correct cursor position when undoing would be ideal.

> One possible mitigation is to exclude old/new text with tab characters from this optimization.

Ideally we can find a workaround that does not involve adding code that excludes this optimization from certain cases.

trkoch:
With the latest changes, I apparently fixed the [tests](https://github.com/vscode-neovim/vscode-neovim/runs/6283858072?check_suite_focus=true) `Navigating with tab indents` and `Insert mode racing with document change`. I seem to have hit another regression though with `Current mode is canceled when switching between editor tabs` now [failing](https://github.com/vscode-neovim/vscode-neovim/runs/6436469176?check_suite_focus=true). At least, we're now down to just one failing test. Progress!

trkoch:
> Ideally we can find a workaround that does not involve adding code that excludes this optimization from certain cases.

@theol0403 If restoring the cursor position really turns out to fix this, we can leave behind those exclusions and end up with net less code: Insert (e.g. undo) affecting only a single line now longer flashes with https://github.com/vscode-neovim/vscode-neovim/pull/830/commits/59810b809d0d37dcecad9855553940c6af9d540b.

trkoch:
All green now 😀! Marking this PR as ready for review.

> I seem to have hit another regression though with Current mode is canceled when switching between editor tabs now [failing](https://github.com/vscode-neovim/vscode-neovim/runs/6436469176?check_suite_focus=true).

Maybe that was just a flaky test?

theol0403:
Thanks for the PR! This is really impressive, and makes a big difference. Awesome that you manage to include all functionality with no regressions! I'm happy to merge now.

At the risk of being greedy, what would be the risk of enabling this change to work on multiple lines? I commented out the section that made it only run on one line and upon quick inspection it seems to work well, and greatly improves snappiness of certain things like multi-line `cs([`. 

However, I just ran into a problem where doing `dib` on a multi-line block only deletes the first line of content and then results in a buffer desync. 

I'd be happy to merge this as-is, but if you think it would not be too risky to let this be used for all editor changes, it would be great if you could look into that!

Thanks

theol0403 (approved):

trkoch:
> At the risk of being greedy, what would be the risk of enabling this change to work on multiple lines?

No worries… Not sure about the risk! One thing to keep in mind is that it may slow things down above a certain threshold. I suggest we tackle multi-line in a follow-up PR. I'd prefer to keep this small so that we can better pinpoint possible regressions. Perhaps we can release (*) in the meantime? That way we would receive feedback early.

> However, I just ran into a problem where doing dib on a multi-line block only deletes the first line of content and then results in a buffer desync.

Can you create a branch with your WIP commits and trigger a CI run? 

(*) Likewise, at the risk of being picky, and for sure completely off-topic I noticed we haven't updated the [CHANGELOG.md](https://github.com/vscode-neovim/vscode-neovim/blob/master/CHANGELOG.md) for a while. I think it would help users to keep track of changes without following here on GitHub, and gain trust this extension is not abandoned.

theol0403:
Sounds good on all counts, I've been meaning to do that.

PR #845, opened 2022-03-28
fix issue '{count}gt command not jumping to correct tab #670'

### LINKED ISSUE #670: {count}gt command not jumping to correct tab
When you use "5gt" for example and you have 5 tabs open, the last tab should open.

If you use 1gt, it should go to the first tab. If you say 3gt, it should go to the third.

Currently, if I press any count and gt, it acts as if I'm pressing gt 5 (next tab) times instead of jumping to the correct tab, it also wraps to the first tab. 

vscode version: 1.57.0
vscode neovim version: v0.0.82

yanshay:
Fixes #670

The fix uses VSCode workbench.action.openEditorAtIndex# where # stands for tab number.
It's normally used for CMD+# switching among tabs so I don't know if it works beyond 9, but even if limited to 9 it's more accurate than current behavior.

It's my first time contributing to this amazing project so pardon me if I did something wrong. Please let me know if there is anything else that I can do for this PR.

### DIFF
--- a/vim/vscode-tab-commands.vim
+++ b/vim/vscode-tab-commands.vim
@@ -6,6 +6,11 @@ function! s:switchEditor(...) abort
     endfor
 endfunction
 
+function! s:gotoEditor(...) abort
+    let count = a:1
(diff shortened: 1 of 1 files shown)


theol0403:
Thanks for the PR! Accidentally pushed to master instead of merging :(

PR #847, opened 2022-04-03
fix: improved `vscode-neovim.escape`'s when condition

wenfangdu:
When the find widget or notification center is visible, pressing `Esc` should close them instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -384,7 +384,7 @@
             {
                 "command": "vscode-neovim.escape",
                 "key": "Escape",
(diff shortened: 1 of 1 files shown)


trkoch:
You can press Shift-Escape.

wenfangdu:
@trkoch If you try `notifications.showList` while in normal mode, shift+escape won't close it.

justinmk:
I don't see any risk here. vscode-neovim isn't active in `findWidgetVisible || notificationCenterVisible`. So let's merge this.

justinmk (approved):

PR #855, opened 2022-04-07
<c-u>/<c-d> moves cursor

harrisonmg:
This continues the work in #528 and addresses #293.

halfPage

The only way I found to get reliable behavior that plays nice with folds is using `cursorMove` and `by: "wrappedLine"`, which gracefully moves over folds. This is because we can only infer folds form visible ranges, which doesn't account for the folds that might be revealed as we scroll. Also, `visibleRanges` only updates after the smooth scrolling animation completes, so it's not feasible to examine it after scrolling for complete information before moving the cursor.

The tradeoff with this approach is that each wrapped line causes the cursor position to be off by one when scrolling. I think this is the best we're going to get for now.

At a glance I'm not sure we can meaningfully unit test this as it's dependent on editor size, which isn't exposed to us.

This is my first work on a VSCode extension, and in JavaScript, so feedback is appreciated.

### DIFF
--- a/src/commands_controller.ts
+++ b/src/commands_controller.ts
@@ -107,7 +107,27 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
 
     /// SCROLL COMMANDS ///
     private scrollPage = (by: "page" | "halfPage", to: "up" | "down"): void => {
-        vscode.commands.executeCommand("editorScroll", { to, by, revealCursor: true });
+        const editor = vscode.window.activeTextEditor;
+        if (!editor) {
+            return;
+        }
+
+        const visibleRanges = editor.visibleRanges.slice();
+        const cursor = editor.selection.active;
+        const cursorVisible = visibleRanges.length > 0 &&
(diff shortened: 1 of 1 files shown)


justinmk:
CI should run if you rebase.

Need some sort of test. If if it's not a full test, need to at least exercise the code.

btw, Nvim HEAD has a WinScroll event, does that help here?

theol0403:
I've found that simply deleting everything the plugin does for this by doing:

```json
   {
        "args": "<C-D>",
        "command": "vscode-neovim.send",
        "key": "ctrl+d",
        "when": "editorTextFocus && neovim.init && neovim.mode != insert || neovim.recording"
    },
    {
        "args": "<C-U>",
        "command": "vscode-neovim.send",
        "key": "ctrl+u",
        "when": "editorTextFocus && neovim.init && neovim.mode != insert || neovim.recording"
    },
```
solves the problem, using nvim's default c-u behaviour. I'm not sure why the workaround was needed in the first place. Am I missing something?

theol0403:
Superseded by #885

Issue #683, opened 2021-06-26
Forward search `/` sometimes drops characters


milesfrain:
Characters are sometimes dropped when typing quickly for forward search (and reverse search). In the below video, I'm typing `string`, but only `sring` makes it (the `t` is dropped).

https://user-images.githubusercontent.com/3578681/123522110-1dd34b80-d670-11eb-8b21-15061f424583.mp4

I'm running with an empty `~/.config/nvim/init.vim` for this test case.

Version info:
- vscode `1.57.1`
- vscode nvim extension: `0.0.82`
- nvim: `v0.5.0-dev+1372-g056c464e8`

angshumanHalder:
I am having the same issue but I am facing it with "w". 

Version info:

vscode: 1.57.1
vscode nvim extension: 0.0.82
nvim:  v0.5.0-dev+24e0c16fd

tlc:
I'm seeing this as well.  I really have to slow my typing to get all chars recognized.

poleprogger:
Same here

vscode 1.58.0
vscode nvim extension: 0.0.82
nvim v0.5.0

> I am having the same issue but I am facing it with "w".
> 
> Version info:
> 
> vscode: 1.57.1
> vscode nvim extension: 0.0.82
> nvim: v0.5.0-dev+24e0c16fd

poleprogger:
I did some digging and found out that the issue happens in the file command_line_managers.ts. There is a method onCmdChange and here

`     private onCmdChange = async (e: string, complete: boolean): Promise<void> => {
        let keys = "<C-u>" + normalizeInputString(e);
        if (complete) {
            keys += "<C-e>";
        }

       await this.client.input(keys);
`

The last line calls the NeovimClient which then triggers another change in the input field and removes the character w and also the character c. Could be that this is not an issue of this extension but on the library it builds on....

poleprogger:
> I am having the same issue but I am facing it with "w".
> 
> Version info:
> 
> vscode: 1.57.1
> vscode nvim extension: 0.0.82
> nvim: v0.5.0-dev+24e0c16fd

I found what is causing the issue with dropping w and c characters @angshumanHalder . I removed the cmap mapping in my **init.vim** and this solved the issue for me.

samsonmking:
I was able to resolve this problem by setting the following in my `init.vim`:
```
set notimeout ttimeout ttimeoutlen=10
```

[...]

PR #857, opened 2022-04-08
Fix dropped characters in search

harrisonmg:
Addresses #683.

Now the quickpick input is only updated upon a `cmdline_show` event from Neovim if the update is expected due to a recent `<c-l>` input.

### DIFF
--- a/package.json
+++ b/package.json
@@ -309,6 +309,10 @@
                 "title": "Neovim: Delete char cmdline"
             },
             {
+                "command": "vscode-neovim.match-cursor-search",
+                "title": "Neovim: Match cursor search"
+            },
+            {
                 "command": "vscode-neovim.complete-selection-cmdline",
                 "title": "Neovim: Complete selection cmdline"
             },
@@ -783,6 +787,11 @@
             {
                 "command": "vscode-neovim.delete-char-left-cmdline",
                 "key": "ctrl+h",
+                "when": "neovim.mode == cmdline_normal || neovim.mode == cmdline_insert || neovim.mode == cmdline_replace"
+            },
+            {
+                "command": "vscode-neovim.match-cursor-search",
+                "key": "ctrl+l",
                 "when": "neovim.mode == cmdline_normal || neovim.mode == cmdline_insert || neovim.mode == cmdline_replace"
             },
             {
--- a/package.json
+++ b/package.json
@@ -309,8 +309,8 @@
                 "title": "Neovim: Delete char cmdline"
             },
             {
-                "command": "vscode-neovim.match-cursor-search",
-                "title": "Neovim: Match cursor search"
+                "command": "vscode-neovim.match-cursor-search-cmdline",
+                "title": "Neovim: Match cursor search cmdline"
             },
             {
                 "command": "vscode-neovim.complete-selection-cmdline",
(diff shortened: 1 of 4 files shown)


theol0403:
Rebase and tests should work.

package.json
```diff
@@ -308,6 +308,10 @@
                 "command": "vscode-neovim.delete-char-left-cmdline",
                 "title": "Neovim: Delete char cmdline"
             },
+            {
+                "command": "vscode-neovim.match-cursor-search",
```
theol0403, line 312: Thanks for the PR! Please rename to `vscode-neovim.match-cursor-search-cmdline`.

theol0403:
I'm a little confused here - this fixes #683 but also seems to introduce a new feature (`<C-l>`). Could you explain further?

harrisonmg:
As Christopher found [here](https://github.com/vscode-neovim/vscode-neovim/issues/683#issuecomment-1001893043), `<C-l>` is already supported, but its implementation causes #683.

My main goal was to fix the bug by improving the `<C-l>` implementation, as described [here](https://github.com/vscode-neovim/vscode-neovim/issues/683#issuecomment-1092207610). That way there would be no regression in features.

theol0403:
Thanks. Would you also be able to investigate the use of `C-l` in command mode (`%s/`)? Then it could be `vscode-neovim.match-cursor-cmdline` and work in command line mode, and possibly the bit of code that checks for search mode can be removed.

harrisonmg:
It seems to work with `%s/` and `'<,'>s/` if you disable the search mode check. It also does nothing as expected when typing other things in the command line. However, if you try it with just `s/` or there is no first character after `/` to match with, it adds an `i` character. Trying this is raw Neovim, it looks like that situation inserts the special `^L` character.

harrisonmg:
Further investigation shows that Neovim will insert a `\f` form feed character when there's no pattern to match, even if you're in a search or `%s`.

If I simply trim that character (which only seems to appear when there's no pattern or you're trying to `<c-l>` in an invalid context), everything works as expected.

src/test/suite/command-line.test.ts
```diff
@@ -25,30 +32,41 @@ describe("Command line", () => {
...
+        await sendVSCodeKeysAtomic("/1");
+        await wait(1000);
         await vscode.commands.executeCommand("vscode-neovim.match-cursor-search-cmdline");
+        await wait(100);
         await vscode.commands.executeCommand("vscode-neovim.match-cursor-search-cmdline");
-        await sendVSCodeKeys("\n");
-        assert.equal(await client.commandOutput("echo getreg('/')"), "1a");
+        await wait(100);
+        await vscode.commands.executeCommand("vscode-neovim.commit-cmdline");
+        assert.equal(await client.commandOutput("echo getreg('/')"), "1ab");
```
harrisonmg, line 42: I believe, when in search, <C-l> is supposed to match starting with the character under the cursor if nothing is provided.

That's what this test was supposed to check.

theol0403: Yeah I'm not sure, I tried in terminal nvim and got the same behavior.

harrisonmg: Do you mean that the original test did work for you?

theol0403:
While addressing a better alternative to #730, I pretty much re-implemented this PR in #908 to the point where I just decided to include the improvements here. I hope that's okay, and thanks a lot for the inspiration. Would it be ok for me to bring the tests over to that PR?

About the differences in `<C-l>` behaviour, strange, I can try to look into it more. Obviously the tests should represent desired behaviour. Sorry for stepping a bit on your PR with the test fix.

harrisonmg:
No worries, I'm happy to help! Feel free to take the tests.

PR #860, opened 2022-04-12
Fix command completion for path and command symbols

joshuabrink1:
## Problem
Command completion uses space to delimit and replace the current command with the incoming completion item.
This causes completion with paths & command symbols to be completely replaced with the incoming completion. (if space is not used)

## Solution
Use path symbols and vim command symbols to delimit the current command, then compare the last delimited item with the incoming completion item.

Fixes #860

### DIFF
--- a/src/command_line.ts
+++ b/src/command_line.ts
@@ -218,7 +218,19 @@ export class CommandLineController implements Disposable {
         if (!sel) {
             return;
         }
-        this.input.value = this.input.value.split(" ").slice(0, -1).concat(sel.label).join(" ");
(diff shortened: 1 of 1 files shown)


theol0403:
Rebase and tests should work.

theol0403 (approved):
Looks good! Ideally this logic should be delegated to nvim (ie completion handling should be delegated to nvim), but for now this works well. Thanks for the PR!

justinmk:
Nvim does have `:help getcompletion()`, wonder if it could be used here.

Issue #149, opened 2020-01-13
ctrl-e and ctrl-y won't update the cursor position when it is out of the editing window.


southp:
## Summary
First of all, sorry that I don't know the exact terms for what I'm trying to describe, so hopefully I can still get my points across.

When on a native neovim, we can scroll the editing window upwards/downwards for one line by ctrl-e/ctrl-y, without changing the cursor position. When the editing window goes beyond the cursor position, it will "push" the cursor along with the boundary. However, on vscode-neovim, the cursor will always stay at the same position, disregards the window. Please refer to the screencasts below for clearer demonstration.

I'd love to work on a patch if you can help me with some pointers :)

## Expected result
Here is what happened in a native neovim:
![demo-cursor-expected](https://user-images.githubusercontent.com/1842898/72231259-75edcd80-35f5-11ea-8db3-4516e9543a0e.gif)

## What happens instead
Here is what happened in vscode-neovim:
![demo-cursor-issue](https://user-images.githubusercontent.com/1842898/72231264-7ab28180-35f5-11ea-9acd-0f2948d10b51.gif)

asvetliakov:
Should be easy as [uncommenting/revisiting these lines](https://github.com/asvetliakov/vscode-neovim/blob/21ba0eb0ac483f45d73fb1333e46ae3178e4448d/src/controller.ts#L640-L708). The jumplist note is no longer relevant. There should be an option to toggle old behavior

nwaywood:
@asvetliakov since you say that uncommenting the lines to make this work will no longer break the jumplist, any chance they could be uncommented for the next release?

Shatur95:
@nwaywood, try it yourself to help us to improve the plugin. The file was moved [here](https://github.com/asvetliakov/vscode-neovim/blob/00d594e0ca7d3c15fe7ca84e52b878d38f735f2c/src/cursor_manager.ts#L444). Uncomment, build the extension and test. If it works - sand a PR.

nwaywood:
@Shatur95 thanks for the link to the new file location. I would be more than happy to test and submit a PR for this. However I have zero experience with developing VSCode extensions and the readme for the project doesn't have any instructions on how to build/run a local version of the extension. 

I am assuming that to build I would just need to run this https://github.com/asvetliakov/vscode-neovim/blob/master/package.json#L1073. But then how do I get vscode to use the local version?

Shatur95:
[...]

PR #868, opened 2022-04-23
feat: support extmark_overlay

windwp:
It allow highlight character different with current character on buffer
it working with hop and lightspeed.

lightspeed need  #869 to have same result like neovim


windwp:
@andyljones @theol0403

andyljones:
Fantastic! I pulled #868 and #869 into the same branch and ran the extension, and it all seems to work! Happy to pay out soon as it's merged.

theol0403:
Great to see this implemented! Funny timing as I was just starting to have the time to give it a shot, but it seems like you've done a great job of the PR :). I'll have to test this more thoroughly, but I've noticed some bugs:
- viewport height broken: I know you've mentioned this, but lightspeed scans much too far out out of the viewport, affecting performance, but also does not work when you scoll down beyond 200 lines. Hopefully the only solution is not to patch lightspeed, and instead provide a way for the native nvim api to know the vscode viewport.
- "instant mode" cursor, which comes after sneaking or f/t-ing, the fake cursor is hidden. Solution is to do `hi LightspeedCursor gui=reverse`
- eol labels: lightspeed supports doing `s<CR>` which then shows a label *after* the last character. These are currently hidden.
- I'll have to figure out how to reproduce, but I've found that sometimes the 1-character highlighting produces some weird overlay text

windwp:
about the viewport height issue. I think @asvetliakov  already try to sync visible range to vscode but it have problem. 
I already try and the scrolling is not good. I am not sure about jumplist.

 https://github.com/vscode-neovim/vscode-neovim/blob/master/src/cursor_manager.ts#L583
 https://github.com/vscode-neovim/vscode-neovim/blob/master/vim/vscode-neovim.vim#L89
 that why i create #869  and do a patch to lighspeed.
 https://gist.github.com/windwp/5fd492d8ace55fbe12bb7be67594bb60
It is a hacking way but it working well and don't break anything.

> I'll have to figure out how to reproduce, but I've found that sometimes the 1-character highlighting produces some weird 
> overlay text

if that character has the same character on buffer then it won't use overlay method. It highlight same as a normal character but with different color. It makes better performance with another plugin because some plugin highlight ext_mark for background text too so every character is overlay and it slow vscode.

- I will fix eol issue and add test.

justinmk:
> I think @asvetliakov already try to sync visible range to vscode but it have problem. I already try and the scrolling is not good. I am not sure about jumplist.
> 
> https://github.com/vscode-neovim/vscode-neovim/blob/master/src/cursor_manager.ts#L583 https://github.com/vscode-neovim/vscode-neovim/blob/master/vim/vscode-neovim.vim#L89

That's an old hack using winline(). Since then, Nvim RPC UI has added the `win_viewport` UI event: 

https://github.com/neovim/neovim/blob/f876a0eeb5af6f41acb187eb9442062c3984b26d/runtime/doc/ui.txt#L603

theol0403:
> That's an old hack using winline(). Since then, Nvim RPC UI has added the `win_viewport` UI event:
> 
> [neovim/neovim@`f876a0e`/runtime/doc/ui.txt#L603](https://github.com/neovim/neovim/blob/f876a0eeb5af6f41acb187eb9442062c3984b26d/runtime/doc/ui.txt#L603)

This should also fix #149. @windwp it would be awesome if you could implement this!

windwp:
I think vscode neovim already implement win_viewport.
https://github.com/vscode-neovim/vscode-neovim/blob/b9def15f267d0e212c989981c883741de4256993/src/highlight_manager.ts#L88-L99
https://github.com/vscode-neovim/vscode-neovim/blob/52ff62612d3bb9ae2a50ce00b9e068c5d175abda/src/cursor_manager.ts#L109

vscode know the viewport of neovim but that viewport is not same of current visible ranges on vscode.

neovim viewport height is 200 and cursor always on a center position . `set scrolloff=100`.
https://github.com/vscode-neovim/vscode-neovim/blob/master/vim/vscode-options.vim#L16

**vscode manage all scroll function** . Neovim doesn't have API to scroll and spam <c-e> <c-d> is not good.

https://github.com/vscode-neovim/vscode-neovim/blob/master/vim/vscode-neovim.vim#L89

andyljones:
> viewport height broken: I know you've mentioned this, but lightspeed scans much too far out out of the viewport, affecting performance, but also does not work when you scoll down beyond 200 lines. Hopefully the only solution is not to patch lightspeed, and instead provide a way for the native nvim api to know the vscode viewport.

Impressed that you spotted this - I was trying to get a minimal example of this behaviour earlier and was pretty puzzled.

trkoch:
> Fantastic! I pulled https://github.com/vscode-neovim/vscode-neovim/pull/868 and https://github.com/vscode-neovim/vscode-neovim/pull/869 into the same branch and ran the extension, and it all seems to work!

I gave this a try with `vim-sneak` and label mode enabled. The overlay itself seems to work now. Great! I noticed, however, the cursor is no longer placed on the first occurrence (as it is still is with label mode disabled). I need to e.g. press `<space>` to do so (or press `w` which moves from the found word onwards). Also, the first occurrence is highlighted in a less visible way than following occurrences. I know this is very early in progress, but wanted to share my findings with you.

windwp:
@trkoch  you should try to change highlight  color on vim-sneak label.

theol0403:
Rebase and tests should work now

Shatur:
@theol0403 Is there anything else blocking the merge?

package.json
```diff
@@ -1302,6 +1303,7 @@
         "glob": "^7.1.6",
         "mocha": "^8.3.2",
         "prettier": "^2.2.1",
+        "sinon": "^13.0.2",
```
justinmk, line 1306: can this avoided? sinon is pretty jank. And we've avoided it up to now.

windwp: ok

justinmk (approved):
nice! thanks for getting rid of sinon. Dependencies can really become a maintenance liability.

theol0403:
> @theol0403 Is there anything else blocking the merge?

Sorry for the delay. There are two things:

1. Avoidance of lightspeed patch. Ideally, this PR would include a solution that does not involve patching lightspeed. This is so that all plugins do not need to be patched like hop/sneak/leap. I was hoping @justinmk could provide a suggestion to https://github.com/vscode-neovim/vscode-neovim/pull/868#issuecomment-1111619159, which basically says that there exists an api for `nvim scrolling->vscode scrolling` but not `vscode scrolling -> nvim set scrolling`. It would be ideal if @windwp could contribute a solution that allows nvim and vscode scrolling to be synchronized, fixing jump plugins and also C-u/C-d scrolling.
2. Some bugs I haven't had time to report:
		- highlighting tabs cause heart symbol to appear as overlay: 
image
		- also, searching in lightspeed in a buffer with tabs cause hearts
		- vscode previous cursor does not disappear when lightspeed tries to hide it (minor)
		- once, I experienced that the grey wash did not disappear after lightspeed was done, and I had to make a new search for the grey to go away (minor)

If the strange heart issue is resolved, I can merge this.

windwp:
@theol0403  can you provide a step to get "heart char"

windwp:
it is the solution for scrolling,
https://github.com/vscode-neovim/vscode-neovim/blob/1442396107c7314af024e6fdd8832cc06dc0792e/src/cursor_manager.ts#L583
https://github.com/vscode-neovim/vscode-neovim/blob/master/vim/vscode-neovim.vim#L89
maybe it have another issue so @asvetliakov comment it.

 lightspeed doesn't need scrolling to work it only need to know the viewport of vscode because on the first key `s` it find a unique text  on screen, another jump target plugin doesn't have that feature.

so  i suggest to make a new option to allow scroll neovim by <c-e> or <c-y> . we can uncomment that code and see what happen.

theol0403:
> @theol0403 can you provide a step to get "heart char"

Copy this C file with tabs:

```c
static uint8_t notify_func(struct bt_conn *conn,
			   struct bt_gatt_subscribe_params *params)
{
	if (!data) 
		printk("[UNSUBSCRIBED]\n");
	printk("[NOTIFICATION] data %p length %u\n", data, length);
	return BT_GATT_ITER_CONTINUE;
}
```

Line select all to get:

image

My theme: Monokai Pro

theol0403:
Thanks for the fix! I found one more bug:
* eol on empty line: lightspeed shows labels when doing `s<CR>` on empty lines. However, label isn't shown on those empty lines in vscode.

I'll merge after this, we can look into the scrolling issues separately. I realized that I've been incorrectly assuming what the problem is, its less major than I thought. I had thought that lightspeed didn't work below line 200 at all, not that it just doesn't know about scrolling (but does know about position). 

I'd say the other significant problem is that lightspeed tries to match all 200 lines in the buffer, which reduces performance and accuracy of lightspeed. The solution is to reduce the nvim buffer size to the vscode window size and solve issues related to that - PRs welcome!

theol0403:
Looks like extmark tests fail on latest commit. Will merge once fixed.

theol0403:
Thank you for the PR! And thanks @andyljones for the incentive :)

Just a reminder that for optimum lightspeed UX:
- `hi LightspeedCursor gui=reverse` for fake cursor
- https://github.com/vscode-neovim/vscode-neovim/pull/869 and https://gist.github.com/windwp/5fd492d8ace55fbe12bb7be67594bb60 for proper viewport height for labeling 

@windwp your code is very well written and you have a great understanding of nvim and this plugin. Even after this PR, would you consider being a contributor? Thanks!

windwp:
@theol0403  Thank i will try to make a sync viewport on neovim currently it have issue with fold.

Maybe we can rewrite some of vim code to lua.  Create _G.vscode and vscode.api.

unTanana:
Hey, great work here :D One question though. Is there any way to disable this feature / highlighting? thank you.

windwp:
@theol0403  is this your email theol0403@gmail.com.
@unTanana it  only highlight when plugin call it

theol0403:
> @theol0403 is this your email [theol0403@gmail.com](mailto:theol0403@gmail.com). 
Yep! Thanks!

unTanana:
Hmmmm, this might be a stupid think to ask, is this the highlight in question?
image

Because it seems to be:
1. Activated when using visual mode for selecting multiple lines (which is fine I guess)
2. Not going away after I'm done selecting
3. Not going away after I'm done jumping with an easymotion plugin either.

If the issue I'm describing is unrelated to this change, I apologise

theol0403:
> Hmmmm, this might be a stupid think to ask, is this the highlight in question? image

Actually, I've noticed that too, but I didn't realize it was from this PR. It appears as though tabs are highlighted extra. Not sure why they aren't going away though.

unTanana:
I can confirm that downgrading fixes this, and also my issue with easymotion excessively triggering the output tab, covering my screen each time I want to do a jump

Muzzamil1:
Hi, thanks for adding support for the lightspeed plug-in.

1. There is a delay using the "f" and "F" keys.
2. With "s" and "S", the cursor won't jump for the first match (white highlighted match) until I press escape, which is kind of confusing sometimes as it is not sure whether we jumped to that place or not until press escape

kevinhwang91:
Sorry for my laziness to open a new issue. I post the issue I am encountering, looks like has some bugs for some plugins using virttext with overlay. 
Using https://github.com/kevinhwang91/nvim-fFHighlight, 
image

theol0403:
> Hi, thanks for adding support for the lightspeed plug-in.
> 
> 1. There is a delay using the "f" and "F" keys.
> 2. With "s" and "S", the cursor won't jump for the first match (white highlighted match) until I press escape, which is kind of confusing sometimes as it is not sure whether we jumped to that place or not until press escape 

https://github.com/vscode-neovim/vscode-neovim/pull/868#issuecomment-1131963354

asvetliakov:
Just in case this code is very old (from the first versions) and i'm not sure if it would work anymore.

theol0403:
> @theol0403 Thank i will try to make a sync viewport on neovim currently it have issue with fold.
> 
> Maybe we can rewrite some of vim code to lua. Create _G.vscode and vscode.api.

Sorry, I missed this. Awesome, thanks a lot! And yes those sound like great places to start!

PR #881, opened 2022-04-27
feat(messages): show more messages in Output

### LINKED ISSUE #880: show Nvim messages in Output channel
When something logs in neovim it shows in statusbar and that's all. I think they should be duplicated in Messages output in vscode.

justinmk:
## Problem:
The vscode-neovim Output channel has some weird behavior:
1. It auto-closes when it loses focus.
2. It clears itself (e.g. on `:messages`) instead of just appending.
3. It doesn't log most Nvim messages.

This makes the Output channel not very useful except for one-off
invocations `:messages`.

## Solution:
1. Always append.
2. Only *show* the Output channel, don't move the cursor there.
3. Don't try to be clever about clearing the Output channel or
   auto-hiding it.
   - This allows the user to keep the Output channel open and see
     results of sequential commands (like a REPL, as `:` is intended).
4. Print all messages (except `return_prompt`).

closes #880
ref #334

@asvetliakov any objections here?

### DIFF
--- a/src/multiline_messages_manager.ts
+++ b/src/multiline_messages_manager.ts
@@ -9,12 +9,9 @@ export class MutlilineMessagesManager implements Disposable, NeovimRedrawProcess
 
     private channel: OutputChannel;
 
-    private isDisplayed = false;
-
     public constructor(private logger: Logger) {
-        this.channel = window.createOutputChannel(`${EXT_NAME} Messages`);
+        this.channel = window.createOutputChannel(`${EXT_NAME}`);
         this.disposables.push(this.channel);
-        this.disposables.push(window.onDidChangeActiveTextEditor(this.onDidChangeActiveTextEditor));
     }
 
     public dispose(): void {
@@ -28,9 +25,6 @@ export class MutlilineMessagesManager implements Disposable, NeovimRedrawProcess
                     let str = "";
                     for (const [type, content, clear] of args as [string, [number, string][], boolean][]) {
                         if (type === "return_prompt") {
-                            continue;
-                        }
-                        if (type) {
                             continue;
                         }
                         if (clear) {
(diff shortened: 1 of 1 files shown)


theol0403:
Thank you! 

I believe @asvetliakov was trying to make it behave more like nvim's message output - once you move away from it, it is gone forever. Instead, this PR makes it more like a logging panel. I personally like the change, but I'm not sure if the auto-closing and clearing is "weird" per se. 

Also, I'm not sure if this is intended behavior, but the output panel does not automatically open when messages are logged.

justinmk:
> the output panel does not automatically open when messages are logged.

It does for multiline messges. You probably don't want that for single-line messages (which already show in statusbar).

theol0403 (approved):

harrisonmg:
> It does for multiline messges.
I find Output automatically opening exceedingly annoying, especially since it often happens while simply searching. How would you feel about a VSCode option for this?

justinmk:
> I find Output automatically opening exceedingly annoying, especially since it often happens while simply searching. How would you feel about a VSCode option for this?

When are you getting multiline messages during search?

theol0403:
> > I find Output automatically opening exceedingly annoying, especially since it often happens while simply searching. How would you feel about a VSCode option for this?
> 
> When are you getting multiline messages during search? 

I've actually run into this too. When doing a search and pressing `n`, the "no more matches found, resuming from start" prompt triggers the panel to open.

justinmk:
We can improve the behavior (maybe 3+ multilines instead of just 2), but an option doesn't really make sense. If you don't want to see output why did you run a command?

harrisonmg:
Here's an example, for what it's worth.
image
I think it actually produced 3 lines.

I only proposed an option as to not interfere with the workflows of others, and I generally don't use the Output functionality. EasyMotion also seems to produce 3 lines of output.

justinmk:
@harrisonmg can you try https://github.com/vscode-neovim/vscode-neovim/pull/902 or adjust it to get a sense of the exact number of lines that you are frequently seeing emitted by Nvim.

If you are seeing 3+ lines, maybe that logic should discard blank lines before making the comparison...

harrisonmg:
Upon further testing, `> 2` seems to work for me. I may have previously made a mistake loading builds.

justinmk:
To be clear, https://github.com/vscode-neovim/vscode-neovim/pull/902 is still needed, but it appears to solve your issue?

harrisonmg:
Correct

Issue #580, opened 2021-03-20
motion does not work properly in visual mode


champignoom:
Steps to reproduce:
1. Open a new file with neovim plugin enabled
2. type `iabc<Esc>yy99p` to insert 100 lines of `abc`
3. type `V`
4. type `M`, or `L`, or `<ctrl-f>`, or `<ctrl-d>`

Expected behavior:
Lines from line 2 to the middle or last line of the screen should be selected.

Actual behavior:
Only line 2 is selected, while the cursor is at the middle or the last line of the screen.

ssipos90:
+1, I use `shift+V` (visual line) with `ctrl+d` or `ctrl+u` for half page selection most of the times.

Only the viewport is moving, not the cursor.

theol0403:
It looks like this is because vscode intercepts these commands and does not update neovim after scrolling. The reason vscode intercepts these commands instead of delegating to neovim is, according to @asvetliakov, "cursor jitter" and other problems.

However, when I disable the interceptions, it actually seems to work alright.

Here is how you can do this:

```json
    {
        "key": "ctrl+d",
        "command": "-vscode-neovim.ctrl-d",
        "when": "editorTextFocus && neovim.ctrlKeysNormal && neovim.init && neovim.mode != 'insert'"
    },
    {
        "key": "ctrl+u",
        "command": "-vscode-neovim.ctrl-u",
        "when": "editorTextFocus && neovim.ctrlKeysNormal && neovim.init && neovim.mode != 'insert'"
    },
    {
        "key": "ctrl+f",
        "command": "-vscode-neovim.ctrl-f",
        "when": "editorTextFocus && neovim.ctrlKeysNormal && neovim.init && neovim.mode != 'insert'"
    },
    {
        "key": "ctrl+b",
        "command": "-vscode-neovim.ctrl-b",
        "when": "editorTextFocus && neovim.ctrlKeysNormal && neovim.init && neovim.mode != 'insert'"
    },
    {
        "command": "vscode-neovim.send",
        "key": "ctrl+d",
        "when": "editorTextFocus && neovim.init && neovim.mode != insert || neovim.recording",
        "args": "<C-D>"
    },
    {
        "command": "vscode-neovim.send",
        "key": "ctrl+u",
        "when": "editorTextFocus && neovim.init && neovim.mode != insert || neovim.recording",
        "args": "<C-U>"
    },
    {
        "command": "vscode-neovim.send",
        "key": "ctrl+f",
[...]

Issue #361, opened 2020-09-11
C-e and C-y modifications not working


nwaywood:
Firstly, thank you so much for this plugin! I just switched from vscode vim plugin and I am already in love!

I have the following lines in my `init.vim` to scroll the viewport faster:

```
nnoremap <C-e> 3<C-e>
nnoremap <C-y> 3<C-y>
```

However I can't get this to work with this plugin.

I have tried with and without the following snippet in `keybindings.json` and neither works:

```
  {
    "key": "ctrl+e",
    "command": "vscode-neovim.send",
    "when": "editorTextFocus && neovim.mode != insert",
    "args": "<C-e>"
  },
  {
    "key": "ctrl+y",
    "command": "vscode-neovim.send",
    "when": "editorTextFocus && neovim.mode != insert",
    "args": "<C-y>"
  },
```

theol0403:
~~This is because when you do `nnoremap <C-e> 3<C-e>` with `nnoremap`, you are bypassing the mapping of `<C-e>` that sends the command to vscode.~~ This would be the case if https://github.com/asvetliakov/vscode-neovim/blob/00d594e0ca7d3c15fe7ca84e52b878d38f735f2c/vim/vscode-scrolling.vim#L28 lines were not commented. 

Here is where it is actually defined:
https://github.com/asvetliakov/vscode-neovim/blob/00d594e0ca7d3c15fe7ca84e52b878d38f735f2c/src/commands_controller.ts#L34

Your best bet is to define your ctrl-e keymap inside vscode keybindings.json to a command that scrolls how much you want.

asvetliakov:
Yep, unfortunately sending scroll commands from neovim produces jitter (at least it did last time when i've checked)

nwaywood:
> It's possible to restore behaviour by disabling default ctrl+e/ctrl+y bindings and using `vscode-neovim.send` command instead and notifying back the extension from neovim. But the extension doesn't support count argument anyway in the handler.

I'm a little confused. Are you saying that it should be possible to do atm or not?

Because I also did try setting

`"vscode-neovim.useCtrlKeysForNormalMode": false,`

while these keybindings were active

```
 {
    "key": "ctrl+e",
    "command": "vscode-neovim.send",
    "when": "editorTextFocus && neovim.mode != insert",
    "args": "<C-e>"
  },
  {
    "key": "ctrl+y",
    "command": "vscode-neovim.send",
[...]

PR #885, opened 2022-04-28
fix(scrolling): fix c-u/c-y/c-f/c-b by deletion

### LINKED ISSUE #293: Page movements don't affect cursor

theol0403:
# Problem
Window scroll commands don't move the cursor and instead just scroll the window. This differs from nvim behavior. This is because the extension intercepts the keybindings and applies vscode commands instead. The solution from #528 and #855 simply improves the logic in the plugin to move the cursor. However, this adds complexity and increases the separation from nvim.

# Solution
Remove all custom code for cursor scrolling and allow nvim to handle it itself. However, I have found two problems:

- due to `scrolloff=100`, the scrolling commands behave strangely, and don't reach the top of the screen. c-d scrolls twice as far as it should. However, this can't be avoided according to https://github.com/vscode-neovim/vscode-neovim/blob/52ff62612d3bb9ae2a50ce00b9e068c5d175abda/src/main_controller.ts#L55. @justinmk can you think of another solution that will avoid the enforcing of scrolloff? 
https://github.com/vscode-neovim/vscode-neovim/blob/d11c2ed69ab9ab26b82bb6d892443f8ece1fe940/vim/vscode-options.vim#L16
- C-b behaves very strangely, even with scrolloff=0. Sometimes, it moves *down* a line. It never makes it to the top. 

Fixes #293, #580, #528, and #855.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,11 +119,6 @@
                     "default": false,
                     "markdownDescription": "Use neovim installed in WSL. If you enable this setting, specify the path to the neovim executable installed in WSL `neovimExecutablePaths.linux` setting"
                 },
-                "vscode-neovim.revealCursorScrollLine": {
-                    "type": "boolean",
-                    "default": false,
-                    "markdownDescription": "If 'true' reveals the cursor when scrolling by line and if it is outside view port"
-                },
                 "vscode-neovim.logPath": {
                     "type": "string",
                     "default": "",
--- a/package.json
+++ b/package.json
@@ -291,24 +291,8 @@
                 "title": "Neovim: Send key in cmdline"
             },
             {
-                "command": "vscode-neovim.ctrl-b",
-                "title": "Neovim: ctrl-b"
-            },
-            {
-                "command": "vscode-neovim.ctrl-d",
-                "title": "Neovim: ctrl-d"
-            },
-            {
                 "command": "vscode-neovim.ctrl-e",
                 "title": "Neovim: ctrl-e"
-            },
-            {
-                "command": "vscode-neovim.ctrl-f",
-                "title": "Neovim: ctrl-f"
-            },
-            {
-                "command": "vscode-neovim.ctrl-u",
-                "title": "Neovim: ctrl-u"
             },
             {
                 "command": "vscode-neovim.ctrl-y",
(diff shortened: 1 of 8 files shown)


theol0403:
I found an explanation for why neovim was not originally handling the scrolling. #361

> Yep, unfortunately sending scroll commands from neovim produces jitter (at least it did last time when i've checked). It's possible to restore behaviour by disabling default ctrl+e/ctrl+y bindings and using vscode-neovim.send command instead and notifying back the extension from neovim.

I don't notice any jitter - maybe its safe to use this now?

justinmk:
> // to not deal with screenrow positioning, we set height to high value and scrolloff to value / 2. so screenrow will be always constant 

I'm missing context here, but I wonder if this hack was done before `win_viewport` UI event was available?

theol0403:
> I'm missing context here, but I wonder if this hack was done before `win_viewport` UI event was added to Nvim?

I'm also quite lacking in knowledge, but this seems related to https://github.com/vscode-neovim/vscode-neovim/pull/868#issuecomment-1111619159 . I think the solution to one would be a solution to the other (as well as the solution to many other issues regarding viewport).

justinmk:
If it's possible to remove the hacks while making some small tradeoffs, I'd lean towards that. And then figure out what is needed from Neovim (or VSCode?) to resolve the tradeoffs.

theol0403:
Yeah, I'd say all that we need to merge this is for the `scrolloff` hack to be removed.

trkoch:
I'm having trouble following. Can you bring us up to date what the status of this PR is?

theol0403:
> I'm having trouble following. Can you bring us up to date what the status of this PR is?

Sorry. Here is what happened:

- created this PR to simplify code and not require special PRs
- realised there was fundamental issue with the viewport height and scrolloff being too high, causing the native bindings to not work
-  https://github.com/vscode-neovim/vscode-neovim/pull/919 was merged, which properly syncs viewport height
- I merged this PR
- It turns out that c-u/c-d was still broken, see https://github.com/vscode-neovim/vscode-neovim/pull/993
- Pending https://github.com/vscode-neovim/vscode-neovim/pull/993, I had to revert this PR

So, the current state is "waiting for https://github.com/vscode-neovim/vscode-neovim/pull/993, currently reverted".

Issue #706, opened 2021-07-26
<ctrl+r><ctrl+o> not a command


levelel:
The official vim doc about this combination be found at `:h i_ctrl-r_ctrl-o` This combination works fine in native Neovim but not in NeoVim - VSCode. Error message: "<ctrl+r><ctrl+o> not a command"

The reason I am using this is to do this surround trick mentioned in:

http://vimcasts.org/episodes/pasting-from-insert-mode/

which talks about a way to add paranthesis around a text. Do `ciw(<C-r><C-o>")<Esc>`, the benefit of this is that it can be repeated by pressing `.`

PR #886, opened 2022-04-29
refactor(insert): remove custom insert mode bindings

### LINKED ISSUE #652: Expression Register
Many thanks for this awesome extension .

When I wanted to use a variable's value by expression register, it reported a problem.

Given to that expression register is not fully supported, I used the way below, which still didn't work. But it works in NeoVim.
```shell
:let a=1
"=a
i
<^R>=
```
I don't know why

theol0403:
# Problem
Currently, two sets of insert mode bindings are used:

1. custom vscode bindings:
commands like `C-w`, `C-r`, `<C-o>` are connected to vscode commands. These differ from nvim functionality and add to code bloat.
2. macro recording passthrough bindings:
when a macro is being recorded, the plugin directly sends the commands to nvim. This then applies the actions.

The problem with only using the nvim bindings is that when you type in insert mode, and then try to use the nvim bindings, nvim is not aware of the text you just typed, causing issues.

# Solution
Make a generic "sync-send" command for all insert-mode bindings that syncs changes and then allows nvim to handle the commands. This seems to work well for C-w and others. This also lets the ext handle C-r and others. 

# Issues
~~The code in typing_manager was not designed for this use and should be refactored. This PR is quite broken and using C-r freezes input. However, this shows promise.~~

C-r now works properly and everything else seems to work well!

Fixes #652 and #706

### DIFF
--- a/src/document_change_manager.ts
+++ b/src/document_change_manager.ts
@@ -232,7 +232,7 @@ export class DocumentChangeManager implements Disposable, NeovimExtensionRequest
         await callAtomic(this.client, requests, this.logger, LOG_PREFIX);
     }
 
-    public async syncDotRepatWithNeovim(): Promise<void> {
+    public async syncDotRepeatWithNeovim(): Promise<void> {
         // dot-repeat executes last change across all buffers. So we'll create a temporary buffer & window,
         // replay last changes here to trick neovim and destroy it after
         if (!this.dotRepeatChange) {
--- a/src/typing_manager.ts
+++ b/src/typing_manager.ts
@@ -42,9 +42,7 @@ export class TypingManager implements Disposable {
         private changeManager: DocumentChangeManager,
     ) {
         this.registerType();
-        this.disposables.push(
-            commands.registerCommand("vscode-neovim.sync-send", (key) => this.onSyncSendCommand(key)),
-        );
+        this.disposables.push(commands.registerCommand("vscode-neovim.send", (key) => this.onSendCommand(key)));
         this.disposables.push(commands.registerCommand("vscode-neovim.escape", this.onEscapeKeyCommand));
         this.disposables.push(commands.registerCommand("vscode-neovim.paste-register", this.onPasteRegisterCommand));
         this.disposables.push(
@@ -127,20 +125,20 @@ export class TypingManager implements Disposable {
         }
     };
 
-    private onSyncSendCommand = async (key: string): Promise<void> => {
-        this.logger.debug(`${LOG_PREFIX}: Sync and send for: ${key}`);
+    private onSendCommand = async (key: string): Promise<void> => {
+        this.logger.debug(`${LOG_PREFIX}: Send for: ${key}`);
         if (this.modeManager.isInsertMode && !(await this.client.mode).blocking) {
             this.logger.debug(`${LOG_PREFIX}: Syncing buffers with neovim (${key})`);
             await this.changeManager.syncDocumentsWithNeovim();
             await this.changeManager.syncDotRepatWithNeovim();
+            const keys = normalizeInputString(this.pendingKeysAfterExit);
+            this.logger.debug(`${LOG_PREFIX}: Pending keys sent with ${key}: ${keys}`);
+            this.pendingKeysAfterExit = "";
+            await this.client.input(`${key}${keys}`);
         } else {
             this.isExitingInsertMode = false;
+            await this.client.input(`${key}`);
         }
-        const keys = normalizeInputString(this.pendingKeysAfterExit);
-        if (this.pendingKeysAfterExit !== "")
-            this.logger.debug(`${LOG_PREFIX}: Pending keys sent with ${key}: ${keys}`);
-        this.pendingKeysAfterExit = "";
-        await this.client.input(`${key}${keys}`);
     };
 
     private onEscapeKeyCommand = async (key = "<Esc>"): Promise<void> => {
@@ -148,12 +146,12 @@ export class TypingManager implements Disposable {
         // see https://github.com/asvetliakov/vscode-neovim/issues/324
         this.registerType();
         this.isExitingInsertMode = true;
-        await this.onSyncSendCommand(key);
+        await this.onSendCommand(key);
     };
 
     private onPasteRegisterCommand = async (): Promise<void> => {
         this.registerType();
-        await this.onSyncSendCommand("<C-r>");
+        await this.onSendCommand("<C-r>");
     };
 
     private handleCompositeEscapeFirstKey = async (key: string): Promise<void> => {
--- a/src/typing_manager.ts
+++ b/src/typing_manager.ts
@@ -41,10 +41,12 @@ export class TypingManager implements Disposable {
         private modeManager: ModeManager,
         private changeManager: DocumentChangeManager,
     ) {
-        this.typeHandlerDisposable = commands.registerTextEditorCommand("type", this.onVSCodeType);
+        this.registerType();
         this.disposables.push(
             commands.registerCommand("vscode-neovim.sync-send", (key) => this.onSyncSendCommand(key)),
         );
+        this.disposables.push(commands.registerCommand("vscode-neovim.escape", this.onEscapeKeyCommand));
+        this.disposables.push(commands.registerCommand("vscode-neovim.paste-register", this.onPasteRegisterCommand));
         this.disposables.push(
             commands.registerCommand("vscode-neovim.compositeEscape1", (key: string) =>
                 this.handleCompositeEscapeFirstKey(key),
@@ -63,6 +65,21 @@ export class TypingManager implements Disposable {
         this.disposables.forEach((d) => d.dispose());
     }
 
+    public registerType(): void {
+        if (!this.typeHandlerDisposable) {
+            this.logger.debug(`${LOG_PREFIX}: Enabling type handler`);
+            this.typeHandlerDisposable = commands.registerTextEditorCommand("type", this.onVSCodeType);
+        }
+    }
+
+    public disposeType(): void {
+        if (this.typeHandlerDisposable) {
+            this.logger.debug(`${LOG_PREFIX}: Disabling type handler`);
+            this.typeHandlerDisposable.dispose();
+            this.typeHandlerDisposable = undefined;
+        }
+    }
+
     private onModeChange = (): void => {
         if (this.modeManager.isInsertMode && this.typeHandlerDisposable && !this.modeManager.isRecordingInInsertMode) {
             this.pendingKeysAfterEnter = "";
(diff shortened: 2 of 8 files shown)


guyguy2001:
It seems like in vscode insiders you can run vscode-neovim from another thread: https://github.com/microsoft/vscode/issues/65876#issuecomment-1078827200
It might allow you to run Ctrl-R+whatever without any vscode keybindings

theol0403:
> It seems like in vscode insiders you can run vscode-neovim from another thread: [microsoft/vscode#65876 (comment)](https://github.com/microsoft/vscode/issues/65876#issuecomment-1078827200) It might allow you to run Ctrl-R+whatever without any vscode keybindings
> 
> EDIT: I was probably rash with this comment, at the end of the day I don't think that the `type` event lets you override ctrl hotkeys. But maybe I'm wrong

Wow, this is huge!! Can't believe I missed it. This doesn't necessarily mean something to C-r, this PR already tries to do it without vscode keybindings. And yes they still need to be sent to nvim. 

However, this opens up the possibility of using nvim during insert mode, which would
- remove dot repeat hack
- remove insert mode issues/desync
- allow `imap` in insert mode
- remove macro recording hacks

wenfangdu:
> It seems like in vscode insiders you can run vscode-neovim from another thread: [microsoft/vscode#65876 (comment)](https://github.com/microsoft/vscode/issues/65876#issuecomment-1078827200) It might allow you to run Ctrl-R+whatever without any vscode keybindings
> 
> EDIT: I was probably rash with this comment, at the end of the day I don't think that the `type` event lets you override ctrl hotkeys. But maybe I'm wrong

@theol0403 Btw, the latest stable vscode supports it too.

theol0403:
Could people please test this PR? Everything seems to be working now!

theol0403:
VSIX for people to test: [vscode-neovim.zip](https://github.com/vscode-neovim/vscode-neovim/files/9021502/vscode-neovim.zip)

theol0403:
One messy and unoptimized diff that seems to resolve the issues:

```diff
diff --git a/src/typing_manager.ts b/src/typing_manager.ts
index 66b1600..841608d 100644
--- a/src/typing_manager.ts
+++ b/src/typing_manager.ts
@@ -108,6 +108,11 @@ export class TypingManager implements Disposable {
         } else {
             this.client.input(normalizeInputString(type.text, !this.modeManager.isRecordingInInsertMode));
         }
+        if (this.modeManager.isInsertMode && this.typeHandlerDisposable) {
+            this.logger.debug(`${LOG_PREFIX}: Type registered in insert mode, unregistering`);
+            this.typeHandlerDisposable.dispose();
+            this.typeHandlerDisposable = undefined;
+        }
     };
 
     private onSyncSendCommand = async (key: string): Promise<void> => {
@@ -116,7 +121,9 @@ export class TypingManager implements Disposable {
             this.logger.debug(`${LOG_PREFIX}: Syncing buffers with neovim (${key})`);
             // rebind early to store fast pressed keys which may happen between sending changes to neovim and exiting insert mode
             // see https://github.com/asvetliakov/vscode-neovim/issues/324
-            if (!this.typeHandlerDisposable) {
+            if (!this.typeHandlerDisposable && ["<esc>", "<c-o>", "<c-r>"].includes(key.toLowerCase())) {
+                this.logger.debug(`${LOG_PREFIX}: Registering type in sync-send`);
+                if (key.toLowerCase() != "<c-r>") this.isExitingInsertMode = true;
                 this.typeHandlerDisposable = commands.registerTextEditorCommand("type", this.onVSCodeType);
             }
             // this.leaveMultipleCursorsForVisualMode = false;
@@ -130,7 +137,6 @@ export class TypingManager implements Disposable {
     };
 
     private onEscapeKeyCommand = async (): Promise<void> => {
-        if (this.modeManager.isInsertMode) this.isExitingInsertMode = true;
         await this.onSyncSendCommand("<Esc>");
     };
 ```

**Comment by theol0403** [comment]
Thanks, I think that's exactly what I was looking for! All I needed was to know if neovim expects input.

I'm not sure if #948 will ever be merged, I did it as a POC and to see if it is possible. It might be if optimizations can be made to improve cursor jitter.

**Comment by theol0403** [comment]
> There is `nvim_get_mode` api which is used in this extension in a couple places. During `i_ctrl-r` it returns:
> 
> ```
> n.request('nvim_get_mode')
> {'mode': 'i', 'blocking': True}
> ```
> 
> which at least indicates "blocking", though I don't know if that's enough for the purposes here.

Is this information available from the `mode_change` event as well? Then I could extract it from here:
https://github.com/theol0403/vscode-neovim/blob/565b37eb3abe5f2257535c532e69cd9fd9e3ff45/src/mode_manager.ts#L58

I'm having trouble finding documentation for `mode_change`.

**Comment by justinmk** [comment]
> @justinmk where else in the plugin is `nvim_get_mode` used?

`getCurrentNeovimMode` is driven by nvim_get_mode, though it's not obvious.

https://github.com/vscode-neovim/vscode-neovim/blob/f783274b4207d66047139113ad790b6ac25e0fd3/src/test/utils.ts#L87

**Comment by theol0403** [comment]
Refactored the logic, and fixed a lot of edge cases and regressions I added.

Now:
- `<C-r><Esc>` behaves as expected
- `<C-r><C-u>` behaves as expected
- `<C-r><C-r>` behaves as expected

Only rebinds when needed, and unbinds if no longer blocking. Only checks if blocking if in insert mode and not recording (async operation). 

Please test.

File: src/typing_manager.ts [hunk 1]
```diff
@@ -86,78 +100,60 @@ export class TypingManager implements Disposable {
...
-            this.isExitingInsertMode = true;
-            // rebind early to store fast pressed keys which may happen between sending changes to neovim and exiting insert mode
-            // see https://github.com/asvetliakov/vscode-neovim/issues/324
-            if (!this.typeHandlerDisposable) {
-                this.typeHandlerDisposable = commands.registerTextEditorCommand("type", this.onVSCodeType);
-            }
-            // this.leaveMultipleCursorsForVisualMode = false;
+    private onSyncSendCommand = async (key: string): Promise<void> => {
+        this.logger.debug(`${LOG_PREFIX}: Sync and send for: ${key}`);
+        if (this.modeManager.isInsertMode && !(await this.client.mode).blocking) {
```
theol0403, line 132: @justinmk this is the best way I could figure out to use `mode.blocking`. Since it is an async call to nvim, waiting for the result adds delay. I have tried to call await the bare minimum, like in this line where I rely on short-circuit logic. However, this await adds lag in the following situation:
- in insert mode
- Esc/C-o/C-r/C-u/C-w/etc is pressed
- it is important to check whether Esc/C-o will actually exit insert mode or require a resync
- it won't need to exit/resync if it is waiting for a C-r register selection

I don't think the lag is very significant, but basically any insert-mode binding needs to do this check. I think it is ok to merge, however. I was just wondering if there is a better way to obtain `mode.blocking` through UI events or similar, so that the information does not need to be requested on-demand.

theol0403:
Otherwise, tests pass and I think this is ready to merge.

src/typing_manager.ts
justinmk: > basically any insert-mode binding needs to do this check. 

My general opinion remains "who cares about insert-mode bindings?" They should be rare, and vscode-neovim should continue to posit that insert-mode is mostly owned by VSCode, not Nvim, and therefore use of insert-mode bindings are "use at your own peril".

> if there is a better way to obtain mode.blocking through UI events or similar, so that the information does not need to be requested on-demand

 I don't think this makes sense because "is blocking?" is inherently a just-in-time question. Receiving an asynchronous `blocking` flag in a UI event would be pointless and misleading.

justinmk (approved):

theol0403:
For reference:

- `vscode-neovim.send`: 
	- normal mode: send key to nvim
	- insert mode:
		- blocking: send key to nvim
		- not blocking: sync vscode buffer with neovim, then send key to nvim
- `vscode-neovim.send-blocking`:
	- like `vscode-neovim.send`
	- insert mode: register `type` event for the next keys (until unregistered by `type`)
- `type`:
	- if entering/exiting insert mode, store key to be replayed in destination mode
	- normal mode: send key to nvim
	- insert mode:
		- blocking: send key to nvim
		- not blocking: unregister and send key to vscode via `default:type`

src/document_change_manager.ts
```diff
@@ -232,7 +232,7 @@ export class DocumentChangeManager implements Disposable, NeovimExtensionRequest
         await callAtomic(this.client, requests, this.logger, LOG_PREFIX);
     }
 
-    public async syncDotRepatWithNeovim(): Promise<void> {
+    public async syncDotReapatWithNeovim(): Promise<void> {
```
justinmk, line 235: :)

```suggestion
    public async syncDotRepeatWithNeovim(): Promise<void> {
```

theol0403:
This PR makes insert mode bindings break dot-repeat. I don't think this is important, as insert mode bindings are rarely used and less often than with dot-repeat. I think it is worth it to merge anyway. If this bothers people, they can add some of the built in commands like C-w and C-u back into keybindings.json.

theol0403:
@justinmk is there a way to clear the dot-repeat register? The issue is that when you type 
`test test<C-w>` to get `test`, neovim just sees `<C-w>`. Then, when dot repeat is synced, neovim sees `<C-w>test`. If dot repeat could be cleared before syncing it, maybe the behavior would be more predictable. I should probably try to focus on more important things though 😅.

justinmk:
> @justinmk is there a way to clear the dot-repeat register?

not yet

theol0403:
I've been struggling because of these two issues:
- supporting native jk
  - we don't want to sync dot repeat every time j or k is pressed, as this causes lag. We should only sync when we exit insert mode
- supporting dot repeat
  - because of the current dot repeat hack, only syncing when exit insert mode causes more issues
  - syncing dot repeat every time sort of solves issue
- reducing code bloat
  - I came up with a solution that only syncs dot repeat when escaping, but properly replays actions. However, it is messy and adds hacky code. 

For now, I will not support the native jk and merge the least amount of code.

theol0403:
@BlueDrink9 thanks for your comments, but I think the alternate method is fundamentally broken. 

It is possible that some of the more recent prs fixed it.

But I think the problem is with lag. If you type a sentence rapidly with j or k in it, the j and k might be typed in the wrong location.

rennsax:
> This PR also provides experimental support for native compositeEscape!
> 
> To use:
> 
> 1. Checkout, build, and install this PR (soon to be released)
> 2. Add to keybindings.json (make sure to remove old compositeEscape bindings):
> 
> ```json
>   {
>     "command": "vscode-neovim.send-blocking",
>     "key": "k",
>     "when": "neovim.mode == insert && editorTextFocus",
>     "args": "k"
>   },
>   {
>     "command": "vscode-neovim.send-blocking",
>     "key": "j",
>     "when": "neovim.mode == insert && editorTextFocus",
>     "args": "j"
>   },
> ```
> 
> 3. use built in nvim map: `imap jk <esc>`
> 
> Please comment with any feedback!
> 
> Note: slight cursor jitter/lag is expected

I find that this mechanism is not a panacea. The cursor behavior is bizarre when I type backspace followed by `j`.

The cursor is already "at the end of the line".
image

image

PR #890, opened 2022-04-30
Fix tests

theol0403:


theol0403:
Looks like I'm having a bit more trouble than I expected - lots of the tests are broken although I can't reproduce myself.

theol0403:
@justinmk this is where my lack of experience shows - is there a better way to do [`025b7e0` (#890)](https://github.com/vscode-neovim/vscode-neovim/pull/890/commits/025b7e04c5789422c9001ffd06c425090401c5ee) ?

src/test/utils.ts
```diff
@@ -30,6 +30,7 @@ export async function wait(timeout = 1000): Promise<void> {
 }
 
 export async function attachTestNvimClient(): Promise<NeovimClient> {
+    process.env.NEOVIM_DEBUG = "1";
     const NV_HOST = process.env.NEOVIM_DEBUG_HOST || "127.0.0.1";
```
justinmk, line 34: > @justinmk this is where my lack of experience shows - is there a better way to do [`025b7e0` (#890)](https://github.com/vscode-neovim/vscode-neovim/pull/890/commits/025b7e04c5789422c9001ffd06c425090401c5ee) ?

better how? `NEOVIM_DEBUG` is something defined by this project (vscode-neovim), probably to change the behavior of the extension while it's being tested (usually not a good idea, but sometimes unavoidable).

I'd probably let the caller or environment define NEOVIM_DEBUG, in this case that would be the CI script or "global test setup". i'm not sure why attachTestNvimClient() is done over and over in every test file, instead of once in the global test setup.

theol0403: I see, I thought that was something defined by neovim and there was a better way of enabling it using node-client. Nevermind, thanks

Issue #563, opened 2021-03-02
More generic composite key escape method (e.g. for jj/jk/kj etc)


BlueDrink9:
## The problem

Currently the [composite key system](https://github.com/asvetliakov/vscode-neovim/blob/ff0082fc0234f069c8cf0983dab9d5820ad9869a/src/typing_manager.ts#L142) is a little strange. In order to handle the case where the two letters are the same (eg `jj`), the first composite key has to trigger an escape if it sees its bound key again. This means handling the case where they are different (eg `kj`) will always also handle the case where they are the same (i.e. `kk` as well as `kj`). Also, it only allows for one combination, so would not allow order-independent mapping of the `kj` chord (i.e. trigger on both `kj` and `jk`

## Proposed solution (WIP)

Have a single function to be mapped to, something like `compositeEscNotifyKeyPressed`. All keys within the chord, whether just `j` or both `k` and `j` should be mapped to this individually, so you might end up with something like:

```
{
    "key": "j",
    "args": "j",
    "command": "vscode-neovim.compositeEscNotifyKeyPressed",
    "when": "neovim.mode == insert && editorTextFocus"
},
{
    "key": "k",
    "args": "k",
    "command": "vscode-neovim.compositeEscNotifyKeyPressed",
    "when": "neovim.mode == insert && editorTextFocus"
}
```

This would be accompanied with a separate option (say, "`compositeEscKeyChords`"), a dictionary holding each chord you want to bind, for example:

```
{
    "k": "v",
    "v": "k",
    "j": "j"
    // Not supported with this algorithm because of the duplicate key:
    // "v": "k"
}
```

Then the callback function should store as a variable the pressed key (passed as an arg) if it is in `compositeEscKeyChords` as a key. If that variable already has something in it, and the current press is within the timeout, check to see if the current press is a value in the map for the stored previous key. If so, do the backspace and esc. If not, send the character.

What do you think about this setup? As far as I can see it would solve the problems above, and do so in a more elegant way than the current solution.

(I think the timeout should also be provided as an option, because people can get quite sensitive about their `timeoutlen` in vim.)

ChasKane:
#892 My stab at this; works for me but could be more robust.

ChasKane:
Ok I implemented a version of this, updated tests, etc. It's general enough to map any key, or just one if you include `escOnDoubleTap: true` in the args list in `keybindings.json`. Also added `timeoutLen` arg!

theol0403:
[...]

Issue #644, opened 2021-05-24
Composite Escape breaks macro recording


mikatpt:
Not sure if I'm doing something wrong, or misinterpreting. Hope somebody has some insight!

init.vim settings:
```
xmap gc <Plug>VSCodeCommentary
nmap gc <Plug>VSCodeCommentary
omap gc <Plug>VSCodeCommentary
nmap gcc <Plug>VSCodeCommentary
set clipboard=unnamedplus
:nnoremap <Space> :noh<CR>  
```
In VS Code, I'm using the suggested composite escape keybindings for `jj` and `jk`.

System: MacBook Pro, Big Sur 11.3

Steps to reproduce:
1. Open an editor and in normal mode, type `qdi`
2. Type any sequence including j or k
3. As soon as any letter besides j or k is typed, all j's and k's are erased.

theol0403:
Can reproduce

theol0403:
Related: #563

BlueDrink9:
After switching to the composite in  #886, I can no longer reproduce this issue. IMO #886 should be considered an official workaround and this can be closed. No reason to be using the old composite behaviour that I can see nowe

BlueDrink9:
After switching to the composite in  #886, I can no longer reproduce this issue. IMO #886 should be considered an official workaround and this can be closed. No reason to be using the old composite behaviour that I can see now

Issue #460, opened 2020-12-27
Configurable composite escape timeout


TheBB:
Would it be possible to make the 200ms hardcoded [here](https://github.com/asvetliakov/vscode-neovim/blob/ff0082fc0234f069c8cf0983dab9d5820ad9869a/src/typing_manager.ts#L142-L166) configurable? I seem to be typing a bit on the fast side and I often find myself unexpectedly in normal mode.

theol0403:
Check out this alternate solution! https://github.com/vscode-neovim/vscode-neovim/pull/886#issuecomment-1175189763

rennsax:
I approve this idea though https://github.com/vscode-neovim/vscode-neovim/pull/886#issuecomment-1175189763 is also a good solution.

Issue #330, opened 2020-08-28
Composite escape handler doesn't check if another non escape character was typed before 


asvetliakov:
@theol0403 that's another issue, second composite key handler doesn't check if you've typed another character between first & second escape keys. Should be easy to fix

_Originally posted by @asvetliakov in https://github.com/asvetliakov/vscode-neovim/issues/324#issuecomment-680853589_

chaorace:
Reposting my concern here since it has a dedicated issue now:

> I question if the current behavior is even consistent with vim's chording mechanism. Wouldn't it make more sense to reject a 'jk' key chord if it's interrupted instead of trying to salvage the interrupting keystrokes?

ChasKane:
Insert mode is handled by vscode. The only time vscode communicates with the neovim instance during insert mode is when a command is executed due to a key sequence registered in `keybindings.json`. Thus, the [composite escape key command code I modified](https://github.com/vscode-neovim/vscode-neovim/pull/892) is only called for the key bindings made by the user, not all key strokes in insert mode.

If the neovim extension isn't notified of all key strokes, I don't think there's a way for neovim to decide whether to issue an `esc` key based on whether keys have been pressed between `j` and `k`. the new compositeEscape code stores the last key that was pressed in insert mode which is registered in `keybindings.json` for the same command, not any other keystrokes.

As far as I know, there's no way to notify neovim of all keystrokes unless you register a command for every key, which would be inelegant and defeat the purpose of having vscode handle insert mode. I don't know if that would lead to performance problems either, or, frankly, if even registering a command for just 2 keys, as in the case of `jk`, would slow it down under any unusual circumstances.

theol0403:
> Insert mode is handled by vscode. The only time vscode communicates with the neovim instance during insert mode is when a command is executed due to a key sequence registered in `keybindings.json`. Thus, the [composite escape key command code I modified](https://github.com/vscode-neovim/vscode-neovim/pull/892) is only called for the key bindings made by the user, not all key strokes in insert mode.
> 
> If the neovim extension isn't notified of all key strokes, I don't think there's a way for neovim to decide whether to issue an `esc` key based on whether keys have been pressed between `j` and `k`. the new compositeEscape code stores the last key that was pressed in insert mode which is registered in `keybindings.json` for the same command, not any other keystrokes.
> 
[...]

PR #892, opened 2022-05-03
combine CompositeEscape funcs

ChasKane:
Allows any chord of any two characters to be mapped to `Esc`

Keymap args:
* `key` (required): specify which key is added to the list of keys that can be combined allows any chord of any two characters to be mapped to `Esc`.
* `timeoutLen` (optional, default `200`): specify timeout length in miliseconds. Specifically, it's the longest time between when the last key mapped to compositeEscape and the given key were pressed that will still trigger a remap to `Esc`.
* `escOnDoubleTap` (optional, default `false`): if `true`, specified key will trigger `Esc` if pressed 2x within `timeoutLen`.

### keybindings.json

```
    {
        "command": "vscode-neovim.compositeEscape",
        "key": "j",
        "when": "neovim.mode == insert && editorTextFocus",
        "args": {
            "key": "j",
            "timeoutLen": 300,
            "escOnDoubleTap": true
    },
    {
        "command": "vscode-neovim.compositeEscape",
        "key": "k",
        "when": "neovim.mode == insert && editorTextFocus",
        "args": "k"
    },
```

ISSUES:
* If you have, for example, j and k set as above, then typing `jk` in less than 300ms will trigger `Esc` but so will typing `jack`, and then `jac` would remain on screen. This is no different than the existing implementation and addressed reasonably by the addition of the `timeoutLen` arg.
* This implementation doesn't allow you to esc on `jk` and not `kj` also; if there's a need for this, lmk.

### DIFF
--- a/src/test/suite/composite-escape.test.ts
+++ b/src/test/suite/composite-escape.test.ts
@@ -17,7 +17,7 @@ describe("Composite escape key", () => {
     before(async () => {
         client = await attachTestNvimClient();
         doc = await vscode.workspace.openTextDocument({
-            content: "",
+            content: "abc",
         });
         await vscode.window.showTextDocument(doc);
         await wait(1000);
@@ -27,7 +27,10 @@ describe("Composite escape key", () => {
         await closeNvimClient(client);
     });
 
-    beforeEach(async () => await sendVSCodeKeys("S"));
+    beforeEach(async () => {
+        // reset content to `abc` and remain in insert mode
+        await sendVSCodeKeys("Sabc");
+    });
 
     it("escapes on jk and removes 'j'", async () => {
         await vscode.commands.executeCommand("vscode-neovim.compositeEscape", { key: "j" });
@@ -36,7 +39,7 @@ describe("Composite escape key", () => {
         await wait(69);
         await assertContent(
             {
-                content: [""],
+                content: ["abc"],
                 mode: "n",
             },
             client,
@@ -50,7 +53,7 @@ describe("Composite escape key", () => {
         await wait(50);
         await assertContent(
             {
-                content: [""],
+                content: ["abc"],
                 mode: "n",
             },
             client,
@@ -72,7 +75,7 @@ describe("Composite escape key", () => {
         await sendEscapeKey();
         await assertContent(
             {
-                content: ["jj"],
+                content: ["abcjj"],
             },
             client,
         );
@@ -92,7 +95,7 @@ describe("Composite escape key", () => {
         await sendEscapeKey();
         await assertContent(
             {
-                content: ["jk"],
+                content: ["abcjk"],
             },
             client,
         );
@@ -105,7 +108,7 @@ describe("Composite escape key", () => {
         await wait(50);
         await assertContent(
             {
-                content: [""],
+                content: ["abc"],
                 mode: "n",
             },
             client,
--- a/src/test/suite/composite-escape.test.ts
+++ b/src/test/suite/composite-escape.test.ts
@@ -27,12 +27,12 @@ describe("Composite escape key", () => {
         await closeNvimClient(client);
     });
 
-    beforeEach(async () => await sendVSCodeKeys("S"))
+    beforeEach(async () => await sendVSCodeKeys("S"));
 
     it("escapes on jk and removes 'j'", async () => {
-        await vscode.commands.executeCommand("vscode-neovim.compositeEscape", {key: 'j'});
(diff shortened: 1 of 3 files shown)


src/typing_manager.ts
```diff
@@ -34,6 +34,10 @@ export class TypingManager implements Disposable {
      * Timestamp when the first composite escape key was pressed. Using timestamp because timer may be delayed if the extension host is busy
      */
     private compositeEscapeFirstPressTimestamp?: number;
+    /**
+     * j or k, the first composite escape key that was pressed.
```
ChasKane, line 38: I'm gonna make this comment more generic

src/typing_manager.ts
```diff
@@ -160,28 +159,27 @@ export class TypingManager implements Disposable {
         await this.client.input(`<c-o>${keys}`);
     };
 
-    private handleCompositeEscapeFirstKey = async (key: string): Promise<void> => {
+    // whenever j or k is pressed,
+    //   if last key pressed was the other (j for k, k for j)
+    //      and pressed recently enough,
+    //   then backspace and simulate esc
+    private handleCompositeEscape = async (key: string): Promise<void> => {
```
ChasKane, line 166: I suspect I can add another param here for and store it as a private variable if it's not `undefined`. 1 how silly is it to pass a variable on every compositeEscape keypress if you only need it on startup? 2 not sure that's how the args list in `keybindings.json` works.

src/typing_manager.ts
```diff
@@ -160,28 +159,27 @@ export class TypingManager implements Disposable {
...
         const now = new Date().getTime();
-        if (this.compositeEscapeFirstPressTimestamp && now - this.compositeEscapeFirstPressTimestamp <= 200) {
-            // jj
+        if (this.compositeEscapeFirstPressKey &&
+            this.compositeEscapeFirstPressKey === (key === 'j' ? 'k' : 'j') &&
```
ChasKane, line 169: This line doesn't seem to do anything and I can't figure out why -- I only wanted `jk` and `kj`, not `jj` and `kk` too. nbd for me but surely strange.

theol0403:
Rebase and tests should work now

src/test/suite/composite-escape.test.ts
```diff
@@ -13,78 +13,99 @@ import {
...
-        await vscode.commands.executeCommand("vscode-neovim.compositeEscape1", "j");
+    it("escapes on jk and removes 'j'", async () => {
+        await vscode.commands.executeCommand("vscode-neovim.compositeEscape", { key: "j" });
+        await wait(50);
+        await vscode.commands.executeCommand("vscode-neovim.compositeEscape", { key: "k" });
+        await wait(69);
         await assertContent(
             {
-                mode: "i",
+                content: [""],
```
BlueDrink9, line 39: I wonder whether it would be more robust to have actual content here, to test you don't accidentally delete more than just one letter/delete one to the left/right instead? i.e. start with content abc, and ensure that's what's left at the end?

BlueDrink9:
Not that I have any experience with vscode dev but that looks like a solid PR to me, other than the testing issues which seem like part of the CI test framework is misbehaving?

src/test/suite/composite-escape.test.ts
ChasKane: good call -- do you know of a quick way to run just the tests in this file, or is modifying line 21 of in `index.ts` the only way?

`        glob("**/**.test.js", { cwd: testsRoot }, (err, files) => {`
to
`        glob("**/<filename>.test.js", { cwd: testsRoot }, (err, files) => {`

theol0403:
> This implementation doesn't allow you to esc on jk and not kj also; if there's a need for this, lmk.

If possible, mapping jk and kj to esc should be two different bindings, or at least an option to not do both ways.

theol0403:
Also, now that you are more familiar with the relevant code, would you be able to take a look at #644?

> If you have, for example, j and k set as above, then typing jk in less than 300ms will trigger Esc but so will typing jack, and then jac would remain on screen. This is no different than the existing implementation and addressed reasonably by the addition of the timeoutLen arg.

I think this is #330. If easy to fix, it would be ideal. Otherwise that's ok.

Finally, would you be able to update the README with the new api.

This PR closes #563 and #460.

Thanks for the PR!

BlueDrink9:
#886 being merged makes this PR redundant IMO. Thanks for your work bringing code and ideas to this issue! I think the PR can be closed

Issue #560, opened 2021-02-23
In normal mode, Chinese input methon will cause strange behavior 


adewalike:
In normal mode, normal behavior should not impact original text, while in `vscode-neovim` will impact original text. In below image, original text is `We need to “break window for emergency” here.`  cursor is at `f` word. And when type `j` it impact original text. In vim or neovim or vim in vscode don't have impact on original text.
![image](https://user-images.githubusercontent.com/26520465/108826404-34ed2880-75ff-11eb-9767-66453169d70b.png)

albert-ying:
+1... It is basically a disaster each time I forgot to change the input method.

evolify:
+1... any solution ?

nbfhscl:
+1. Seems related to #751 . The indicator shows still in insert mode.

nbfhscl:
This works for me with miscrosoft pinyin input. Need to download im-select.exe manually [here](https://github.com/daipeihust/im-select/raw/master/im-select-win/out/x86/im-select.exe).
`
if exists('g:vscode')
    autocmd InsertLeave * :silent :!C:\\tools\\neovim\\im-select.exe 1033
    autocmd InsertEnter * :silent :!C:\\tools\\neovim\\im-select.exe 2052
else
endif
`

theol0403:
Fixed by #1009

PR #900, opened 2022-05-11
fix(typing_manager): Non-English input method can replace chars

BTBMan:
Fix bug #560

## Problem

It can replace previous chars when type Non-English chars without insert mode. It's very annoying.

## Solution

I register `replacePreviousChar` command to handle this behavior

It can stop replace previous chars without insert mode


BTBMan:
@asvetliakov 🙏 Hope to merge this. Thanks!

theol0403:
I don't understand what this is supposed to do. Can you please explain the problem and how this solves it?

BTBMan:
Current:

bad

Optimized:

good

It can replace previous chars before optimized, I reference [VSCodeVim](https://github.com/VSCodeVim/Vim/blob/fe9b77de6de3ab73ff2f9a1eeee7c94ea159a771/extensionBase.ts#L365).

But I realized It also has some bugs, It can also replace chars sometimes and It cause Non-English text blink.

BTBMan:
VSCodeVim also has text blink problem, I don't know whether it is VScode problem, It don't had that problem when I use VSCodeVim long time ago.

theol0403:
> VSCodeVim also has text blink problem, I don't know whether it is VScode problem, It don't had that problem when I use VSCodeVim long time ago.

Can you give a link for where they talk about this?

Can you try adding to settings:
```js
  "extensions.experimental.affinity": {
    "vscodevim.vim": 1,
    "asvetliakov.vscode-neovim": 1
  },
```

BTBMan:
Sorry, I don't have, I have given up using VSCodeVim to use VSCodeNeovim for a long time, I just found that this extension has some bad experience, So I refer to the idea in VSCodeVim and use it.

BTBMan:
I just found a issues about this in VSCodeVim.
[https://github.com/VSCodeVim/Vim/issues/341](https://github.com/VSCodeVim/Vim/issues/341)

BTBMan:
It seems that text jitter is indeed a problem of VSCodeVim

theol0403:
Is it possible to instead use https://github.com/microsoft/vscode/issues/8133#issuecomment-232958381 instead of `replacePreviousChar`? I am just worried about side effects.

BTBMan:
I tried, I think It's useless, Now only `replacePreviousChar` can fix it.

theol0403 (approved):

wenfangdu:
I think this PR should be reverted, compared to the jittering IME input, the return is negligible.

@BTBMan Dude, I saw your comment [here](https://github.com/VSCodeVim/Vim/issues/7496#issuecomment-1181803905), if that's true, then we should wait for `replacePreviousChar` to be fixed before merging this.

BTBMan:
@theol0403 I think it's a solution. What did you think?

theol0403:
Sure, I'll revert and please let me know when I can merge, if #984 is also fixed.

BTBMan:
Ok!

PR #902, opened 2022-05-13
feat(messages): show Output less frequently

justinmk:
Problem:
Nvim can produce multiline messages for common actions like search.

Solution:
Don't focus the Output view for fewer than 3 lines.

ref #881
ref https://github.com/vscode-neovim/vscode-neovim/pull/881#issuecomment-1124323231


justinmk:
waiting for approval

PR #908, opened 2022-05-19
wip: fix command line history

theol0403:
This replaces #730 and fixes https://github.com/vscode-neovim/vscode-neovim/issues/708 and https://github.com/vscode-neovim/vscode-neovim/issues/689.

This PR will do a few things:
- fix command line history 
	- don't replace command line in nvim if no real change
		- currently, when you scroll it goes `cmdline_show` -> `value = content` -> `onChange` -> `input(value)` -> `cmdline_show`, which breaks history scrolling
		- fixed by detecting if update originated from nvim
		- `cmdline_show` now actually updates the quickpick instead of needing polling
	- this broke completions because spamming `wildchar` would insert new text
		- using `<C-d>` instead of `wildchar` fixed inserting, but if no completion it would insert `^d`
		- solution was to delete `^d`
	- better solution is to give up on live autocomplete and do what nvim does and just show completions on `wildchar`
- auto value updating means nvim can be used for `C-u`, `C-w`, etc in command line
TODO:
- let tab/CR cycle and select completions

In short: https://github.com/theol0403/vscode-neovim/tree/eb1f97521d21e6d008c14dd72de0c8d82db8a458 is a complete fix for #730, but requires hack for completions. 
This PR will totally refactor completions, removing code and edge cases.


theol0403:
All cmdline bindings (paste, C-r C-w, etc) works now!! As well as history navigation, and fixes the above mentioned issues.

justinmk (approved):
looks like a nice code reduction. are these codepaths exercised by the tests at all?

PR #916, opened 2022-05-22
fix(vscode-options): set zero foldcolumn

kevinhwang91:


kevinhwang91:
I am not familiar with the codebase, but I will try if I have spare time.

theol0403 (approved):

PR #919, opened 2022-05-23
feat: allow vscode sync viewport with neovim

windwp:
Problem:
Currently, the neovim viewport is always 200 and it uses scrolloff=100 to make the cursor always on center.

Solution:
this PR remove scrolloff=100 . It resize the height and scroll the neovim viewport. 

PS: i use it for test.
```lua
vim.keymap.set('n', '<leader>ua', function()
    local line_start = vim.fn.line('w0')
    local line_end = vim.fn.line('w$')
    print('S:' .. line_start .. ' E:' .. line_end)
end)

```


theol0403:
Thanks for this PR!! It seems to work really well and also makes https://github.com/vscode-neovim/vscode-neovim/pull/885 work great! 

One thing I noticed is that when I do the lightspeed jump, it seems to consistently think the viewport is ~2 lines smaller than it actually is:
image

theol0403:
I've found another bug with this PR: "jump to definition" or using the outline panel causes the viewport to jump to the new position and then instantly jump back.

src/cursor_manager.ts
```diff
@@ -576,73 +580,21 @@ export class CursorManager
...
-    //                 this.updateCursorPositionInNeovim(winId, e.selection.active.line, e.selection.active.character);
-    //             }
-    //         }
-    //     },
-    //     500,
-    //     { leading: false },
-    // );
-    // private commitScrollingFast = throttle(this.updateScreenRowFromScrolling, 200, { leading: false });
+    private onDidChangeVisibleRange = async (e: TextEditorVisibleRangesChangeEvent): Promise<void> => {
+        if (e.textEditor !== window.activeTextEditor || !this.modeManager.isNormalMode) {
```
theol0403, line 595: Should this be expanded to also include visual mode? Why is this check needed?

theol0403:
Unfortnately I'm not sure if the new commit solved the problem with navigating in the outline. It has to trigger a scroll for the issue to appear. 

Additionaly, would it be possible to rebase your commits on https://github.com/vscode-neovim/vscode-neovim/pull/775? I'm planning on merging it soon and it greatly simplifies the cursor setting logic. There is an obvious place to put `this.scrollNeovim(editor.visibleRanges);`. However, it doesn't solve the problem of being unable to jump using vscode commands (outline, jumplist, etc).

windwp:
you can merge it to master and I will rebase from the master.

I click on the outline panel and it works.

theol0403:
I think this is a timing problem. For me the issue is not from `updateCursorPosInEditor`, the issue is from `onDidChangeVisibleRange`. I think what is happening is that both the `onDidChangeVisibleRange` and `onSelectionChanged` triggers at the same time, and then when `scroll_viewport` is called, it causes nvim to update vscode with the old cursor position.

For me it happens about 80% of the time, especially on big files and big jumps. It jumps to the new position and then instantly jumps back.

theol0403:
Pressing `<C-o>` which uses vscode jumplist to try to jump me to the top of the file:

windwp:
i just fix it. sorry, i am not using vscode daily. thank for helping me test

theol0403:
Thanks! It seems to work for me now, I'll use it at work today and then merge if everything seems fine.

theol0403:
Sorry for the long delay, got busy with work. Unfortunately, the new changes are pretty buggy, but I'm not sure how to fully reproduce the issue.

Essentially, when I navigate around using lightspeed, and I jump to the botton of the file, and then jump in reverse, it doesn't make the whole screen grey, and I can't jump to the top ~5 lines. Moving the cursor does not fix this, only scrolling the viewport works.

theol0403:
It seems as though the test, shown in the PR description, works perfectly. However, it seems the problem is somehow caused by lightspeed.

theol0403 (approved):

theol0403:
Thanks @windwp for the initial work! I think this works very well now.

theol0403:
I noticed that @zbw8388 made a branch on their fork that made me realize a bug I missed - when scrolling with C-u/C-d (using #885), the veiwport is wrong.

This is because neovim internally scrolls the content while using C-u, so that the cursor position stays the same relative to the window. However, vscode moves the cursor and does not scroll the content. This causes the viewport to be unsynced. 

I can think of three solutions:
- sync neovim scroll with vscode. This could be done with `win_viewport`, or `winsaveview`. This would also remove the need for a custom implementation of C-e/C-y/zt/zb/zz. I believe @asvetliakov said this could cause jitter, but we can try. 
- modify vscode scrolling strategy. There might be a vscode api option where moving the cursor long distances will scroll the content instead of moving the cursor
- modify neovim scrolling strategy. Make neovim scroll how vscode does, which only scrolls the content if needed

I think the first option is ideal, if it doesn't rely on too much code.

windwp:
@theol0403 nice work :+1:

wenfangdu:
@windwp Please take a look at https://github.com/vscode-neovim/vscode-neovim/pull/993#issuecomment-1228119958, this PR is causing that issue.

wenfangdu:
@windwp @theol0403 Please take a look at https://github.com/vscode-neovim/vscode-neovim/pull/993#issuecomment-1228119958, this PR is causing that issue.

PR #925, opened 2022-05-29
only cd to cwd if not over ssh-remote

erees1:
I was getting errors when opening folders over ssh-remote using this extension eg:

```
Error detected while processing command line:
E344: Cannot find directory "/home/edwardr/git/dotfiles" in cdpath
E472: Command failed
```

This is because the `-c cd ${cwd}` is trying to cd to a folder on my local machine but the path is from my remote machine, so it doesn't exist. This PR adds a check to prevent this from happening.


ppwwyyxx:
Thanks for the fix. Can any of the maintainers please take a look?

theol0403:
Thanks for the PR! I combined it with some other conditions and committed to master.

PR #942, opened 2022-06-22
feat(keymap): support ctrl+r numpad commands in insert mode

DiogoDoreto:
This PR adds support to pasting from registers while in insert mode using NumPad keys.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1006,7 +1006,19 @@
             },
             {
                 "command": "vscode-neovim.paste-register",
+                "key": "ctrl+r NumPad0",
+                "when": "editorTextFocus && neovim.mode == insert && !neovim.recording && neovim.ctrlKeysInsert",
+                "args": "0"
+            },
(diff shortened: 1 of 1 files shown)


DiogoDoreto:
@justinmk yeah, for now I think that's the only solution. When #886 gets completed then we should be able to simplify it.

theol0403:
Thanks for the PR! #886 will indeed make this unnecessary, but I need to find the time to refactor the controller to make the PR work.

theol0403:
If you have the time, can you fix the merge conflicts by rebasing? Nothing major, I think all that changed was using `shift+` instead of `shift-` for consistency. I'll fix myself once I have the time if it isn't done.

theol0403:
Closed in favor of https://github.com/vscode-neovim/vscode-neovim/pull/886. Thanks anyway!
Please test it to see if it works for you.

PR #943, opened 2022-06-22
Update README to mention how to make explorer keybindings work.

vidhanio:


DiogoDoreto:
I fixed mine by setting `"workbench.list.automaticKeyboardNavigation": false` which completely disables the jumping by matching text. `"workbench.list.keyboardNavigation": "simple"` only changes it to prefix match, which could be confusing sometimes.

Anyway I think that it's a good idea to document this behavior 👍

justinmk (approved):

justinmk:
whoops, feel free to revert. Ultimately I hope the readme can be cut in half or more, it's too long, because there's too many workarounds. Fewer workarounds => less docs needed.

E.g. why do we need giant tables listing things like `vsplit: Split editor vertically.` ? That is an expected behavior. It's redundant with Nvim's own docs.

PR #948, opened 2022-06-30
experimental: always listen to type commands in insert mode

theol0403:
This has many advantages:
- remove dot repeat workarounds
- remove unbind/rebind to type, and associated sync
- makes insert mode bindings work out of the box
- remove composite escape hacks
- much code removed/simplified (I'd say another 50% of what has been removed can still be removed)

Disadvantages (why the workaround was added in the first place):
- typing lag
- cursor jitter
- random bugs (this can be fixed by the PR)
- completions not appearing

However, now you can do 
```js
  "extensions.experimental.affinity": {
    "asvetliakov.vscode-neovim": 1
  }, 
```

which makes the typing much faster. There is still cursor jitter, but everything seems very responsive and working out of the box.

### DIFF
--- a/package.json
+++ b/package.json
@@ -281,16 +281,16 @@
                 "title": "Neovim: send key"
             },
             {
+                "command": "vscode-neovim.sync-send",
+                "title": "Neovim: Sync pending insert-mode changes and send key"
+            },
+            {
                 "command": "vscode-neovim.compositeEscape1",
                 "title": "Composite escape key 1"
             },
             {
                 "command": "vscode-neovim.compositeEscape2",
                 "title": "Composite escape key 2"
-            },
-            {
-                "command": "vscode-neovim.escape",
-                "title": "Escape key"
             },
             {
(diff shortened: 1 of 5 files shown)


justinmk:
> makes insert mode bindings work out of the box

To avoid fighting with Nvim's insert-mode completion, and plugin mappings, I wonder if we should still (optionally) disable Nvim insert-mode mappings and features like ctrl-n. `:set paste` gets some of the way there, but ctrl-n still does completion.

theol0403:
I think it might be possible to fix cursor jitter and completions not working. 

Currently, when bound in insert mode, all `type` events gets forwarded to nvim. Then, nvim notifies vscode that edits have been made to the buffer, and the vscode buffer is updated. Then, nvim sets the new cursor pos (one character to the right). These two actions prevent completions from triggering and adds jitter (character gets added, then cursor moves). 

The solution would involve refactoring the edit application so that `type` effectively gets passed through and the cursor set is skipped, so that vscode handles moving the cursor forward.

Issue #746, opened 2021-09-18
[Feature Request] pass `--clean` to neovim


goolord:
motivation: i want to keep 2 separate configs for vscode-neovim and standalone nvim, including full control over the rtp. this way, i can forgo loading plugins that are broken or otherwise made unnecessary by vscode

goolord:
worth noting that this plugin basically broken for me unless i run with `-u NONE` since i can't pass `--clean`

trkoch:
You can distinguish Plugin/Standalone with `if exists('g:vscode')` in case you didn't know.

goolord:
my 2 configs are very different, I would basically be wrapping entire files like this and i would still need to manually manage my rubtime path which is a sisyphean task i don't want to do

theol0403:
Also you can specify a different init.vim location

goolord:
it doesn't matter if you can pass a different init.vim, the runtime path is global state. if your two init.vims use different runtime paths, this doesn't work

goolord:
> worth noting that this plugin basically broken for me unless i run with -u NONE since i can't pass --clean

jim-chien:
@goolord you can add this to your `init.vim` file:
```vim
if(exists('g:vscode')
  set noloadplugins
endif
```
The answer is at the section 9 here: https://neovim.io/doc/user/starting.html

theol0403:
I think it is more useful to have a shared config than two separate configs. If you want to separate them, add the logic yourself. Unless you want to make a PR for a config option.

goolord:
> Unless you want to make a PR for a config option.

so, is it worth having or is it not? 😞

I only opened the issue instead of making a PR myself because it's not at all obvious from 30 minutes of hacking or so how I would even add this feature, but I was hoping it's trivial for someone with a little more familiarity with the codebase / writing vs code plugins

goolord:
> I think it is more useful to have a shared config than two separate configs 

(´。＿。｀)
![image](https://user-images.githubusercontent.com/24906808/177023012-69386f06-a291-4516-baef-67b1a47eb6cb.png)

PR #952, opened 2022-07-03
feat: `--clean` option

goolord:
https://github.com/vscode-neovim/vscode-neovim/issues/746
allows finer control of the rtp, useful for users of https://github.com/wbthomason/packer.nvim


theol0403 (approved):
Thanks for the PR!

PR #953, opened 2022-07-03
Fix normal mode freezing after switching windows

### LINKED ISSUE #912: vscode lose focus and refocus Mouse cannot move in Normal mode
After I clicked on other software with my mouse
I found that the mouse can be moved in insert mode
all other modes cannot be moved

https://user-images.githubusercontent.com/26178326/169540998-edecf8f0-e7c9-411e-9e25-9dda3eed7598.mov

pranavrajpal:
This fixes a race condition that causes the extension to freeze in normal mode if someone changes tabs fast enough.

The race condition happens because the code in `syncLayout` doesn't handle the case where the visible editors and the active editor changes while `syncLayout` is still running correctly. In that case, while `syncLayout` is running due to the initial change, it will resolve `changeLayoutPromise` after the visible editors change. That causes `syncActiveEditor` to try to set the current window in Neovim, but will use the new active editor, so it will fail since that new editor wasn't in the list of visible editors when `syncLayout` was run, meaning that the Neovim window for that editor wasn't set up.

This PR:
- fixes the race condition by using a cancellation token to avoid resolving `changeLayoutPromise` if the list of visible editors has changed, which avoids `syncActiveEditor` continuing until `syncLayout` is rerun with the correct visible editors
- turns the logged warning about `syncActiveEditor` not finding a corresponding Neovim window for the active editor into an error since we should never get to that point unless something has gone wrong and it makes future freezes like this easier to debug

I'm open to suggestions on alternatives to using a cancellation token to fix the race condition, since that was the first method I thought of but I'm not sure if there are better methods to do the same thing.

Fixes #912

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -4,6 +4,8 @@ import { debounce } from "lodash-es";
 import { Buffer, NeovimClient, Window } from "neovim";
 import { ATTACH } from "neovim/lib/api/Buffer";
 import {
+    CancellationToken,
+    CancellationTokenSource,
     commands,
     Disposable,
     EndOfLine,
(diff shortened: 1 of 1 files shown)


theol0403:
Thanks for the PR! Please rebase on master to fix CI.

theol0403:
It is cleaner to rebase: 
`git rebase --interactive upstream/master`
this way there are no merge commits

theol0403:
Not sure why macos tests are currently failing

theol0403 (approved):

Issue #677, opened 2021-06-22
First hit of ESC is swallowed by suggestions, have to hit it twice to leave insert mode


lesiki:
When the IntelliSense window is visible, hitting escape removes the suggestions but keeps me in insert mode. Initially it 'felt' as though some escape key presses were being dropped, I only figured out that this was what was happening by slowly trying to recreate it.

I'm fairly sure this was just recently introduced in the latest VSCode 1.57.1 release.

The gif below shows me pressing escape (⎋) twice to leave insert mode.

![Screenshot 2021-06-22 at 10 05 46](https://user-images.githubusercontent.com/1673588/122879478-9c16b180-d341-11eb-83ab-e1d67f806df8.gif)

theol0403:
Yes, this was introduced by the latest version. 

https://github.com/asvetliakov/vscode-neovim/pull/513/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R539

https://github.com/asvetliakov/vscode-neovim/pull/604

Specifically, it allows parameter hints to swallow, not suggestions. I personally prefer the new way, but I understand if you prefer the old way. A quick fix is to change it in your user keybindings.json.

However, if this bothers too many people, I am happy to revert it, or maybe provide a configuration toggle for it. 

Please let me know if this is a popular issue.

lesiki:
Thanks @theol0403 , your suggested quick fix worked.

For anyone else interested in reverting to the old behaviour:
1. Open keyboard shortcuts (`Preferences → Keyboard Shortcuts`)
2. Click the button to the right of the search bar to enter JSON editing mode
3. Add a key binding with the original `where` clause:

```
// Place your key bindings in this file to override the defaults
[
    {
        "command": "vscode-neovim.escape",
        "key": "Escape",
        "when": "editorTextFocus && neovim.init"
    },
]
```

theol0403:
👍 

Should we keep the issue open? I don't know how much this change bothers other people.

Btw, I think you should keep *some* of the when conditions, like the peek escape which is useful (when you have a peek open, but then you return to the original editor, escape will close the peek).

Also, iirc doom emacs, one of the best examples of ide functionality + vim, makes escape close popups and notifications before it escapes normal mode.

theol0403:
Actually, I am wrong with doom emacs, and this is a good issue.

Using escape to close things in normal mode is valid, but not in insert mode.

[...]

PR #680, opened 2021-06-24
Refactor package.json bindings 

theol0403:
I hope to solve some of the issues I have introduced and improve the vscode bindings file.

Background info: 
- keybindings are evaluated bottom to top
- the first keybinding that matches the key and when clause is selected
- `&&` has more priority than `||` in when clause, ex. `a || b && c` == `a || (b && c)`

- [x] Refactor escape bindings (fixes #677)

The way the new escape bindings work are:
- in insert mode or visual mode, escape will *always* return to normal mode
- in normal mode, if certain vscode elements are open, escape will close them. If not, it will send to neovim

Situations where escape is sent to vscode **only in normal mode**: `markersNavigationVisible, parameterHintsVisible, inReferenceSearchEditor, referenceSearchVisible, dirtyDiffVisible, notebookCellFocused`

- [x] Allow notebook cell navigation (fixes #210, fixes #669, closes #676). Makes enter select the cell editor, and allows escape to exit the cell.

- [x] Fix peek mode bindings

`inReferenceSearchEditor && neovim.mode == normal || referenceSearchVisible` evaluates to:
- if mode is normal and search editor is focused, tab moves to the reference list
- if the reference search is visible, but the editor is not focused, then either the list is focused, or the main editor is focused. Either way, tab return to the reference editor.

- [x] C-o insert mode with macro recording (fixes #679). Now, you can use C-o while recording a macro.
- [x] Send tab during recording, which can no longer be used for vscode actions like completions (I don't think it worked before). Using tab to select selections don't work, everything must be sent to neovim. The cmdline when clause is removed, because it is moved up above the `vscode-neovim.complete-selection-cmdline` line, which will take priority.

- [x] Shuffle bindings
  - Move escape bindings to the top, because they are the most interesting
  - Move send bindings to top. 
    - This combines with the existing C-S-P send bindings
    - Sending a key to neovim is not as important as doing other actions. For example, tab to navigate the peek editor needs to be below the send bindings. Therefore, move the send bindings to the top, where there is low priority.
  - Move peek bindings up, and move list c-p bindings down. This lets you use c-p and c-n to navigate completion list in reference view, without navigating the peek list accidentally.

PR #960, opened 2022-07-06
refactor: merge two escape keybindings into one

wenfangdu:
Previously:

image

Merged two escape keybindings into one, improved escape's when condition.

### DIFF
--- a/package.json
+++ b/package.json
@@ -367,13 +367,8 @@
             },
             {
                 "command": "vscode-neovim.escape",
-                "key": "Escape",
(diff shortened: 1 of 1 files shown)


theol0403:
This was intentional, from #680 (fixes #677). 

What the two bindings do is
- insert mode will *always* send escape to vim, regardless of popups. This way you can always be sure you exit insert mode.
- normal mode will *only* send escape to vim if there are no popups. This way the escape gets sent to vscode which closes the popups.

PR #971, opened 2022-07-14
fix: fix cursor and highlight position related to long lines via `WinScrolled` autocmd

zbw8388:
This extension has multiple issues related to long lines, mainly due to lack of horizontal scroll notification. This PR implements a hack using `WinScrolled` autocmd event with `winsaveview()` to provide column offset changes notification, similar to `win_viewport` but with richer information. 

This PR also refactors part of the codebase so that now all operations related to viewport tracking is in one file. Future features such as handling scrolling by neovim can also be implemented in that file. 

Note that this PR cannot resolve highlight position issue related to searching on long lines due to event `WinScrolled` is not fired when neovim window scrolls due to searching.


theol0403 (approved):
Thanks for the PR!!

PR #972, opened 2022-07-14
README: added conditional init.lua

Lyr-7D1h:


tobealive (approved):
Why not `vim.g.vscode ~= 0`?

Additionally, I would check for equality with 1 instead of inequality with 0 here. 
When I'm still getting it right, I think the reason why I learned to prefer this is that I ran into occasions where such variable checks returned a boolean and an unequal  0 didn't get hooked. But I might confuse something here and changing the equality check won't make a difference, no matter the scenario.

tobealive (commented):
Why not using vim.g.vscode ~= 0 😊?

theol0403:
Yes, when I got to this I was going to suggest `if vim.g.vscode then`.

I was also planning on giving some tips on conditional loading with packer.nvim.

Lyr-7D1h:
@theol0403 yes definitely cleaner. I updated the docs!

theol0403 (approved):

theol0403:
Thank you for the PR!

PR #982, opened 2022-07-30
feat(highlights): improve compatibility with lightspeed/leap

theol0403:
There was a bug where if a highlight was set to 0 (for black), vscode would not display it. This made leap labels hard to read. This also removes the lightspeed/leap cursor workaround. 

This works well with #919


Jendker:
Thank you so much!

Issue #899, opened 2022-05-10
Cursor not placed on newly inserted line with `O`


joelostblom:
The expected behavior of hitting `O` is that a line is inserted above the current line and the cursor is placed on this new line. Sometimes I have noted that the cursor remain on the current line (ie, it appears to be "moved down" a line together with the text). In this video you can see an example of this. Every time I enter insert mode in the video it is by pressing `O`. The first couple of times work as expected, but then the cursor is repeatadly positioned at the line below the newly inserted line:

https://user-images.githubusercontent.com/4560057/167729963-5f4846ca-17ef-4cf1-a3e8-22c9d6331180.mp4

This particular clip is from a Jupyter Notebook, but I have seen the same behavior in regular text files. This is on vscode 1.67 with nvim 0.5.0 on PopOS 21.10.

theol0403:
Related: #864

theol0403:
This might be fixed by https://github.com/vscode-neovim/vscode-neovim/pull/992, can you give it a try?

andmis:
Just tried this, can't reproduce.

joelostblom:
I think you are correct @theol0403 (and sorry for the late reply). I just tried this again and I can't reproduce anymore so closing.

Issue #543, opened 2021-01-29
Lower-case marks lost after changes to line of mark in insert mode


trkoch:
**Summary**

It appears changing the buffer (e.g. `cw`) can mess up local marks.

**Steps**

Given following buffer (with cursor at `|`)

```
|Lorem ipsum dolor sit amet.
```

`ma`, `w`, `cw`, `foo`, `<esc>`, `'a`

**Actual**

```
Lorem foo| dolor sit amet.
```

(E20: Mark not set)

**Expected** 

```
|Lorem foo dolor sit amet.
```

(No error)

**Notes**

Upper-case marks (global) are not affected.

trkoch:
I did some further debugging:
* Only marks of changed lines are affected (i.e. marks in other lines are not lost)
* Marks are only lost if entering/exiting insert mode (e.g. `cw` not `D`)

Reading debug logs and looking through the code, my best guess so far is that lines [changed](https://github.com/asvetliakov/vscode-neovim/blob/master/src/document_change_manager.ts#L208) in Neovim buffer (after exiting insert mode) with API [method](https://neovim.io/doc/user/api.html#nvim_buf_set_lines()) `nvim_buf_set_lines` loose their marks.

What are your thoughts?

asvetliakov:
>Reading debug logs and looking through the code, my best guess so far is that lines changed in Neovim buffer (after exiting insert mode) with API method nvim_buf_set_lines loose their marks.

Yes, you're right. `nvim_buf_set_lines` is not preserving marks. Recent neovim nightly builds have new `nvim_buf_set_character` method (if i recall correctly) which doesn't suffer from this, but i haven't got time to look into this yet

trkoch:
Reading through related issues (https://github.com/neovim/neovim/issues/11368, https://github.com/neovim/neovim/pull/11833), it appears this is now called `nvim_buf_set_text`. 

```
nvim_buf_set_text({buffer}, {start_row}, {start_col}, {end_row}, {end_col},{replacement})
                Sets (replaces) a range in the buffer

                This is recommended over nvim_buf_set_lines when only
                modifying parts of a line, as extmarks will be preserved on
                non-modified parts of the touched lines.
```

[prepareEditRangesFromDiff](https://github.com/asvetliakov/vscode-neovim/blob/master/src/utils.ts#L151) does not appear to  report start/end columns yet. Is that feasible with diffs generated by [fast-diff](https://github.com/jhchen/fast-diff)?

[...]

Issue #84, opened 2019-12-09
No undo points when moving cursor in insert mode


NeoRaider:
When moving the cursor in insert mode using the arrow keys (or the mouse, when mouse support is enabled), (n)vim creates an undo point for each move. This does not work in vscode-neovim.

asvetliakov:
>When moving the cursor in insert mode using the arrow keys (n)vim creates an undo point for each move.

wow, I didn't know !! Anyway, this won't be fixed probably until there will be character range insert support/sending updates as edits API in neovim, see [this comment](https://github.com/asvetliakov/vscode-neovim/issues/76#issuecomment-562903490)

NeoRaider:
Wouldn't it be possible to do whatever happens when the insert mode is left each time the cursor moves while in insert mode (if something was edited)? Or would it just not make sense to implement it based on current APIs, as it would be replaced soon anyways?

asvetliakov:
yes, it's possible. I'd wait a little though - sending insert mode changes as edits immediately will solve many problems with cleaner way

Issue #864, opened 2022-04-16
Inconsistent results when pressing o [normal mode]


Dandarprox:
Sometimes when I press o the new line is created right after the current line, but for some reason the cursor jumps several lines at a time and it is not placed in the new line. I don't have anything in my current config remotely related with `o` nor my VSCode settings modify the cursor behavior. What can I do?

zhuzhzh:
I hit the same issue.

guyguy2001:
When you get this issue, does the rest of neovim work correctly?
I sometimes get into a corrupted state where it seems like neovim thinks it's editing different text from what actually is in vscode, and I may or may not have encountered this issue there (I'm not sure whether what I remember is this issue or not).
So I wonder if this problem is just a symptom of the other one, or if it happens on it's own.

Could you share footage of this if you encounter this issue again?

xgrei:
I hit the same issue.

I try to disable all extensions except neovim, the problem still exists.

xgrei:
https://user-images.githubusercontent.com/57981640/167289646-2c7010df-bce9-434f-9c68-26e2e44a3ed6.mp4

this is screen recording, only enabled neovim extension.

theol0403:
This might be fixed by https://github.com/vscode-neovim/vscode-neovim/pull/992, can you give it a try?

Jendker:
This issue occurs with neovim nightly but not v0.8.3. Just for the reference.

PR #992, opened 2022-08-04
feat(document_change_manager): replace `nvim_buf_set_lines` with `nvim_buf_set_text`

theol0403:
I started to work on fixing #543. 
Helps #84

So far it seems to work well, though a couple tests fail.

Unfortunately, it does not seem to fix either dot-repeat or marks. Marks are lost like before, and dot-repeat does not replay any of the changes sent using set_text.

I tried sending the changes to nvim in real-time, which showed promise. With this, when you type, it sends individual characters to nvim asynchronously (so no typing lag). However, this also didn't fix dot-repeat, and had cursor pos syncronization issues.

This does reduce code bloat however, and might improve performance (especially in the case where vscode makes changes in normal mode).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
             "name": "vscode-neovim",
             "version": "0.0.90",
             "dependencies": {
+                "async-mutex": "^0.3.2",
                 "fast-diff": "^1.2.0",
                 "lodash-es": "^4.17.21",
                 "neovim": "^4.9.0",
@@ -909,6 +910,19 @@
             "version": "3.2.4",
             "resolved": "https://registry.npmjs.org/async/-/async-3.2.4.tgz",
             "integrity": "sha512-iAB+JbDEGXhyIUavoDl9WP/Jj106Kz9DEn1DPgYw5ruDn0e3Wgi3sKFm55sASdGBNOQB8F59d9qQ7deqrHA8wQ=="
+        },
+        "node_modules/async-mutex": {
+            "version": "0.3.2",
+            "resolved": "https://registry.npmjs.org/async-mutex/-/async-mutex-0.3.2.tgz",
+            "integrity": "sha512-HuTK7E7MT7jZEh1P9GtRW9+aTWiDWWi9InbZ5hjxrnRa39KS4BW04+xLBhYNS2aXhHUIKZSw3gj4Pn1pj+qGAA==",
+            "dependencies": {
+                "tslib": "^2.3.1"
+            }
+        },
+        "node_modules/async-mutex/node_modules/tslib": {
+            "version": "2.4.0",
+            "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.4.0.tgz",
+            "integrity": "sha512-d6xOpEDfsi2CZVlPQzGeux8XMwLT9hssAsaPYExaQMuYskwb+x1x7J371tWlbBdWHroy99KnVB6qIkUbs5X3UQ=="
         },
         "node_modules/azure-devops-node-api": {
             "version": "11.2.0",
@@ -6225,6 +6239,21 @@
             "version": "3.2.4",
             "resolved": "https://registry.npmjs.org/async/-/async-3.2.4.tgz",
             "integrity": "sha512-iAB+JbDEGXhyIUavoDl9WP/Jj106Kz9DEn1DPgYw5ruDn0e3Wgi3sKFm55sASdGBNOQB8F59d9qQ7deqrHA8wQ=="
+        },
+        "async-mutex": {
+            "version": "0.3.2",
+            "resolved": "https://registry.npmjs.org/async-mutex/-/async-mutex-0.3.2.tgz",
+            "integrity": "sha512-HuTK7E7MT7jZEh1P9GtRW9+aTWiDWWi9InbZ5hjxrnRa39KS4BW04+xLBhYNS2aXhHUIKZSw3gj4Pn1pj+qGAA==",
+            "requires": {
+                "tslib": "^2.3.1"
+            },
+            "dependencies": {
+                "tslib": {
+                    "version": "2.4.0",
+                    "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.4.0.tgz",
+                    "integrity": "sha512-d6xOpEDfsi2CZVlPQzGeux8XMwLT9hssAsaPYExaQMuYskwb+x1x7J371tWlbBdWHroy99KnVB6qIkUbs5X3UQ=="
+                }
+            }
         },
         "azure-devops-node-api": {
             "version": "11.2.0",
--- a/package.json
+++ b/package.json
@@ -1285,6 +1285,7 @@
     },
     "dependencies": {
         "fast-diff": "^1.2.0",
+        "async-mutex": "^0.3.2",
         "lodash-es": "^4.17.21",
         "neovim": "^4.9.0",
         "ts-wcwidth": "^2.0.3"
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -3,6 +3,7 @@ import { NeovimClient } from "neovim";
 import {
     commands,
     Disposable,
+    Position,
     Selection,
     TextEditor,
     TextEditorCursorStyle,
@@ -49,11 +50,19 @@ export class CursorManager
      * ! Note: we should track this because setting cursor as consequence of neovim event will trigger onDidChangeTextEditorSelection with Command kind
      * ! And we should skip it and don't try to send cursor update into neovim again, otherwise few things may break, especially jumplist
      */
-    private neovimCursorPosition: WeakMap<TextEditor, { line: number; col: number }> = new WeakMap();
+    private neovimCursorPosition: WeakMap<TextEditor, Position> = new WeakMap();
     /**
      * Special workaround flag to ignore editor selection events
      */
     private ignoreSelectionEvents = false;
+    /**
+     * In insert mode, cursor updates can be sent due to document changes. We should ignore them to
+     * avoid interfering with vscode typing. However, they are important for various actions, such as
+     * cursor updates while entering insert mode and insert mode commands. Thus, when those events occur,
+     * this flag is used to disable ignoring the update. This is set to true when entering insert
+     * mode or running insert mode command, and set to false before document updates in insert mode.
+     */
+    public wantInsertCursorUpdate = true;
     /**
      * Set of grid that needs to undergo cursor update
      */
@@ -166,7 +175,7 @@ export class CursorManager
             const docPromises = this.main.changeManager.getDocumentChangeCompletionLock(editor.document);
             if (docPromises) {
                 this.logger.debug(
-                    `${LOG_PREFIX}: Waiting for document change completion before setting the cursor, gridId: ${gridId}`,
+                    `${LOG_PREFIX}: Waiting for document change completion before setting the cursor, gridId: ${gridId}, pos: [${cursorPos.line}, ${cursorPos.col}]`,
                 );
                 docPromises.then(() => {
                     try {
@@ -178,7 +187,7 @@ export class CursorManager
                             cursorPos.isByteCol ? 1 : (editor.options.tabSize as number),
                             cursorPos.isByteCol,
                         );
-                        this.neovimCursorPosition.set(editor, { line: cursorPos.line, col: finalCol });
+                        this.neovimCursorPosition.set(editor, new Position(cursorPos.line, finalCol));
                         // !Often, especially with complex multi-command operations, neovim sends multiple cursor updates in multiple batches
                         // !To not mess the cursor, try to debounce the update
                         this.getDebouncedUpdateCursorPos(editor)(editor, cursorPos.line, finalCol);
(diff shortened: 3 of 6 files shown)


theol0403:
I managed to reproduce the issue that is causing the error on the neovim side. 

Steps:
- `nvim --listen /tmp/nvim`
- create buffer with contents: 
```
one


two
```
- NVIM_LISTEN_ADDRESS=/tmp/nvim node --experimental-repl-await
- const nvim = await require('neovim/scripts/nvim')
- `await nvim.callAtomic([["nvim_buf_set_text",[2,0,3,1,0,[""]]],["nvim_buf_set_text",[2,0,3,1,0,[""]]]])`

The buffer contains gibberish.

theol0403:
Made upstream issue: https://github.com/neovim/neovim/issues/19643

theol0403:
Related #336

theol0403:
Been using this, and have been experiencing quite a few buffer desyncs. I think the code that counts changes is broken.

theol0403:
These new changes look really promising! A TON of code deleted, and performance is quite improved. No more delay when escaping!

Essentially, this relays changes to nvim *as you type*, without blocking the `type` event. Instead, it uses document changes to asyncronously drive the syncing.

However, currently dot-repeat is broken, because `nvim_buf_set_text` does not work with dot repeat. If that can be fixed by nvim devs, this would be a huge improvement.

Specifically, inserting text in insert mode with `nvim_buf_set_text`, in a continuous typing-like fashion, and then escaping, is not repeatable. 

cc @justinmk

theol0403:
I have re-added the dot-repeat hacks. This might cause a bit more escape lag. However, dot-repeat is important enough to support until nvim adds better support for `nvim_buf_set_text`. I'll make an issue for neovim soon.

There are a few desyncs still.

justinmk (approved):

theol0403:
Sorry for all the noise, this has been a rollercoaster of ideas, workarounds, and issues.

# Vscode -> neovim sync

Currently, most implementations rely on keeping the dot-repeat hack (make new buffer, replay changes), until `nvim_buf_set_text` can update dot-repeat. This involves storing some stuff in `onChangeTextDocument`, and replaying changes on insert exit. Unless an alternate dot-repeat workaround is provided, the original workaround is stil needed.

## Previous implementation
* Whenever the vscode buffer changes, `onChangeTextDocument` is called. 
	* Mark the changed document
* Whenever insert mode is exited, call `syncDocumentsWithNeovim`. 
	* For every changed document:
		* Generate a hash for every line, and then run a diff on the hashes
		* For every changed line, do some magic processing and then call `nvim_buf_set_lines`

### Issues
* Lots of gross code. 
* `nvim_buf_set_lines` is slower and less efficient than `nvim_buf_set_text`. 
* Loss of potential mark and dot-repeat improvements made to `nvim_buf_set_text` in the future.
* Need to diff on leaving insert mode.

## Idea 1 - live editing
`onChangeTextDocument` also gives you `contentChanges`, which give you a start/end/replacement to surgically edit text. 

* Whenever the vscode buffer changes, `onChangeTextDocument` is called. 
	* immediately call `nvim_buf_set_text` with `contentChanges`

This deletes a ton of code, and improves efficiency. Now, changes are replayed to nvim in real-time, so no syncing on exit insert mode. This does not cause typing lag, as the changes are replayed async. 

### Issues
1
Calling `nvim_buf_set_text` can cause the cursor to update in insert mode. This interferes with vscode editing, as it can move the cursor while you are trying to type, or break multicursor. 
However, ignoring cursor updates in insert mode is not possible, because we need it for:
* `a/A/I/ciw/c0/etc`
* `i_CTRL-W/R/etc`
* `i_CTRL-O l`

If the cursor update issue can be fixed, this is the best method.
Maybe we can tell vscode to not ignore a cursor update that comes right after entering insert mode, or that comes from insert mode binding. 

2
Another issue is that while `contentChanges` uses change positions that are relative to the previous doc version, we need to apply `nvim_buf_set_text` one-by-one, which can mess up positions. We need to compensate for this. 

### Alternate dot-repeat workaround
With specific changes (`rangeLength == 0` or `text == ""`), convert it to `nvim_input` with `text`/`<BS>`/`<Del>` based on cursor position. However, this worsens the cursor update issue.

Also, this cannot handle very specific edits. For example, typing `f(1)` in vscode will break this workaround, because smart parenthesis will be added: `f`, `f()`, `f(1)`. To excecute this using `nvim_input`, the sequence would look like `f()<Left>1`, which would break dot-repeat. 

However, this workaround would make exiting from insert mode **very fast**, as there is no buffer sync or dot-repeat replay needed on exit insert mode.

## Idea 2 - replayed editing
* Whenever the vscode buffer changes, `onChangeTextDocument` is called. 
 	* store `contentChanges`
* Whenever insert mode is exited, call `syncDocumentsWithNeovim`. 
	* for every `contentChange`, call `nvim_buf_set_text`

This also deletes lots of code and avoids the cursor issue.

### Issues
Same as Idea 1 issue 2.
Also, replaying is quite slow as the changes caused by typing are one character at a time, resulting in a ton of `nvim_buf_set_text`. 

[`2025a62` (#992)](https://github.com/vscode-neovim/vscode-neovim/pull/992/commits/2025a6241a166796936be8b2cb6d0c3afab6fd41) is this idea with the issues. 

### Alternate dot-repeat workaround
Same as Idea 1 workaround.

### Slow replaying workaround
Process the changes to combine compatible changes before applying. For example convert 
```
[nvim_buf_set_text(0, 0, 0, 0, 0, "a"), 
nvim_buf_set_text(0, 0, 1, 0, 1, "b"), 
nvim_buf_set_text(0, 0, 1, 0, 2, ""), 
nvim_buf_set_text(0, 0, 0, 0, 1, "c")]` 
```
into `nvim_buf_set_text(0, 0, 0, 0, 0, "c")`. 
This is tricky. 
Additionally, Idea 1 dot-repeat workaround would also fix this for edits that are dot-repeatable. 

## Idea 3 - alternate diffing
This is essentially the same as the current implementation, but uses `fast-meyers-diff` instead of the custom diffing algorithm. This then supports editing using `nvim_buf_set_text`. A function was written to convert between the diff output and `nvim_buf_set_text`, solving the issues similar to Idea 1 issue 2. 

This deletes original diffing code, and uses `nvim_buf_set_text`. 

### Issues
The conversion function might have a bug, I've experienced a desync in very demanding situtation. However, might not have been caused by this. 

Does a character-wise diff every time you exit insert mode. This seems to be quite fast with `fast-meyers-diff`, however. 

An alternate dot-repeat workaround is not possible, because of the way the diff can work. If you start with `a c`, put cursor on `a`, type `a b`, the diff could dell you `nvim_buf_set_text(0, 0, 1, 0, 1, " b")` or `nvim_buf_set_text(0, 0, 2, 0, 2, "b ")`. Thus information about how the change was typed is lost. 

[`1a29161` (#992)](https://github.com/vscode-neovim/vscode-neovim/pull/992/commits/1a29161da27bdd8eb22951431fe0b8a6fd885d41) is this. 

## Conclusion
I think the best option is Idea 1, if the cursor problem can be fixed, with the existing dot-repeat workaround until `nvim_buf_set_text` supports dot-repeat. The alternate workaround could work if being unable to type `f(1)` is a minor issue, and would speed up exiting from insert mode.

Idea 2 is good if we can write an algorithm that combines the changes into one. This seems challenging though. 

Idea 3 is also fine, but could have performance issues (or not). 

# Neovim -> vscode sync

This change extends the scope of the PR to go the other way. However, the change is related, and has some shared code.

## Previous implementation
* on neovim buffer update
* reconstruct neovim buffer from neovim edits
* Generate a hash for every line, and then run a diff on the hashes
* Do some magic to process the lines
* For every changed range, replace the vscode lines in that range
	* If the range is one line, run `fast-diff` to edit within the line

## Idea 1 - fast-meyers-diff
Instead of doing custom diffing, run fast-meyers-diff on the entire buffer, and directly replay the changes. 

This deletes lots of gross code. 

### Issues
This could be slow to run a diff on every neovim update. Maybe the line hashing is needed. Could do line hashing with fast-meyers-diff maybe. 

Might be a desync somewhere? Not totally sure.

## Idea 2 - directly apply nvim edits
Not sure why this wasn't done before, but maybe it is possible.

# Conclusion
Wow that was a lot of typing. 
Hopefully this makes some sense. The goal is to delete old gross code, and replace with smaller, efficient and faster code. 

Hopefully I will be able to finish this before school starts in a couple weeks.

@justinmk You don't have to read this but maybe you would find it interesting or have insight. Either approach has a lot of deleted code. Also, maybe there are some things you can speed up on neovim side?
@asvetliakov if you are not too busy, some insight into performance would be very valuable!

theol0403:
Next steps for me:

- [ ] investigate smartly ignoring cursor updates
- [ ] fix edits offset for live editing method
- [ ] investigate doing hash diffing for neovim-vscode sync
- [ ] investgate directly applying neovim edits

theol0403:
It is becoming clear that this is two separate changes, for each way of syncing.

To avoid delaying this PR, I have separated the vscode -> neovim sync, which seems to be working fairly well now.

theol0403:
This is ready for review, and barring bugs, should be ready for merge. 

Summary of changes:
- when vscode reports changed document, changes get sen immediately to nvim
- in insert mode, changes get sent immediately, instead of on exit
- nvim cursor updates are ignored in insert mode except for specific situations (mode change, special keys)
- this should make exiting insert mode faster, but there is still delay due to dot repeat

justinmk:
SGTM! Your summary https://github.com/vscode-neovim/vscode-neovim/pull/992#issuecomment-1242993483 would be helpful to have in the commit message.

> @justinmk ... Also, maybe there are some things you can speed up on neovim side (dot repeat w/ nvim_buf_set_text)?

We will eventually allow plugins to explicitly set the dot-repeat register. Does that help?

theol0403:
> We will eventually allow plugins to explicitly set the dot-repeat register. Does that help?

Yes, that should be enough to improve performance and remove the hack. There will still have to be some algorithm that converts a sequence of `nvim_buf_set_text` into a dot-repeat string (if the change is continuous), but it would be better than the current solution (which involves replaying the changes).

theol0403:
@justinmk how long until you think that will be supported? There is no big rush for this PR, and if you think it will be supported in the next month I'd rather wait.

justinmk:
> @justinmk how long until you think that will be supported? There is no big rush for this PR, and if you think it will be supported in the next month I'd rather wait.

Probably not that near-term. Meanwhile, I've reviewed your analysis above, here are my thoughts.

> until nvim_buf_set_text can update dot-repeat

`nvim_buf_set_text` already correctly sets dot-repeat, confirmed with python client: 

```
>>> import pynvim
>>> n = pynvim.attach('socket', path='.../nvim.64463.0')
>>> n.request('nvim_buf_set_text', 0, 0, 0, 0, 0, ['foo', 'bar'])
```

...however, IIUC from your later comments, what you actually want is for the "atoms" of change to be broken into "undo units". Correct?

> Process the changes to combine compatible changes before applying. For example convert
> ... This is tricky.

Don't bother. That reduces to reimplementing Vim normal-mode. Even Vim/Nvim itself doesn't understand its own "atoms", yet. (This is planned work for Nvim.)

>  fast-meyers-diff

Not sure where you're going with this, but it sounds like the plan would be to insert undo-points at random places?

My recommendations:

1. keep vscode-neovim ruthlessly simple so that we can reason about it.
2. what problem are we actually solving? are the "undo chunks" currently too big? 
3. are you aware of `i_CTRL-G_u` and `i_CTRL-G_U` ?

theol0403:
Related: https://github.com/vscode-neovim/vscode-neovim/issues/76#issuecomment-562903490

theol0403:
Unless CI fails... I believe I've finally reached something that works!

Needs lots of testing though, this is a pretty section of the project.

theol0403:
@justinmk replying to your review:

> nvim_buf_set_text already correctly sets dot-repeat, confirmed with python client:

This is not the behavior I was referring to. See https://github.com/neovim/neovim/issues/19806.

> what you actually want is for the "atoms" of change to be broken into "undo units". Correct?

I don't think this is correct. All this code is related to is sending edits to neovim, like https://github.com/vscode-neovim/vscode-neovim/issues/76#issuecomment-562903490. The issue is that sending edits while in insert mode does not update dot repeat.

The vscode -> neovim sync (what the scope of this PR is reduced to) is regarding replaying vscode edits in neovim, either real-time or upon exiting insert mode. The options were:
- idea 1: relay vscode edits to nvim in real time (what this PR currently does)
- idea 2: record vscode edits and then replay them upon insert exit
- idea 3: diff the vscode and neovim buffer upon insert exit, use diff to generate (efficient) edits, replay those edits

The neovim -> vscode sync is regarding updating the vscode buffer from edits sent from neovim. The options included different diffing methods to apply the edits.

> The fact that nvim_buf_set_text is being used makes me think I'm misunderstanding the problem here.

Might be the case.

theol0403:
This PR also fixes a bug where when the buffer is **first** opened, pressing `o` will trigger a vscode -> neovim edit, which can mess up the cursor. Could fix https://github.com/vscode-neovim/vscode-neovim/issues/899 and https://github.com/vscode-neovim/vscode-neovim/issues/864, not positive though.

Issue #983, opened 2022-07-31
Viewport scrolling issue with C-d/C-u


theol0403:
I noticed that @zbw8388 made a branch on their fork that made me realize a bug I missed when merging #919 and #885 - when scrolling with C-u/C-d, the veiwport is wrong.

This is because neovim internally scrolls the content while using C-u, so that the cursor position stays the same relative to the window. However, vscode moves the cursor and does not scroll the content. This causes the viewport to be unsynced.

I can think of three solutions:

    sync neovim scroll with vscode. This could be done with win_viewport, or winsaveview. This would also remove the need for a custom implementation of C-e/C-y/zt/zb/zz. I believe @asvetliakov said this could cause jitter, but we can try.
    modify vscode scrolling strategy. There might be a vscode api option where moving the cursor long distances will scroll the content instead of moving the cursor
    modify neovim scrolling strategy. Make neovim scroll how vscode does, which only scrolls the content if needed

I think the first option is ideal, if it doesn't rely on too much code.

theol0403:
@zbw8388 branch: https://github.com/zbw8388/vscode-neovim

You might find this commit useful, it changes the scroll command and reduces jitter: https://github.com/theol0403/vscode-neovim/commit/b21f7fc8eafeb1be992d0dd69b70d81922f23486

I made a branch based on @zbw8388 's work where it totally relies on neovim for scrolling behaviour. This seems to work fairly well but has some jitter. This deletes a lot of code.

https://github.com/theol0403/vscode-neovim/tree/experimental-scroll

tobealive:
As vscode is generally a bit troublesome with keyboard scrolling, I wrote a plugin recently that I think works fairly well. Maybe it also improves the experience for others who use vscode-neovim until there is a final implementation and solution.
https://github.com/tobealive/german-scroll.code  

And so great to see the skyrocketing development in recent days. I think slowly but steady, you make it better than the VSCodeVim extension 😊

zbw8388:
Hi @theol0403, thank you for getting on this! I created a branch in this repo (https://github.com/vscode-neovim/vscode-neovim/tree/experimental-scroll) so it would be easier to collaborate. 

[...]

PR #502, opened 2021-01-02
Folding support

asvetliakov:
Fixes #58 

Seems easy to add basic support. By design vscode folds must be used, and neovim won't be aware of them. However we'll skip fold block when moving line-by-line in vscode. 

The fold will be opened automatically if you perform `<num>j/k` where `num > 1` and the target line is within a fold

- [x] Movement over folds
- [ ] Commands/Mappings
- [ ] Tests

Issue #58, opened 2019-11-25
Folding


asvetliakov:
Need to sync vscode folds to neovim. from the initial look ```onDidChangeTextEditorVisibleRanges``` could be used with ```:{range}fold``` command
Rebind neovim folding keys/commands to call vscode folds commands

trkoch:
Perhaps add a note about folding not being supported just now to known issues in README?

pianocomposer321:
Any progress on this? Opening folds makes this extension almost unusable for me.

asvetliakov:
@pianocomposer321 This one is very tricky, because the folding API has a sad story in vscode. The range event is very generic and there are many limitations/issues with it.

pianocomposer321:
I found [this](https://github.com/VSCodeVim/Vim/issues/4517) solution on another issue. It's actually for vscodevim, but I like this extension much better, and it still works for it, with a few changes. Instead of changing the keybindings in vscode's keybindins.json do this:

if exists('g:vscode')
nnoremap j :call VSCodeCall('cursorDown')<cr>
nnoremap k :call VSCodeCall('cursorUp')<cr>
endif

I was ssssoooo happy that this works, and I don't know why nobody ever thought of it before! it doesn't even slow vscode down!

asvetliakov:
You need `<cr>` on the ends, but anyway good workaround, if you don't use something like `10j`. Although this could be modified by looking into `v:count` and calling it in the loop (but cursor movement may be slow in this case). Alternative way is the `cursorMove` command which supports jump count, but it opens folds as i remember correctly

pianocomposer321:
I meant to use <cr>, forgot that when typing it. But it is in my init.vim.

Actually, things like `10j` do work, but a little slower than before.

pianocomposer321:
So I actually have noticed a bit of a lag, but a much bigger problem is that remapping the keys to cursorDown and cursorUp makes the neovim instance think that the cursor is on a different line then it is in some cases. This causes problems like `dd` deleting the wrong line, `yy` yanking the wrong line, etc.

There is another solution that I think may work, but I don't know how to implement it. Pressing gj and gk skip over folds as they should, but in my init.vim I actually had already mapped `j` and `k` to `gj` and `gk` before I started using vscode neovim, but vscode neovim does not seem to use those mappings. Any ideas?

pianocomposer321:
[...]

PR #993, opened 2022-08-05
feat(scrolling): sync neovim scroll with vscode

zbw8388:
See #983 for more discussion.
This PR takes the first approach in that issue to fix content scrolling. Creating this PR to link issue and branch together. 

Tracking: 
Legend: ~~not planned~~

- [x] synchronize viewport as soon as a buffer is created 
- [x] (good to have) when executing lua, specify winId as currently the message is not passed on instantly
- [x] ~~(good to have) merge set cursor with the lua as winrestview natively support them~~
- [x] fix mouse wheel scrolling
- [x] ~~support folding~~
- [ ] fix quick-preview induced buffer height change 
- [ ] tests


theol0403:
Thank you so much for continuing to work on this!! I've found a few issues with the WIP, but I'm sure you are aware and are continuing to work at it.

Please let me know if there is anything you want me to test! I'll let you work on this and focus on some other tasks for now.

zbw8388:
@theol0403 I think this is mostly stable if people don't use code folding as that would mess up height computation. I made it so that mouse related events (clicking, scrolling by mouse wheel, scrolling by clicking on scrollbar, scrolling by dragging minimap) to be as close to vanilla vscode experience as possible. I'll be using it at work next week for discovering bugs. It'll be great if you can try it out as well. 
I also hope people with mac can help testing it out as this extension seem to have issues on those machines.

zbw8388:
Update on folding: 

I created a new branch dealing with code folding: https://github.com/vscode-neovim/vscode-neovim/tree/experimental-scroll-fold 
using techniques found in https://github.com/kevinhwang91/nvim-ufo/blob/56a8538d65ed69657b7a4d4163fc6cf87cbcf192/lua/ufo/fold/driver.lua
This lets vscode handle all folding related stuff and sync folded lines (basically the idea from https://github.com/vscode-neovim/vscode-neovim/issues/58)
Note that I chose to let vscode handle it as it is dependent on language syntax, and we should allow only one thing handle syntax

---
However, this branch suffers from multiple issues:
- Cannot process global folding (meaning C-d cannot correctly detect folded region if it is outside of current viewport)
- Costly updates as every syncing of viewport requires a window update
- Need to handle highlighting etc.

Possible fixes:
- Check if folding has changed for the current document *every* time visible region changes 
  - Might be costly 
  - Need to skip folded regions for highlighting 
- Stop passing folded region to neovim and let vscode handle possible fold region collision
  - Will break neovim jumplist but we're not using it anyways 
  - Will break highlighting (what if lightspeed labeled something in the folded region?)

theol0403:
@zbw8388 these updates look very promising! 
It's too bad about the folding, but I think for now its a feature that we can't support. Btw, have you seen #502? Also, VSCodeVim seems to have figured out a solution to folds, maybe you can look at that (https://github.com/VSCodeVim/Vim/pull/1552). Chances are we are talking about totally different issues though. 

I've been trying out your PR.
If I disable vscode -> neovim scrolling, scrolling (in neovim) works very well! C-u/C-d, zt/zb/zz, j/k, lightspeed, etc work perfectly, and scrolling inside vscode feels native.

However, I've been experiencing problems with two-way sync enabled.
Essentially, any scrolling motion results in the viewport jumping around sporadically, almost like it is being interrupted mid-scroll.

Vscode -> neovim disabled:

https://i.imgur.com/NRoujfJ.gif

Vscode -> neovim enabled:

https://i.imgur.com/Twe9Lhy.gif

https://i.imgur.com/yK5T4aK.gif

Let me know if you want any help reproducing this!

theol0403:
I've found that increasing the 20ms delay to 50ms fixes the problem for me. Perhaps this is a workaround that can be removed?

zbw8388:
> I've found that increasing the 20ms delay to 100ms almost fixes the problem for me. Perhaps this is a workaround that can be removed?
> 
> Edit:
> With a timeout of 200ms, if I double-press C-u, I still get the jumpy viewport (which settles in the wrong spot).

Yeah I can reproduce this scrolling bug. I always had `smoothScrolling` disabled so I did not notice it when I was working on it. It's kinda stupid: VSCode would issue `onDidChangeTextEditorVisibleRanges` *during* the animation, and due to https://github.com/microsoft/vscode/issues/84351, we don't really know if we need to deal with it or not. A simple fix would be finding out the animation time hidden in their repo and replace `viewport.ts:324` with an `await new Promise((resolve) => setTimeout(resolve, ...))`, which asks vscode-side code to ignore requests by `onDidChangeTextEditorVisibleRanges` for at least that amount of time when it is scrolling. 

I also imagine other issues will pop up due to this weird behavior such as going to definition-induced scroll.

> Btw, have you seen https://github.com/vscode-neovim/vscode-neovim/pull/502? Also, VSCodeVim seems to have figured out a solution to folds, maybe you can look at that (https://github.com/VSCodeVim/Vim/pull/1552). Chances are we are talking about totally different issues though.

I read through https://github.com/VSCodeVim/Vim/pull/1552. They use multiple chained vscode-side window scroll commands to fix this issue, which would probably also cause animation issue. 
I think my other branch https://github.com/vscode-neovim/vscode-neovim/tree/experimental-scroll-fold would cover the solution in https://github.com/vscode-neovim/vscode-neovim/pull/502: only the same-screen folds would get taken care of. That might be a nice partial solution though. Doing some careful optimization we can also let neovim register all folds generated in the visibleRegions. 

Another thing is that both solutions introduce vscode scrolling logic, which we would like to get rid of in the first place, so I think opting for them might be a bit extra.

---

I might not have time for at least this week to work on this. Feel free to work on it though! I also found a couple bugs that needs to get addressed:
- Checking (if vscode -> neovim message is necessary) should also get viewport height involved (will need some refactoring of code)
- Switching to a desynced tab requires ignore first neovim viewport request  
- Removing <= 1 checks (good for when there's only one folded line on top)
- (might be good to have) Stopping neovim from echoing vscode's scrolling messages back 
- ... or just sequence all messages but I haven't thought it through and might be an overkill

theol0403:
Thanks for all the great info! Yeah, that is a bit dumb.

Unfortunately, there is still a bit of trouble with smooth scrolling w/ two-way (occasional jitter, although correct end pos), and scrolling with the mouse is still sometimes jittery even with 200ms delay (the momentum gets stopped).

Requiring no smooth scrolling might be a concevable compromise. 

All good about being busy! You've already done so much. I'll see what I can improve when I have a chance, but the techincal bits feel a bit beyond me currently. No rush to get this shipped. I might even release master as-is, because the only bug is lightspeed being broken immediately after C-u/C-d.

zbw8388:
Try the latest commit and see if it solves the issue :) Maybe it'd be better to look at config prior to changing setting that timeout (but who would scroll by mouse in 126ms after clicking on their keyboard?)

I would be a bit hesitant to release master now, as it involves (I think) a major change to user experience, and we are expecting another change once this PR is ready. Maybe one way to do it would be excluding https://github.com/vscode-neovim/vscode-neovim/commit/29233d5c9fad4fd5ea0cd4c8a14c7f50385a0cd2 for releasing until we are fixed on how people scroll stuff

theol0403:
That seems to fix the issue with C-u/C-d! I'm still running into the momentum clamping with mouse scroll, but it is better. 

Sounds good about the release!

theol0403:
It's much better now! Unfortunately, I still experience the jitter when scrolling (slowly) with C-u/C-d about 1/10 times, and mouse scrolling is still broken. What if we were to debounce incoming scroll changes from vscode?

theol0403:
Applying the following diff to `f446bcb52ce55957108ae9f63318a9d495045a39` seems to make things work much better for me:

```diff
diff --git a/src/viewport_manager.ts b/src/viewport_manager.ts
index e3a287c..22e1f8f 100644
--- a/src/viewport_manager.ts
+++ b/src/viewport_manager.ts
@@ -1,3 +1,4 @@
+import debounce from "lodash-es/debounce";
 import { NeovimClient } from "neovim";
 import vscode, {
     Disposable,
@@ -77,7 +78,11 @@ export class ViewportManager implements Disposable, NeovimRedrawProcessable, Neo
         private changeManager: DocumentChangeManager,
     ) {
         this.disposables.push(window.onDidChangeTextEditorSelection(this.onDidChangeTextEditorSelection));
-        this.disposables.push(window.onDidChangeTextEditorVisibleRanges(this.onDidChangeTextEditorVisibleRanges));
+        this.disposables.push(
+            window.onDidChangeTextEditorVisibleRanges(
+                debounce(this.onDidChangeTextEditorVisibleRanges, SMOOTH_SCROLLING_TIME + 1),
+            ),
+        );
     }
 
     public dispose(): void {
```

theol0403:
Another bug that needs to be fixed: it seems like the viewport does not scroll to show search matches.

theol0403:
> Ah have you get a chance to try [6757f39](https://github.com/vscode-neovim/vscode-neovim/pull/993/commits/6757f39ee12e7fc6850d8dcb91a759fbc8e31052)? I increased the timeout, which is somewhat similar to debounce, except it starts the timer right after the first firing of the event.

I believe that was what I was referring to in my last comment. Yeah, in your last commit, the mouse scrolling is still broken, although I am having trouble reproducing C-u/C-d jitter (though I think it happens).

theol0403:
Another bug: in neovim, you can press `C-e` to scroll down beyond the last line. 
If `editor.scrollBeyondLastLine` is disabled, this leads to discontinuity. 
If it is enabled, it seems to work, except when you scroll all the way down, `C-y` does not seem to be able to scroll back up.

zbw8388:
> I believe that was what I was referring to in my last comment. Yeah, in your last commit, the mouse scrolling is still broken, although I am having trouble reproducing C-u/C-d jitter (though I think it happens).

May I see your log? If possible, could you also add logging of `events` at `main_controller.ts:280`? Also what system are you on? My mouse scrolling currently looks like this:

[Screencast from 08-10-2022 05:12:41 PM.webm](https://user-images.githubusercontent.com/60573138/184021469-4c46a65d-95fa-4a2c-9919-cb4238b1506a.webm)

I'll need time to debug `editor.scrollBeyondLastLine`, and I think the other one is caused by <= 1 check

theol0403:
I'm doing these tests on my work computer using windows.

It's somewhat irregular to produce, but here are the logs and screenshot:

https://pastebin.com/WRPnEPgU

https://gfycat.com/UnitedWeepyArizonaalligatorlizard

Edit: sorry, saw you wanted some extra event logs too. I'm not sure how to log these, it crashes when I try to json stringify and it contains a lot of data.

theol0403:
Here we are, events using `util.inspect`:

https://i.imgur.com/8Er5QA3.gif


https://pastebin.com/Y0856qzP

zbw8388:
Thank you for the info! 

> Are your latest commits ideal? It seems to make a lot of use of the magic 125 number, and seems to wait a lot. Would directly debouncing/timeout the incoming vsocde scroll position be a cleaner solution? For me it required less code.

This 125 is vscode's default scrolling duration as you can see below. For some reason, the smooth scrolling duration for your vscode is a tad bit longer than 125ms: 677 - 530 = 147ms for the first scroll you made, which causes the "bouncy" behavior at your end. This should also explain your issue.

https://github.com/microsoft/vscode/blob/125ae4af2999c13096d8c1bb9d188b4c6e24bd0d/src/vs/base/browser/ui/list/listView.ts#L350
https://github.com/microsoft/vscode/blob/380ad48e3240676b48d96343f8ad565d4fea8063/src/vs/editor/common/viewLayout/viewLayout.ts#L16

I looked around and found that there's a way to stop autocommand, which might be the most ideal solution: https://neovim.io/doc/user/autocmd.html#:noautocmd
I think once we implemented this we then don't need any of the 125 and we could use debounce instead.

wenfangdu:
@zbw8388 @theol0403 
I found this PR through [this comment](https://github.com/vscode-neovim/vscode-neovim/pull/1009#issuecomment-1220933485), seems like it's causing `workbench.action.compareEditor.nextChange` and `workbench.action.compareEditor.previousChange` not working properly, v0.0.90 has this issue, v0.0.89 does not, I suppose this got merged into master in v0.0.90?

theol0403:
@wenfangdu this PR has not been merged yet. I was referring to https://github.com/vscode-neovim/vscode-neovim/pull/885 and https://github.com/vscode-neovim/vscode-neovim/pull/919. Either way, nothing merged in master should affect those commands.

wenfangdu:
@theol0403 I did a commit bisect, the code from #919 caused that issue.

theol0403:
@zbw8388 
Hey, how's it going? I assume you are busy (as I am), but do you know if you will have time to finish this up? 

Currently the release is in a weird place because there were the first scrolling PRs that were merged but this one fixes them, so I can't release without this. I might make a hotfix release for the other PRs though. 

This PR has a lot of potential so I would love to see it merged. If not, it will be a while until I can try to tackle it myself.

Thanks!

zbw8388:
Hi! I'm sorry about the inactivity. I'm super busy these days but I always wanted to finish this PR. It's hard to find a large chunk of time currently. I remembered I was dealing with a VSCode API limitation last time I was working on this. I might be able to look at it in November but it depends. Please release a hotfix if this is currently blocking the development! At the same time, I'm wondering if it would be possible to revert and split the merged scrolling modification commit into an individual branch so that normal development can still happen.

theol0403:
@zbw8388 thanks for the reply! I totally understand, especially regarding finding a large chunk of time to work on it lol. This project is not the kind of project you can expect to make significant progress with two-hour chunks. 

I've released all pending PRs and have put the commits holding up development back in master (sync viewport, delete c-u). I imagine I won't have time to make many more PRs until december so hopefully this lines up with this PR being complete.

I've gone ahead and merged master back into this PR, to reduce friction for you to get started when you have time. I modified a bit of code doing so, but if you don't like the changes I don't mind if you change it back.

MohamedOsman1998:
hi guys, great work here, any updates on this?

archilkarchava:
These look somewhat related – https://github.com/vscode-neovim/vscode-neovim/issues/1131#issuecomment-1434745398

adueck:
Hey everyone don't want to pressure the awesome people working on this but I would LOVE to see this happen. 🙇‍♂️

MarcosH911:
I discovered this extension recently and I find it way better than the Vim extension - thanks for the awesome work!

Any updates with this? I would like to remap <C-d> to <C-d>zz and <C-u> to <C-u>zz

thetechank:
Just wanted to check, would there be an update to make the scroll with folds work. It is a major issue in my workflow and navigating folds if they open automatically is a pain. Otherwise the extension is fantastic!!!

tom-pollak:
This is a nice workaround for now, shouldn't expand folds unless you land directly on them. Remove the centred scrolling if not needed.

https://github.com/vscode-neovim/vscode-neovim/issues/983#issuecomment-1960709776

Issue #225, opened 2020-03-13
Remote Extension: Does not work with SSH (OSX -> Ubuntu)


christo-ph:
Running Visual Studio Code on OSX.

Development env on remote Ubuntu.

Settings for User and Remote[SSH: ubuntu] done.

1. Local (User) to local nvim and init.vim -> works
1. Remote (SSH: Ubuntu) to remote nvim and remote init.vim -> does not work
1. Remote without any settings -> does not work

I am unable to insert anything to the editor. Also `:` or `/` do not seem to do anything.

asvetliakov:
I don't have any plans to support remote neovim. Perhaps it's possible to force extension to be installed remotely through VSCode and use remote extension host, but you'll be anyway disappointed for lag & slowness

hgezim:
Porting those (vim) settings from machine to machine instead of keeping them on remote host is quite unfortunate :(

theol0403:
Maybe #668 can help, but I know it's not exactly what you are looking for. I don't think there is a good solution for using remote nvim instance, nor would it work very well (latency).

still-dreaming-1:
I want this so badly. It would be the killer app of both VS Code and Neovim. The way Remote Extension works is so much better than anything similar other IDEs support. I can't seem to configure language servers properly in Neovim, but it works great in VS Code. But I can't just use VS Code without Neovim, that would be nuts. I'm so used to just sshing into a machine and using Neovim from there, and so this would be a great way to effectively do the same thing from VS Code. I'm confused, the reason you don't want to support it is because you are concerned it will be slow? Whenever I use VS Code with the Remote Extension it is not slow normally, so why would it all of a sudden become slow just because the server side is communicating a Neovim running on that same server?

theol0403:
I think this is alright possible with https://github.com/vscode-neovim/vscode-neovim/pull/1000

still-dreaming-1:
> I think this is already possible with #1000

So it's just waiting for someone to review and merge it, since Aug 10, 2022?

PR #1000, opened 2022-08-10
add support for Remote Development and GitHub Codespaces

### LINKED ISSUE #262: Allow installation on Github Codespaces
Hello!

I'd like to try this plugin on Github Codespaces, but the extension says it cannot be installed on the server.

I'm guessing (hoping) maybe it could work like how it works with WSL?

Pretty please with a cherry on top! <3

exil0867:
- Resolves #262 by making the extension run where the workspace is located.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
     "icon": "images/icon.png",
     "publisher": "asvetliakov",
     "extensionKind": [
-        "ui",
         "workspace"
     ],
     "version": "0.0.90",


theol0403:
I was thinking, too bad this isn't a setting. Personally, I'd rather want nvim to run on my local machine, because I have config/plugins/etc that work regardless of remote desktop. However, if you only work with a remote desktop, I see the appeal of having nvim run on the remote.

Maybe we should keep default behavior, but add to the readme how you can override the setting so that  you can run nvim on the remote?

exilvm:
I totally agree.

still-dreaming-1:
What is the status of this? Looks like it's just awaiting a review before it can be merged.

still-dreaming-1:
Possibly this would fix #225.

theol0403:
This should not be merged, because it disables the behaviour people expect (nvim runs on local machine). You can apply the effect of this PR using your settings.

still-dreaming-1:
> 

Ok, it seemed like something like that might be the case. But I'm still confused which settings I need to modify.

theol0403:
> Ok, it seemed like something like that might be the case. But I'm still confused which settings I need to modify.

https://github.com/vscode-neovim/vscode-neovim/issues/262#issuecomment-1208062772

still-dreaming-1:
It's working!

> > Ok, it seemed like something like that might be the case. But I'm still confused which settings I need to modify.
> 
> [#262 (comment)](https://github.com/vscode-neovim/vscode-neovim/issues/262#issuecomment-1208062772)
This part did it from a settings.json file, thanks!:
```
"remote.extensionKind": {
    "asvetliakov.vscode-neovim": ["workspace"]
}
```

still-dreaming-1:
Actually I think this pull request might still be a good candidate to be merged. For one thing, it doesn't prevent this extension from using a local Neovim in the case where everything was local to begin with, as you were not using the remote extensions.

Additionally, if I'm not mistaken, this change would only make it possible, but not required to install this plugin on the remote instance. The reason I am saying this is, even with the change I made to my settings, I still needed to use a little blue button in the VS Code Extensions UI to actually install it on the remote machine. Of course it would be good to test this before merging, but it might still be possible to install and use this extension locally instead, even with this change, even while using the remote extensions.

If that is the case, it seems like a great idea as part of the reasons people want to use the remote extensions is so they can install and run their other VS Code extensions on the remote machine. It's not just about accessing remote files you want to edit.

justinmk:
This PR needs a clear Problem/Solution statement. vscode-neovim already works with vscode remote, so the title is very misleading.

From what I can gather, is this the use-case being solved here https://github.com/vscode-neovim/vscode-neovim/issues/262#issuecomment-1208062772 :

> so i no longer need to install Neovim locally.

still-dreaming-1:
> vscode-neovim already works with vscode remote

Technically you can get it to work, but you have to know about an obscure and strange setting you need to change to get it working. You have to edit this setting by hand as well, it's not editable through the menu. I went all these years believing I can't use VS Code with a remote Neovim in this way because of this shortcoming.

My understanding, which could be wrong, is that if this were merged, it would simply make it possible to install this extension on a remote machine, through the GUI using the extensions panel on the side, without having to change any settings. I don't think it would ever break using a local Neovim, you would simply install the extension on your local machine and not the remote one.

I'm not aware of any use case that would work worse with this merged, but then again I don't understand all the ramifications of changing the "extensionKind" from "ui" and "workspace" to just "workspace".

justinmk:
> Technically you can get it to work, but you have to know about an obscure and strange setting you need to change.

I think it depends on what "remote" means. So part of the Probem/Solution statement for this PR, needs to say exactly what steps this applies to. 

I didn't have to set a strange setting, here's my vscode config: https://github.com/justinmk/config/blob/master/Library/Application%20Support/Code/User/settings.json

I use vscode remote-ssh https://code.visualstudio.com/docs/remote/ssh and it just works.

still-dreaming-1:
> I use vscode remote-ssh https://code.visualstudio.com/docs/remote/ssh and it just works.

I wonder if you might have an additional settings.json file somewhere with this override setting:
```
"remote.extensionKind": {
    "asvetliakov.vscode-neovim": ["workspace"]
}
```
Perhaps it is in the settings.json file that is on and for the remote machine (mine is on the remote machine at ~/.vscode-server/data/Machine/settings.json), or perhaps the workspace settings.json that can be in the workspace on the remote machine?

still-dreaming-1:
I think I jumped in here and started commenting without explaining what I was trying to achieve. I apologize for that. I'm trying to connect this extension to the installation of Neovim on the remote machine. If the Neovim being used by this extension does not have access to the environment where the code lives, it will be extremely crippled.

theol0403:
> If the Neovim being used by this extension does not have access to the environment where the code lives, it will be extremely crippled.

Can you give an example? Remember that the neovim instance is not aware of the file location/cwd/project, it is just provided a scratch buffer with the editor contents.

still-dreaming-1:
> Can you give an example? Remember that the neovim instance is not aware of the file location/cwd/project, it is just provided a scratch buffer with the editor contents.

I guess currently I can't come up with any great use cases for running Neovim remotely in this way. I was hoping it would unlock the full power of Neovim, but I am able to run Neovim remotely like this from this extension with the right settings, and it is currently not making what I want to do possible anyways. So it seems like a great topic of conversation to explain what I want to do in more detail, but until those use cases are supported, I don't think merging this PR will help anyone.

I was hoping to have the option of either doing things the VS Code way, or the Neovim way or both. I would like to be able to use a Neovim file tree plugin and use that instead of the built in VS Code file explorer. I would like to be able to use things like fzf through a Neovim plugin. Each person's reason for wanting to integrate Neovim into VS Code will be different. In my case I am almost completely happy with Neovim by itself, I just can't seem to configure the LSP and LSP based auto complete to work the way I want, even after years of trying. But that part works great in VS Code.

I also really want to be able to use an Vim plugin I made that greatly enhances my productivity. One way it can be used is while the cursor is over a particular word, you can press <leader>* to search for that word in the contents of all the files matching the same file type you have open. This opens a new window and places the search results in that window. It shows the path to the file, the line number it appears on, and the contents on that line. This is intentionally not a regex search as I hate reges, it is just a raw search for the characters you entered. Then you can use vim regex search within that window and buffer to search more specifically for what you want from there using the forward slash key the normal Vim way. This buffer is intentionally editable. You can edit it in any way you want. This is very useful because you can delete lines that are not relevant to what you were searching for, or that were relevant, but now you are done dealing with it and want it to disappear from the list. You can also press gF while the cursor is over a file from a search result line to go to that exact file and line number in the window above without the search result window disappearing. You can also edit the line contents directly in the search result window and any time you save from this buffer/window, it will modify that line in that file to match the contents you have there. So you can search for something, edit all or some of the results right there, and then save them all at once or discard the changes by closing the window without saving. I think of it like extradimensional editing. It is the perfect example of how modal editing is useful not only for editing a particular file, but also as a way of controlling the editor itself. It also illustrates the power of being able to make your own plugins and gives you a feel for why someone would not want to give up this power within Neovim even when trying to integrate it into something like VS Code.

theol0403:
> I would like to be able to use a Neovim file tree plugin and use that instead of the built in VS Code file explorer.

This will never be possible, because there isn't the ability to render arbitrary elements in VSCode. There are very specific and limited APIs for working with VSCode elements like quickpick, editor, etc. Only option is webviews or custom explorer filesystem provider (still using vscode ui).

> I would like to be able to use things like fzf through a Neovim plugin. 

Not possible for same reasons. 

> you can press * to search for that word in the contents of all the files matching the same file type you have open.

This is possible, but would need to be implemented 100% on the vscode side. 

>  You can also press gF while the cursor is over a file from a search result line to go to that exact file and line number in the window above without the search result window disappearing

Again, possible on the vscode side if the search results provides a "go to definition" implementation. 

You can never turn VSCode UI into something modal, because of the limitations of the API.

still-dreaming-1:
So really this has all the same limitations all the other IDE Vim emulators have. It's just more accurate since it is not an emulator.

theol0403:
Correct, and it supports many useful nvim plugins that augment motions & textobj, lua config, etc.

justinmk:
> > I would like to be able to use a Neovim file tree plugin and use that instead of the built in VS Code file explorer.
> 
> This will never be possible, because there isn't the ability to render arbitrary elements in VSCode.

Neovim file tree plugin is just a buffer, so this should be possible by showing such a buffer in a vscode editor tab.

xiyaowong:
@still-dreaming-1 Have you investigated the built-in features of VSCode? Have you forgotten about the extension marketplace of VSCode? If you are unsatisfied with only a portion of nvim's lsp functionality, why not work on improving nvim-lsp?

Many people switch from nvim to VSCode and find it different and not as user-friendly. They raise various needs and requirements but are unwilling to spend some time learning the features of VSCode.

> I would like to be able to use a Neovim file tree plugin and use that instead of the built in VS Code file explorer

Can you list some features you are not satisfied with? VSCode's file explorer supports Git and diagnostic decorations, and all the operations can be customized with shortcuts. The primary requirement for most people is simply shortcuts, but thanks to the flexible shortcut design of VSCode, all shortcuts can be configured just like the common file tree plugins in nvim.

> search

Have you tried using these for your search needs?
image

Here is my relevant configuration.

```vim
map <silent> /         <Cmd>call VSCodeCall('actions.find')<CR>
map <silent> <space>/  <Cmd>call VSCodeCall('actions.find')<CR>
map <silent> <space>fw <Cmd>call VSCodeCall('workbench.action.quickOpen', '% '.expand('<cword>'))<CR>
map <silent> <space>fb <Cmd>call VSCodeCall('workbench.action.quickOpen', 'edt active ')<CR>
map <silent> <space>fo <Cmd>call VSCodeCall('workbench.action.quickOpen', '@')<CR>
map <silent> <space>fd <Cmd>call VSCodeCall('workbench.action.quickOpen', 'debug ')<CR>
map <silent> <space>fr <Cmd>call VSCodeCall('workbench.action.quickOpen', 'task ')<CR> " run
map <silent> <space>ft <Cmd>call VSCodeCall('workbench.action.quickOpen', 'term ')<CR>
```

justinmk:
PR as-is isn't mergable and hasn't addressed https://github.com/vscode-neovim/vscode-neovim/pull/1000#issuecomment-1424052985

theol0403:
I kept this open as a reminder to add related documentation on how to run the ext remotely to the README

PR #1015, opened 2022-08-19
feat: list.toggleKeyboardNavigation => list.find

### LINKED ISSUE #995: [Bug] In VSCode v1.70.0, `workbench.list.automaticKeyboardNavigation` is gone
image

image

This causes explorer list navigation not functional.

wenfangdu:
closes #995

### DIFF
--- a/package.json
+++ b/package.json
@@ -556,6 +556,11 @@
                 "when": "listFocus && listSupportsFind && !inputFocus"
             },
             {
+                "key": "enter",
+                "command": "list.closeFind",
+                "when": "listFocus && treeFindOpen && inputFocus"
+            },
+            {
                 "key": "r",
                 "command": "renameFile",
                 "when": "explorerViewletVisible && filesExplorerFocus && !explorerResourceIsRoot && !explorerResourceReadonly && !inputFocus"
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
         "email": "devpieron@gmail.com"
     },
     "engines": {
-        "vscode": "^1.69.0"
+        "vscode": "^1.70.0"
     },
     "keywords": [
         "keybindings",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -551,14 +551,9 @@
...
             {
-                "key": "/",
-                "command": "list.toggleKeyboardNavigation",
-                "when": "listFocus && !inputFocus && listSupportsKeyboardNavigation"
-            },
-            {
-                "key": "escape",
-                "command": "list.toggleKeyboardNavigation",
-                "when": "listFocus && inputFocus && listSupportsKeyboardNavigation && !notebookCellListFocused"
+                "key": "oem_2",
```
theol0403, line 554: Why do you use `oem_2`? Why not `/` for readibility?
Also please re-add `!inputFocus` as it can interfere with file renaming in explorer.

theol0403:
I was also looking for some bindings to select the file after opening. I have found two options:

```json
            {
                "key": "enter",
                "command": "list.select",
                "when": "listFocus && inputFocus && listSupportsFind && treeFindOpen"
            },
```

This directly opens the file, but leaves the find widget open.

```json
            {
                "key": "enter",
                "command": "list.closeFind",
                "when": "listFocus && inputFocus && listSupportsFind && treeFindOpen"
            },
```
This closes the find widget, but requires two presses of enter to open the file.

I vote for the second option.

theol0403:
Ref: https://github.com/microsoft/vscode/pull/152481

wenfangdu:
I've updated the PR, but for the `enter`, I think [this feature](https://github.com/microsoft/vscode/issues/158643) is lacking from upstream.

justinmk (approved):
anything blocking this?

wenfangdu:
@justinmk From my side, no.
@theol0403 Is there anything missing?

theol0403:
Why can't you add one of the enter bindings? Don't they fix the problem? Just because you made an issue on VSCode does not mean it will be addressed, in fact, I can't think of a single feature request made by this project that wasn't ignored by vscode devs.

wenfangdu:
@theol0403 Added.

theol0403 (approved):

PR #1016, opened 2022-08-19
fix: settings.clean no longer overides settings.customInitFile

goolord:
followup on https://github.com/vscode-neovim/vscode-neovim/pull/952

this just moves the arguments around so that --clean doesn't override the config passed by -u


theol0403 (approved):

goolord:
i appreciate this feature being merged in, i've been using it to great effect at https://github.com/goolord/dotfiles/blob/main/home/.config/nvim-code/init.lua

i feel very comfortable and at home in remote sessions with my coworkers now :-)

PR #1028, opened 2022-09-03
docs: update quickfix section

nix6839:
Support custom code action menu shortcut on VSCode >= 1.71.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,31 +261,32 @@ See gif in action:
 
 ![multicursors](/images/multicursor.gif)
 
-### Keyboard Quickfix
+### Code Action Menu Shortcut
 
-By default, the quickfix menu can be opened using <kbd>z=</kbd> or <kbd>C-.</kbd>. However, it is currently
-[not possible](https://github.com/microsoft/vscode/issues/55111) to add mappings to the quickfix menu, so it can only be
-navigated with arrow keys. A
-[workaround vscode extension](https://marketplace.visualstudio.com/items?itemName=pascalsenn.keyboard-quickfix) has been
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -261,32 +261,33 @@ See gif in action:
...
 ```
 
 and add to your init.vim:
 
 ```vim
-nnoremap z= <Cmd>call VSCodeNotify('keyboard-quickfix.openQuickFix')<CR>
+nnoremap z= <Cmd>call VSCodeNotify('editor.action.quickFix')<CR>
 ```
 
+You can now use <kbd>z=</kbd> to open the code action menu, and <kbd>C-n</kbd> and <kbd>C-p</kbd> to navigate it.
```
justinmk, line 289: this is a markdown doc, why not use markdown?
```suggestion
You can now use `z=` to open the code action menu, ...
```

nix6839: Visual difference and markdown allows inline html. See <https://daringfireball.net/projects/markdown/syntax#html>
image

nix6839: And I want to ask. Do `<code>` and `<kbd>` mean the same thing? You can see following:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code

theol0403: We should remove that section entirely from readme and add to package.json bindings.

theol0403:
We should remove that section entirely from readme and add to package.json bindings.

nix6839:
I agree! I'll close the issue.

PR #1029, opened 2022-09-03
feat: add shortcut for navigate code action menu

nix6839:
Support shortcut for navigate custom code action menu on VSCode >= 1.71.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -656,6 +656,16 @@
                 "when": "inQuickOpen"
             },
             {
+                "key": "ctrl+n",
+                "command": "selectNextCodeAction",
+                "when": "codeActionMenuVisible"
+            },
+            {
+                "key": "ctrl+p",
+                "command": "selectPrevCodeAction",
+                "when": "codeActionMenuVisible"
+            },
+            {
                 "command": "vscode-neovim.send",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1213,6 +1213,16 @@
                 "command": "vscode-neovim.complete-selection-cmdline",
                 "key": "tab",
                 "when": "neovim.mode == cmdline_normal || neovim.mode == cmdline_insert || neovim.mode == cmdline_replace"
+            },
+            {
+                "key": "ctrl+n",
```
justinmk, line 1218: is this analogous to `]q` from https://github.com/tpope/vim-unimpaired ? vim-unimpaired is a very well-established convention so I suggest we use its mappings, i.e. `]q` and `[q`.

justinmk (approved):
Great to eliminate these kinds of things from the readme in a simple way.

package.json
theol0403: No, this is for navigating quick pick (code actions mostly). However, connecting unimpaired bindings to vscode like `]q` to jump between LSP errors would be a great addition!

theol0403:
Thanks for the PR!

Issue #794, opened 2021-12-28
"Toggle Vim Mode" setting (enable/disable extension)


azaslavsky:
I'm not sure if this is possible given the architecture of this plugin (ie, defers to an external instance of neovim for actual implementation), but it would be nice to have a programmable `Toggle Vim Mode` hotkey, similar to what is currently available in the `vscodevim` extension.  There are a number of cases in which I would like to completely "disable" vim and go back to regular old VSCode editing, but this is not currently possible.

trkoch:
If by toggling we define Vscode stays in insert mode, my wild guess would be that this should be feasible. You can already sort of emulate it by not forwarding `<esc>` to the extension, this way requiring `<ctrl+c>` to be pressed.

bencodesall:
+1 to this feature request!

Wonder is something like this is feasible:

1. `<command+shift+P>` to bring up "command palette"
2. start typing `vim` or `toggle` and brings up `Toggle vscode-neovim` on off (with indicator on bottom status bar)

I believe this is how `vscodevim` does it. That would be sufficient.

P.S. Love that I can now use my `neovim` setup within VS Code. Thanks for the work.

aquadesk:
> 



> If by toggling we define Vscode stays in insert mode, my wild guess would be that this should be feasible. You can already sort of emulate it by not forwarding `<esc>` to the extension, this way requiring `<ctrl+c>` to be pressed.

it maybe possible but in some situations, it leave insert mode and we should keep getting into it.
i.e. start editing different files or click on search results..

Valleeh:
Hey, sometimes in pair programming the plugin hinders me, so this functionality would be great.
After waiting for about half a year, I decide to take my luck into my own hands. 
In my fork, it works. I added a switch to keep the insert mode even if Esc or Buffer(file)change happens.
I added a second switch that just makes insert mode persist when buffer change happens.
And I added a setting (which oddly can't be changed(which I think is a bug in the app programming) that switches to insert mode instead of VisualMode when text is selected with the mouse.
I guess it's not really PR ready yet, and there's no testing and so on. I wanted to ask if this functionality is wanted, and how it should be implemented, then i would take care for it if wanted. 
https://github.com/Valleeh/vscode-neovim
Best
Valentin Wunderlich

justinmk:
[...]

PR #1033, opened 2022-09-09
Neovimtoggle

Valleeh:
Two Toggles:
1. "Disable" vscode-neovim by not processing Esc and stay in insert mode while jumping between files
2. Just stay in insert-mode till Esc is recievied. Buffer changes while in insert mode will not go switch back to Normalmode (out of scope will be handled in https://github.com/vscode-neovim/vscode-neovim/pull/1050)


package.json
```diff
@@ -259,6 +259,14 @@
             }
         },
         "commands": [
+            {
+                "command": "vscode-neovim.toggle",
+                "title": "Neovim Toggle: Stay in InsertMode Forever"
+            },
+            {
+                "command": "vscode-neovim.toggle2",
+                "title": "Neovim Toggle: Stay in InsertMode OnBufferChange"
```
justinmk, line 268: could take an arg instead of adding an extra command?

Valleeh: Conceptually, I think it would be better if the second toggle (stay in InsertMode on BufferChange) was a setting, since it's more of a decision you make once and don't change that often. The first toggle, on the other hand, should be a toggle. If you agree, I will add the second toggle to the settings.
But I think I need your help for that, because when I tried to add a setting some time ago, it seemed like I implemented it wrong or something.

package.json
```diff
@@ -259,6 +259,14 @@
             }
         },
         "commands": [
+            {
+                "command": "vscode-neovim.toggle",
```
trkoch, line 263:
```suggestion
                "command": "vscode-neovim.toggleStayInsert",
```

src/mode_manager.ts
```diff
@@ -19,7 +19,8 @@ export class ModeManager implements Disposable, NeovimRedrawProcessable, NeovimE
      * True when macro recording in insert mode
      */
     private isRecording = false;
-
+    public neovimToggle = true;
```
trkoch, line 22: Can we make this more descriptive. How about?

```suggestion
    public neovimToggleStayInsert = true;
```

trkoch (changes requested):
Thank you for taking the initiative to implement this! I haven't tested your changes locally yet. Will revisit. In general, please make sure not to change any defaults.

package.json
trkoch: `toggle2` will probably run into https://github.com/vscode-neovim/vscode-neovim/issues/642. I suggest to focus on implementing just `toggle` (forever) to avoid [scope creep](https://mtlynch.io/code-review-love/#5-narrowly-scope-changes).

src/mode_manager.ts
justinmk: "stay in insert" is not really a correct description, it's more of a side-effect. The actual thing being done is that vscode-neovim is being disabled. Right?

Valleeh: both in a way right. To actually disable the extension, one must go into insertmode after switched the toggle. So currently it describes the name @trkoch want to give it.
So the Toggle would need also to send a "GoToInsertMode"-Request.
For understandability and a clean feature @justinmk 's approach would be better. 
If you both agree i go with neovimToggle(since neovimEnable is not really what it does, it also disables the Extension. And neovimToggleStayInsert is what it does but doesn't reflect the intent)

package.json
Valleeh: I think you are right. I will remove all the StayInInsertModeOnBufferChange functionality from this pull request. But I'd still like to implement it in another pull request. In the end, after implementing both functions, I won't touch any of it because StayInInsertModeOnBufferChange is what I really want, and once set it's super handy. This way, the transition from normal Vscode to using the keyboard over the extension feels very smooth to me. When colleagues are remotely connected or I need to do something in normal Vscode fashion, I just press "i"... and as long as I don't press Escape, the Vscode-Neovim extension doesn't interfere with anything I'm doing (as it would if I were changing the buffer without this feature). 
The feature actually works fine, but let's see.... have not checked the problems in #642

src/mode_manager.ts
justinmk: If the intention is to toggle (almost) all vscode-neovim functionality, then let's name it as such. Whatever internal compromises are made (hopefully temporary) are irrelevant.

trkoch: Thank you for your insight! Upon reflection I agree.

> The actual thing being done is that vscode-neovim is being disabled.

theol0403:
Let me know if this is no longer "wip" and I will take a look

Valleeh:
Hey,Sorry i was busy, also my dev-setup had an issue. After switching to Linux, I worked on it again and was searching how to send a command/msg/promise to the modemanager that it should enter insertmode. I debbuged trough and saw some possibility’s but nothing felt right, i would really appreciate if someone could give me a clue on how to do this properly. Best ValentinOn 10. Oct 2022, at 23:38, Theo Lemay ***@***.***> wrote:﻿
Let me know if this is no longer "wip" and I will take a look

theol0403:
I don't think you need to do that. Just use the nvim api to call the `startinsert` command. 
Fyi part of this PR is replaced by #1050, so you can now focus on the "toggle" aspect.

Valleeh:
feature should work now as expected

src/typing_manager.ts
```diff
@@ -52,6 +52,21 @@ export class TypingManager implements Disposable {
         this.disposables.push(commands.registerCommand("vscode-neovim.send", this.onSendCommand));
         this.disposables.push(commands.registerCommand("vscode-neovim.send-blocking", this.onSendBlockingCommand));
         this.disposables.push(commands.registerCommand("vscode-neovim.escape", this.onEscapeKeyCommand));
+        this.disposables.push(
+            commands.registerCommand("vscode-neovim.toggle", () => {
```
theol0403, line 56: can you factor this out to a `onToggleCommand` function

src/typing_manager.ts
```diff
@@ -52,6 +52,21 @@ export class TypingManager implements Disposable {
...
+                this.main.modeManager.neovimToggle = !this.main.modeManager.neovimToggle;
+                let batch: [string, ...unknown[]][];
+                if (!this.main.modeManager.neovimToggle) {
+                    this.logger.debug(`${LOG_PREFIX}: Disable Vscode-neovim`);
+                    batch = [["mode_change", ["insert"]]];
```
theol0403, line 61: I don't have time to try right now, but before I merge this I'll want to take a look to see if there is a better way to enter insert mode. This seems like a bit of a hack, ideally we just tell nvim to enter/leave insert mode directly using `startinsert` `stopinsert`.

src/typing_manager.ts
```diff
@@ -192,8 +207,10 @@ export class TypingManager implements Disposable {
     private onEscapeKeyCommand = async (key = "<Esc>"): Promise<void> => {
         // rebind early to store fast pressed keys which may happen between sending changes to neovim and exiting insert mode
         // see https://github.com/asvetliakov/vscode-neovim/issues/324
-        this.isExitingInsertMode = true;
-        await this.onSendBlockingCommand(key);
+        if (this.main.modeManager.neovimToggle) {
```
theol0403, line 210: This will break `<C-o>` in insert mode from working. Id say this would be fine if this PR totally disables neovim, but currently it just keeps nvim stuck in insert mode, meaning other commands like `<C-r>`/`<C-w>`/etc still work.

Probably just do `|| key !== "<Esc>"`.

src/typing_manager.ts
```diff
@@ -52,6 +52,21 @@ export class TypingManager implements Disposable {
...
+                let batch: [string, ...unknown[]][];
+                if (!this.main.modeManager.neovimToggle) {
+                    this.logger.debug(`${LOG_PREFIX}: Disable Vscode-neovim`);
+                    batch = [["mode_change", ["insert"]]];
+                } else {
+                    this.logger.debug(`${LOG_PREFIX}: Enable Vscode-neovim`);
+                    batch = [["mode_change", ["normal"]]];
+                }
+                this.main.modeManager.handleRedrawBatch(batch);
+                this.main.cursorManager.handleRedrawBatch(batch);
```
theol0403, line 67: Yeah this just seems like a hack. How come the other managers don't want a redraw batch as well? Really all this does is sets cursor to line and disables `type` event binding, which doesn't actually put nvim in insert mode.

theol0403 (changes requested):
Thanks for your work! I left some reviews.

I'd also not be opposed to having an `enable` and `disable` command, instead of `toggle`, so it is more clear exactly which state you are switching into.

src/typing_manager.ts
Valleeh: in modemanger it calls 
``` 
commands.executeCommand("setContext", "neovim.mode", this.mode);
this.eventEmitter.emit("neovimModeChanged", modeName);
``` 
i can't find the reference to _setContext_.

Could be a hack, i mean it's working and my intent was not tell nvim to enter insertmode i'd rather tell the vscode extension to go into "insertmode" which is basically as statet in the docs act like normal vscode.
It's really my first interaction with extension-programming  and my first time typscript and so on. Help is highly appreciated. I tried to figure out how to send via api _startinsert_ didn't find something useful.

theol0403: `setContext` is just used for vscode keybindings - it's what you use when you do `when: "neovim.mode == insert"`.

Tricking the extension to think it's in insert mode without actually placing nvim in insert mode can cause problems and doesn't serve a purpose (I think it will break insert mode bindings as well). If we tell nvim to go to insert mode, both sides will be in agreement. 

I'm not totally sure but it might be something like `client.command('startinsert')`.

src/typing_manager.ts
Valleeh: tried with both setting "Vscode-neovim: Use Ctrl Keys For Insert Mode" on and off.. the keys C-r, C-w and C-o do what is stated in the Keyboard shortcut stated. It only enters this function when pressing <Esc> or when i do jk which is the same. maybe it changes after restart Vscode completly.. I just reloaded the window to let the  "Vscode-neovim: Use Ctrl Keys For Insert Mode" Setting going active

src/typing_manager.ts
theol0403: If this PR actually decoupled nvim from vscode (killed the subprocess, disabled all bindings, killed all listeners, then re-init the buffers upon enable), then it would be ok to have both sides disagree on the state.

But since all this PR does is make you stuck in insert mode, the model where vscode is simply a "window into nvim" should not be broken.

src/typing_manager.ts
Valleeh: would be highly appreciated.

src/typing_manager.ts
theol0403: - settings only apply upon vscode restart
- all the keys do work normally when `toggle` is true
- C-r/C-r works normally when `toggle` is false
- C-o does NOT work when `toggle` is false

src/typing_manager.ts
theol0403: Btw I believe one use of `this.eventEmitter.emit("neovimModeChanged", modeName);` is that it calls `cursorManager.modeChanged` which sets the cursor to be a line instead of block. It also calls `typingManager.modeChanged` which unregisters `type` event.

Valleeh: > I'm not totally sure but it might be something like client.command('startinsert').

It works. can you tell me where to find references or the related code for this. I was really stuck and feel stupid now

theol0403: It's tricky. Most of the commands you can find in vim help. `:h startinsert`. It's just a normal ex-command.
Node-client docs are pretty bad, but that is where you find `command` to call a command. https://neovim.io/node-client/classes/NeovimClient.html
Also try to use vscode typescript go to definition for random stuff in the codebase.
For specific ui commands there is `:help ui` and `:help api`. The node client is just a wrapper for `api`.

Valleeh: thanks alot!!

src/typing_manager.ts
Valleeh: okay, i could reproduce after i disabled karabiner on mac.
It's as you said. Fixed now

src/typing_manager.ts
Valleeh: done

Valleeh:
> to having an `enable` and `disable` command, instead of `toggle`, so it is more clear exactly which state you are switching

i would really like go with toggle since normally i like one shortcut and not two.  

Pro 'enable', 'disable': 
clear State switching
Following your 'enable' and 'disable' approach a toggle would be still possible by introducing https://marketplace.visualstudio.com/items?itemName=rebornix.toggle from user side.

Con:
Two shortcuts
Issue https://github.com/vscode-neovim/vscode-neovim/issues/794 refers to other VimMode-Extension which does it similarly.

If wanted i can also include 'enable', 'disable' next to 'toggle'.

tobealive:
> If wanted i can also include 'enable', 'disable' next to 'toggle'.

Having both seems to give all possibilities without need for compromises 👍

src/typing_manager.ts
```diff
@@ -188,12 +195,32 @@ export class TypingManager implements Disposable {
         this.registerReplacePrevChar();
         await this.onSendCommand(key);
     };
+    public onDisableCommand = (): void => {
+        this.neovimToggle = true;
+        this.client.command("stopinsert");
+    };
+
+    public onEnableCommand = (): void => {
```
justinmk, line 203: this could take an `enable: 'enable'|'disable'|'toggle'` param, then all the commands could use a single function.

src/typing_manager.ts
```diff
@@ -52,6 +56,9 @@ export class TypingManager implements Disposable {
         this.disposables.push(commands.registerCommand("vscode-neovim.send", this.onSendCommand));
         this.disposables.push(commands.registerCommand("vscode-neovim.send-blocking", this.onSendBlockingCommand));
         this.disposables.push(commands.registerCommand("vscode-neovim.escape", this.onEscapeKeyCommand));
+        this.disposables.push(commands.registerCommand("vscode-neovim.toggle", this.onToggleCommand));
+        this.disposables.push(commands.registerCommand("vscode-neovim.enabled", this.onToggleCommand));
+        this.disposables.push(commands.registerCommand("vscode-neovim.disabled", this.onToggleCommand));
```
justinmk, line 61: pointing to the wrong function?

Valleeh: thanks for the review!
it's changed now

src/typing_manager.ts
Valleeh: Changed as requested

justinmk (approved):
usually the readme/docs don't need/want to mention every command and setting (because vscode already exposes that in its UI), but this one is probably worth calling out in the readme.

theol0403:
> usually the readme/docs don't need/want to mention every command and setting (because vscode already exposes that in its UI), but this one is probably worth calling out in the readme.

It's at the top of the list so should be fairly easy to discover. It doesn't seem to naturally fit anywhere in the readme so I think its good as is.

theol0403 (approved):
Thank you for the PR!

aarondill:
I am aware that this is a merged commit, and feedback has already been left, but I'd like to put in my two cents as a new user. I am attempting to integrate my vim knowledge into my VSCode setup, and tried VSCodeVim first, but found it's support for the vim runpath to be lackluster compared to your extension, however a large consideration was the seeming lack of a toggle in the readme, having to navigate the issues of the repo to determine its presence (or lack thereof) is not a great expectation on new users, and considering the difficulty of using vim for the first time, a simple mention of this command in the readme to ease new users would be a good idea (IMO).

PR #1043, opened 2022-09-30
fix(cursor_manager): Make mouse clicks extend selection in visual mode

trkoch:
In the early times of this extension, there were attempts to make mouse clicks start a visual selection, similar to `set mouse=a` in Vim. This behaviour is controlled by the `mouseSelectionStartVisualMode` setting which is off by default. As everybody can probably attest who has given this option a try, it is broken badly.  I don't think anyone is actually using it on a daily basis.

However, it turns out that parts of the code being written can be very useful. This PR enables the fraction of this feature which works smoothly by default and removes the setting `mouseSelectionStartVisualMode`.

Example

* Intention: You want to select a couple of lines using your mouse
* Press `V` where you want your selection to start
* Scroll around and click where you want your selection to end
* Operate on the selection as you please

Personally, I find this very useful to select large chunks of code, especially when I want to rely on the (visual) indent guide.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,11 +115,6 @@
                     "default": true,
                     "description": "Use Ctrl keys in insert mode"
                 },
-                "vscode-neovim.mouseSelectionStartVisualMode": {
-                    "type": "boolean",
-                    "default": false,
-                    "description": "Mouse selection starts visual mode"
-                },
                 "vscode-neovim.useWSL": {
                     "type": "boolean",
                     "default": false,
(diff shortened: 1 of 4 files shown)


theol0403 (approved):
This is a good comprimize for now. I believe a full proper solution *is* possible, but requires more work.

Thanks!

Edit: while reviewing this I decided to dig into the reason why it wasn't working and I believe I fixed it. I'll make a new PR!

trkoch:
@theol0403 Thank you for going the extra mile!

PR #1049, opened 2022-10-12
fact(main_manager): allow circular manager dependencies

theol0403:
This reworks dependency management and allows managers to access each other without passing through constructors and allows circular dependency.


justinmk:
Does this prevent webpack/esbuild minifiers from working?

theol0403:
> Does this prevent webpack/esbuild minifiers from working?

No idea but I'd assume not? It's just how the managers refer to each other.

Issue #1097, opened 2022-12-12
Search highlight is broken on lines starting with tab (\t)


guyguy2001:
**Reproduction**
Paste the following code into a file:
```cpp
						Variant::construct(base.type.builtin_type, base_val, nullptr, 0, err);
```
(The first 6 characters are tabs, you might need to use convert indentation to tabs)
and search for `base` via `/`.
![image](https://user-images.githubusercontent.com/6932278/207110500-c706f9f4-6c9e-49d5-bedd-22a781dffd86.png)
(The image is after also performing a visual select)
There are 2 problems:
* The highlights are too much to the right
* The correct highlight (on the first `base`) was created when I did a visual select, and now it's still there even after I exited it

I started debugging this, so I wanna continue working on this until I yield or finish.
One question that I have - is there any documentation on how either vscode-neovim or neovim itself handles columns? I see a lot of stuff like `useBytes`, and it feels to me like I need to understand it to work with the relevant code, even if I'm not using multibyte characters in my code

guyguy2001:
Oh it's a regression caused by 9a06918 - this explains why I haven't found any issue about it.
I'll try to see why it wasn't obvious to the author that they were creating a bug, and make that code clearer

guyguy2001:
I found this piece of code:
https://github.com/vscode-neovim/vscode-neovim/blob/9d1288eea92c547d2b6259208164b81e325d5889/src/viewport_manager.ts#L42-L52
What does `isByteCol` mean? Is it "are we counting the columns as bytes"? If so, what is the alternative? (because `isByteCol` is used for `calculateEditorColFromVimScreenCol`, which also accepts `false` there in other places.

@theol0403 @zbw8388 - I saw that the two of you co-authored this code (according to git at least), could you give me input on this?
(Sorry if I shouldn't have pinged here - I figured that this situation is what it's for, and I couldn't find any info on when it's okay/not okay to ping)

theol0403:
Sorry don't have time to address this rn. I should be able to in a week. 
A lot of this code is written by the original author of the ext.

PR #1099, opened 2022-12-13
fix(highlight): limit special emoji handing to where it's relevant (#1097)

guyguy2001:


src/utils.ts
```diff
@@ -256,15 +256,21 @@ export function calculateEditorColFromVimScreenCol(
...
+        if (line[currentCharIdx] === "\t") {
+            currentVimCol += tabSize - (currentVimCol % tabSize);
+            currentCharIdx++;
+        }
+        else if (useBytes) {
+            const bytes = getBytesFromCodePoint(line.codePointAt(currentCharIdx));
+            currentVimCol += bytes;
+            // Characters which take 4 bytes also take 2 string indices.
+            // (Only relevant here since wcwidth returns half of the value to each part)
+            currentCharIdx += bytes === 4 ? 2 : 1;
```
guyguy2001, line 268: I understand why this check is relevant only here, and therefore feel comfortable with limiting it here, but I still don't understand this function - why do we want to handle tabs differently? they count as just one column when using `col`.
Also, I don't get _why_ wcwidth is used, but I do understand that it doesn't suffer from the problem emoji had

src/utils.ts
```diff
@@ -256,15 +256,21 @@ export function calculateEditorColFromVimScreenCol(
     let currentCharIdx = 0;
```
guyguy2001, line 256: Can't we just use `charidx`, either as an async call or copying from the sources?

theol0403:
Off the top of my head, tabs may be treated as emojis because https://github.com/vscode-neovim/vscode-neovim/blob/48900ba60cb9da2f6da50da782ee9bc9e01a0dca/vim/vscode-options.vim#L34

There was a reason for the author to write it but I currently forget.

theol0403:
Can't help right now, middle of exam season. However, feel free to think critically of the code and if you think you can simplify workarounds/reduce bloated code, that would be appreciated. A lot of these workarounds can either be made more clean, or exist for legacy reasons.

guyguy2001:
These are 2-byte emoji though (`'❥'.length === 1`), and from my understanding they aren't used normally in this calculatio

guyguy2001:
I am trying to look at how to improve the relating code, and might make a PR for that soon.
However, this PR solves the bug, and doesn't _seem_ to cause any new ones, from my minimal testing, so I'll upload tests for it (forgot to do that lol) and will appreciate a merge (unless you want me to sanity it for longer, e.g. using it in a real codebase)

guyguy2001:
I worked with it for just a bit, but it seems stable

guyguy2001:
How can I fix all of the linting messages? `npm run lint --fix` doesn't pass the `fix` to the program

theol0403:
Try `npm run lint && npm run format`

theol0403 (approved):

theol0403:
Thank you for the PR! I'm not really able to help too much with your questions, because I didn't write this part of the code, but feel free to keep tweaking things and see if you can reach a simpler implementation.

PR #1100, opened 2022-12-15
fix(highlight): check overlay from config

windwp:
@theol0403  sorry I just realize It is not only search and incsearch group.
 highlight group of that config can make the ext_mark overlay broken
https://github.com/vscode-neovim/vscode-neovim/blob/master/package.json#L234


theol0403 (approved):

trkoch:
@windwp So the purpose of this change is to special-case Directory, Visual, Conceal, Substitute as well, not just Search? Am I following correctly? Otherwise overlays are broken for those highlights? Note the config may be expanded by the user, so depending on the user settings more than those highlights may be special-cased.

PR #1110, opened 2022-12-22
Improve callAtomic error logging

pranavrajpal:
I spent a lot of time trying to figure out why this extension and Neovim's view of the current file always seemed to ignore any changes made in insert mode, only to realize that it was caused by my installed Neovim being too old to have `nvim_buf_set_text` defined (which was necessary after #992).

This PR makes a few changes to make errors like that easier to discover by:
- Correcting the minimum Neovim version check to check for the actual minimum version of 0.8.0
- Changing `callAtomic`'s error handling to actually report all errors to the user (I suspect the old implementation wasn't actually reporting any errors since it seems like it was looking for errors in the wrong place)
- Running CI on the oldest supported Neovim version to avoid accidentally depending on newer features

The extra error reporting in `callAtomic` seems to be surfacing some old errors (so far I've found errors in both `nvim_win_close` and `nvim_buf_set_name`) and I'm not sure whether we should silence those for now or leave them on until they're fixed. For now, I've silenced `nvim_win_close` errors but I'm not sure whether that's the right decision.


pranavrajpal:
I don't really understand why the CI is failing. The "extmark display overlay" test is causing the error "nvim_call_function: Vim:E5555: API call: col value outside range", but I can't reproduce that error on my computer.

theol0403:
Thank you for the PR! The error logging changes need a bit more attention (to make sure it won't cause problems, and maybe fix the underlying problems), so I have gone ahead and cherry-picked the simpler changes in https://github.com/vscode-neovim/vscode-neovim/pull/1115. Feel free to keep working on this PR!

I'm not sure if the failing test is your fault.

theol0403 (approved):
See if the window close error can be avoided

xiyaowong:
In most cases, we don't care about the results of `nvim_win_close`, just keep the log for the maintainers to see.

xiyaowong:
I just discovered that there are so many descriptions. In fact, these are useless and should be deleted.
image

theol0403:
It's true that most of them aren't relevant, but I don't think it really matters

PR #1119, opened 2023-01-02
typo? use k instead of i to moveEditorToAboveGroup

ctstacey:


justinmk (approved):

theol0403:
Not a typo. Mentioned in readme. C-k clashes with many vscode bindings.

ctstacey:
Hi @theol0403,
Thanks for pointing it out in the readme, I missed that :)

I wonder though...
It looks like <kbd>Ctrl</kbd>+<kbd>k</kbd> is already being sent to neovim when not in insert mode (and the user has set `neovim.ctrlKeysNormal` causing ctrl key combos to be sent to neovim when not in insert mode).
https://github.com/vscode-neovim/vscode-neovim/blob/30f268b5c22643d4d6fe9d52b25f1e275ad5720b/package.json#L791-L796

`<C-w><C-k>` mapping would only be used in normal and visual mode (`n` and `x`)
https://github.com/vscode-neovim/vscode-neovim/blob/30f268b5c22643d4d6fe9d52b25f1e275ad5720b/vim/vscode-window-commands.vim#L111-L112

If user has not set `neovim.ctrlKeysNormal`, <kbd>Ctrl</kbd>+<kbd>w</kbd> wouldn't be sent in either normal or visual modes (`n` or `x`) regardless.

If you're willing to re-consider, please re-open this PR and I'll update the readme as well.
Sincerely,
Chris

ctstacey:
Ready for review/suggestions, thanks :)

ctstacey:
Was the version already bumped?
Would this actually be released in 0.0.96?

theol0403:
This would be added to 0.0.97.

theol0403:
Thank you for all your PRs!

PR #1127, opened 2023-01-08
docs(readme): fix visual studio marketplace badge

jonz94:
It seems that the https://vsmarketplacebadge.apphb.com is down for a while, so I replace the marketplace badge with the one generated by https://shields.io.

Extra color `#4c1` and label `Visual Studio Marketplace` settings were added to make it looks like the original badge.

vscode-neovim-bagde
Screenshot source: https://web.archive.org/web/20220413024110/https://github.com/vscode-neovim/vscode-neovim

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 <p align="center"><strong>VSCode Neovim Integration</strong></p>
 
 <p align=center>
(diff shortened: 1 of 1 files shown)


justinmk (approved):
thanks! shields.io has been reliable for many years.

Issue #576, opened 2021-03-13
visual mode: gj/gk not working in wrapped lines


Perry3D:
This is based on issue #90 but for visual mode instead of normal mode. 
Currently it is not possible to jump over wrapped lines with gj/gk when visual mode is on.

I tried to copy these lines: 
https://github.com/asvetliakov/vscode-neovim/blob/6469c216f30c4bf29d5718f6d3fe99d517a33db0/vim/vscode-motion.vim#L15-L16

and changed them to vnoremap:
```vim
vnoremap gk <Cmd>call VSCodeNotify('cursorMove', { 'to': 'up', 'by': 'wrappedLine', 'value': v:count ? v:count : 1 })<CR> 
vnoremap gj <Cmd>call VSCodeNotify('cursorMove', { 'to': 'down', 'by': 'wrappedLine', 'value': v:count ? v:count : 1 })<CR> 
```

Afterwards the movement works but it doesn't update the selection. Here the cursor is in line 17 but the selection is still stuck in line 18.
![image](https://user-images.githubusercontent.com/986287/111029965-02656d00-8400-11eb-854a-15260ac129d5.png)

I checked the [API](https://code.visualstudio.com/api/references/commands) and there is the `select` argument for cursorMove.
```vim
vnoremap gk <Cmd>call VSCodeNotify('cursorMove', { 'to': 'up', 'by': 'wrappedLine', 'value': v:count ? v:count : 1, 'select': true})<CR>
vnoremap gj <Cmd>call VSCodeNotify('cursorMove', { 'to': 'down', 'by': 'wrappedLine', 'value': v:count ? v:count : 1, 'select': true})<CR
```
But i get an error that the variable `true` is undefined: `emsg: E121: Undefined variable: true`

Any ideas how i can use the `select` argument?

guyguy2001:
From https://devhints.io/vimscript:
> "No booleans. 0 is false, 1 is true."
![image](https://user-images.githubusercontent.com/6932278/111707895-92b31000-884d-11eb-8bb2-bea85d3cbe61.png)
I tried to google around to see if the neovim package for node.js (which is the thing vscode-neovim uses) supports some magic for sending booleans, but I couldn't find anything yet. I'll look further if sending 0 and 1 don't work for you.

theol0403:
Related: #580

albert-ying:
Any update on this? Thank you!

theol0403:
[...]

PR #1148, opened 2023-02-04
fix(map): move wrapped lines in visual mode with gk/gj

ctstacey:
Add mapping to move up/down virtual lines (i.e. wrapped lines), in visual modes.

Complements the current mapping for moving up/down virtual lines in normal mode.


theol0403:
Does this still encounter the issues present in https://github.com/vscode-neovim/vscode-neovim/issues/576?

theol0403:
This might work now with the new visual selection PR, will give it a try.

theol0403:
Actually, now it definitely does not work, because vscode does not know where the actual (fake) cursor is.

xiyaowong:
From the current perspective, implementing this feature requires continuously saving the cursor positions in both the editor and Neovim, and then performing a series of checks during the bidirectional synchronization process. However, moving by `wrappedLine` in visual mode is a trivial functionality.

PR #1150, opened 2023-02-05
update substr (deprecated) to substring

ctstacey:
`substr` is deprecated (was used for compatibility with Internet Explorer).

Replaces the two occurrences of `substr` with `substring`

### DIFF
--- a/src/highlight_manager.ts
+++ b/src/highlight_manager.ts
@@ -196,7 +196,7 @@ export class HighlightManager implements Disposable, NeovimRedrawProcessable, Ne
                         // use a single text decoration but modify it when a
                         // new decoration would be pushed on top of it.
                         const ogText = drawnAt.get(mapKey).renderOptions.after.contentText;
-                        drawnAt.get(mapKey).renderOptions.after.contentText = (text[0] + ogText).substr(
+                        drawnAt.get(mapKey).renderOptions.after.contentText = (text[0] + ogText).substring(
                             0,
                             ogText.length,
                         );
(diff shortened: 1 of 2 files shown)


ctstacey:
Is it possible the test is flaky? (three `wait`s) Could you please re-run the tests?

image

justinmk:
reran

ctstacey:
> reran

Cheers

justinmk (dismissed):

src/highlight_manager.ts
```diff
@@ -196,7 +196,7 @@ export class HighlightManager implements Disposable, NeovimRedrawProcessable, Ne
                         // use a single text decoration but modify it when a
                         // new decoration would be pushed on top of it.
                         const ogText = drawnAt.get(mapKey).renderOptions.after.contentText;
-                        drawnAt.get(mapKey).renderOptions.after.contentText = (text[0] + ogText).substr(
+                        drawnAt.get(mapKey).renderOptions.after.contentText = (text[0] + ogText).substring(
                             0,
                             ogText.length,
```
justinmk, line 201: The 2nd param of substring() is "end" , not "length". Are you sure this is correct?

ctstacey: In this case, yes, they act the same because the start index is zero.

But you're right, in the case where the start index is non-zero, they act differently.

image

justinmk: should we drop the `ogText.length` argument here?

theol0403 (approved):

PR #1180, opened 2023-03-18
fix: VSCodeNotifyVisual and VSCodeCallVisual produce incorrect VSCode selections in visual and visual block modes if the cursor is placed at the beginning of the visual selection

archilkarchava:
Fixes VSCodeNotifyVisual and VSCodeCallVisual producing incorrect VSCode selections in visual and visual block modes when the cursor is placed at the beginning of the visual selection.

`vim.keymap.set("x", "<D-d>, "<Cmd>call VSCodeNotifyVisual('editor.action.addSelectionToNextFindMatch', 1)<CR><Esc>i", opts)` mapping was used in the examples below.
But the problem exists for all mappings that call either `VSCodeNotifyVisual` or `VSCodeCallVisual`.

#### Before
https://user-images.githubusercontent.com/16840190/226107500-f0731da4-2b6e-4236-b273-251ed4c10b78.mp4

#### After
https://user-images.githubusercontent.com/16840190/226107801-37d475cb-1661-4933-97c3-c93cccf54503.mp4

### DIFF
--- a/vim/vscode-neovim.vim
+++ b/vim/vscode-neovim.vim
@@ -58,7 +58,13 @@ function! VSCodeCallVisual(cmd, leaveSelection, ...) abort
     elseif mode ==# 'v' || mode ==# "\<C-v>"
         let startPos = getpos('v')
         let endPos = getpos('.')
-        call VSCodeCallRangePos(a:cmd, startPos[1], endPos[1], startPos[2], endPos[2] + 1, a:leaveSelection, a:000)
+        "  Check if the cursor is at the beginning of the visual selection
+        if (startPos[1] == endPos[1] && startPos[2] > endPos[2]) || startPos[1] > endPos[1]
+            let startPos[2] = startPos[2] + 1
+        else
+            let endPos[2] = endPos[2] + 1
+        endif
+        call VSCodeCallRangePos(a:cmd, startPos[1], endPos[1], startPos[2], endPos[2], a:leaveSelection, a:000)
     else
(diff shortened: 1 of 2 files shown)


justinmk (approved):
Nice, thanks.

vim/vscode-neovim.vim
```diff
@@ -58,7 +58,13 @@ function! VSCodeCallVisual(cmd, leaveSelection, ...) abort
     elseif mode ==# 'v' || mode ==# "\<C-v>"
         let startPos = getpos('v')
         let endPos = getpos('.')
-        call VSCodeCallRangePos(a:cmd, startPos[1], endPos[1], startPos[2], endPos[2] + 1, a:leaveSelection, a:000)
+        "  Check if the cursor is at the beginning of the visual selection
+        if (startPos[1] == endPos[1] && startPos[2] > endPos[2]) || startPos[1] > endPos[1]
+            let startPos[2] = startPos[2] + 1
+        else
+            let endPos[2] = endPos[2] + 1
+        endif
```
justinmk, line 66: since this is used in multiple places, it's worth extracting it to a function like `fixVisualPos()`. That gives a hint to future readers that happen to look at this function. Also helps if a bug fix is needed. 

Is it possible to add a test?

archilkarchava: @justinmk, I've extracted duplicated logic into the `s:fixVisualPos` function and added a test.

justinmk (approved):

PR #1186, opened 2023-03-24
feat: add $NVIM_APPNAME option

nullchilly:
https://github.com/neovim/neovim/pull/22128

Personally I want a separate neovim config just for vscode, which can be achieved by setting `$NVIM_APPNAME`

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,11 @@
                     "default": false,
                     "markdownDescription": "Equivalent to the startup option `--clean`."
                 },
+                "vscode-neovim.NVIM_APPNAME": {
+                    "type": "string",
+                    "default": "",
+                    "markdownDescription": "See `:h NVIM_APPNAME`."
+                },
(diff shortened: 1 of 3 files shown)


justinmk (approved):
Things like https://github.com/direnv/direnv-vscode are a generalized way to set the vscode environment. 

But this PR probably probably is still needed so that the setting can be vscode-specific. 

I wish vscode would just make it possible to _set_ env vars in settings.json. (You can read them via `${env:FOO}` but I don't know if you can set them?

nullchilly:
Now that [neovim 0.9](https://github.com/neovim/neovim/releases/tag/v0.9.0) was released, can this be merged?

nullchilly:
> Does this PR break the extension when neovim is < 0.9? If so we should at least bump the minimal required version and probably wait a while until users have had the chance to update.

No, works fine on 0.8.3

PR #1225, opened 2023-06-04
docs(readme): clean up the instruction using vim-plug

ghiden:
I myself first encountered `Cond` function not found error.
So, it might be easier for having one more way to enable easymotion without defining `Cond` function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,21 +99,20 @@ else
 end
 ```
 
-To conditionally activate plugins, `vim-plug` has a
-[few solutions](https://github.com/junegunn/vim-plug/wiki/tips#conditional-activation). For example, using the `Cond`
-helper, you can conditionally activate installed plugins
-([source](https://github.com/asvetliakov/vscode-neovim/issues/415#issuecomment-715533865)):
-
+Here is how to enable easy-motion using `vim-plug`:
 ```vim
 " inside plug#begin:
-" use normal easymotion when in VIM mode
-Plug 'easymotion/vim-easymotion', Cond(!exists('g:vscode'))
-" use VSCode easymotion when in VSCode mode
-Plug 'asvetliakov/vim-easymotion', Cond(exists('g:vscode'), { 'as': 'vsc-easymotion' })
+if exists('g:vscode')
+  " use VSCode easymotion when in VSCode mode
+  Plug 'asvetliakov/vim-easymotion', { 'as': 'vsc-easymotion' }
+else
+  " use normal easymotion when in VIM mode
+  Plug 'easymotion/vim-easymotion'
+endif
 ```
+For more information please visit `vim-plug` [wiki page](https://github.com/junegunn/vim-plug/wiki/tips#conditional-activation).
 
-See [plugins](https://github.com/vscode-neovim/vscode-neovim/wiki/Plugins) in the wiki for tips on configuring VIM
-plugins.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,6 +112,18 @@ Plug 'easymotion/vim-easymotion', Cond(!exists('g:vscode'))
 Plug 'asvetliakov/vim-easymotion', Cond(exists('g:vscode'), { 'as': 'vsc-easymotion' })
 ```
 
+If you do not want to define `Cond` function, here is how you do it:
```
justinmk, line 115: Not really sure we need to mention `Cond` at all. We should probably remove it from the docs and code entirely. It's noisy to mention multiple ways to do something in the docs, for no reason.

ghiden: Yeah, I agree with you. How about cleaning the whole section like this? I've updated my commit. Please have a look.

README.md
```diff
@@ -99,21 +99,20 @@ else
...
-Plug 'asvetliakov/vim-easymotion', Cond(exists('g:vscode'), { 'as': 'vsc-easymotion' })
+if exists('g:vscode')
+  " use VSCode easymotion when in VSCode mode
+  Plug 'asvetliakov/vim-easymotion', { 'as': 'vsc-easymotion' }
+else
+  " use normal easymotion when in VIM mode
+  Plug 'easymotion/vim-easymotion'
+endif
 ```
+For more information please visit `vim-plug` [wiki page](https://github.com/junegunn/vim-plug/wiki/tips#conditional-activation).
```
justinmk, line 113: There are zillions of plugin managers, we don't need to link to their docs. Users are expected to know how to use things they have chosen to install.

justinmk (approved):

README.md
ghiden: Do you want me to remove this sentence?

justinmk: yeah, otherwise we might get more and more of this, people tend to mimic what's already there. the readme is too verbose already.

theol0403:
I personally believe Cond is very important for a vscode-neovim config, for other plugins as well. I think explicitly pointing to it is good.

justinmk:
> I personally believe Cond is very important for a vscode-neovim config

Ok, but the readme doesn't explain why it's important. I personally have never used it, and it's pretty confusing that we have this bespoke alternative to basic control flow.

theol0403:
> > I personally believe Cond is very important for a vscode-neovim config
> 
> Ok, but the readme doesn't explain why it's important. I personally have never used it, and it's pretty confusing that we have this bespoke alternative to basic control flow.

It's needed to specify which plugins should be enabled in vscode vs terminal neovim. Using if/else doesn't work because you have to run install whenever you switch. I agree it could be explained better. Packer/lazy has `cond=` built in.

ghiden:
> > > I personally believe Cond is very important for a vscode-neovim config
> > 
> > 
> > Ok, but the readme doesn't explain why it's important. I personally have never used it, and it's pretty confusing that we have this bespoke alternative to basic control flow.
> 
> It's needed to specify which plugins should be enabled in vscode vs terminal neovim (if you want to share config). Using if/else doesn't work because you have to run install whenever you switch. I agree it could be explained better. Packer/lazy has `cond=` built in.

Thanks for having this discussion. @theol0403 yes, `Cond` is far superior than `if/else` as I have to install once in each platform.
I think the current description needs to be enhanced for sure. I'm closing this ticket.

PR #1244, opened 2023-06-27
feat(ci): add automatic releases

theol0403:
Using release-please to streamline releases. The vsce key currently does not seem to be working, however. 

Additionally, switching to SemVer versioning.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -2,6 +2,10 @@ name: Code Check & Test
 
 on: [pull_request, push]
 
+permissions:
+    contents: write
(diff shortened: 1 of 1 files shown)


justinmk:
Are there any human-driven release steps / checklist that should be written down somewhere (I typically just put it at the end of the readme in a "Release" section)?

theol0403:
Not anymore, other than ensure all PRs have conventional commits when merged. It should also publish to marketplace, but currently my keys are broken. 

Sure, explaining the process can be helpful. There are no manual steps though, it pushes the tags for you. Might have to renew the secrets every few months.

PR #1246, opened 2023-06-28
refactor(tests): cleanup and remove delays

theol0403:
This should cut CI duration by half. The challenge will be ensuring it has not become more flaky.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -26,9 +26,7 @@ jobs:
               id: vim
               with:
                   neovim: true
-                  # Use the oldest supported Neovim version here so we don't accidentally
-                  # break the extension on older versions
-                  version: v0.8.0
+                  version: v0.9.0
             - name: npm install
               run: npm ci --silent
             - name: Build
--- a/src/test/suite/index.ts
+++ b/src/test/suite/index.ts
@@ -14,8 +14,8 @@ export async function run(): Promise<void> {
         bail: false,
         slow: 20000,
         fullTrace: true,
-        retries: 3,
         grep: ".*",
+        retries: 1,
     });
     const testsRoot = path.resolve(__dirname, "..");
     return new Promise((c, e) => {
--- a/src/test/suite/index.ts
+++ b/src/test/suite/index.ts
@@ -15,7 +15,7 @@ export async function run(): Promise<void> {
         slow: 20000,
         fullTrace: true,
         grep: ".*",
-        retries: 1,
+        retries: 0,
     });
     const testsRoot = path.resolve(__dirname, "..");
     return new Promise((c, e) => {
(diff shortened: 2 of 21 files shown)


theol0403:
Looks like I broke ~6 tests, but the duration has gone from 20min to 5min!

.github/workflows/build_test.yml
```diff
@@ -26,9 +26,7 @@ jobs:
               id: vim
               with:
                   neovim: true
-                  # Use the oldest supported Neovim version here so we don't accidentally
-                  # break the extension on older versions
-                  version: v0.8.0
+                  version: v0.9.0
```
justinmk, line 29: somewhat off topic: does vscode-neovim actively check the nvim version and show a user message if it doesn't meet the requirement?

src/main_controller.ts
```diff
@@ -124,6 +124,7 @@ export class MainController implements vscode.Disposable {
             args.push(
                 "-u",
                 "NONE",
+                "--clean",
```
justinmk, line 127: could remove `-u NONE` above

justinmk (approved):
Very nice code reducation. To unblock this, could the wait() be added to `openTextDocument` and then revisit the flaky failures in a separate PR? So at least we get the cleanup, though the tests might still be slow.

.github/workflows/build_test.yml
theol0403: Yes, it currently checks for v0.8. I bumped the CI version because some tests failed on my machine (running 0.9) due to some minor changes in nvim (specifically :help), and I wanted to update the test to futureproof.

src/test/suite/index.ts
```diff
@@ -14,8 +14,8 @@ export async function run(): Promise<void> {
         bail: false,
         slow: 20000,
         fullTrace: true,
-        retries: 3,
         grep: ".*",
+        retries: 0,
```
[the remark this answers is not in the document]
theol0403: This is tricky. I added retries last year because I wanted green CI, but the problem is that many bugs relating to this project are timing related and retrying with CI could hide some reliability regressions. A test that fails 2/3 times is still a failing test. For now, I removed the retries just so I could really examine which tests were flaky and if I could fix them, but undecided if I will keep them.

theol0403:
> To unblock this, could the wait() be added to openTextDocument and then revisit the flaky failures in a separate PR? So at least we get the cleanup, though the tests might still be slow.

Most of the flaky failures are caused from me removing waits throughout the tests, + some consistently failing tests that I've broken. I think it's too coupled to bother separating.

theol0403:
If anyone is subscribed... sorry for all the noise. Fixing these tests are like wack-a-mole. Took me 1 day to get 95% there... and 3 days to get the last 5%,

theol0403:
Ugh, I don't know what I missed, but for some reason the tests have a catastrophic failure and fail a whole bunch of tests ever since the visual selection PR. I'll try to figure out the issue and merge this at the same time before making a release.

PR #1248, opened 2023-07-01
feat: global window navigations

martin-braun:
Improved window navigation in- and outside of the text editor. Allows to enter and leave the file explorer using window navigations and so much more.


martin-braun:
> @theol0403
> 
> > I don't entirely see the problem with binding `navigateDown` and it leading you to the terminal... is that not a sane and general default? Essentially, it treats the terminal kind of like another buffer.
> 
> It depends on the user I guess. I think there are vim plugins that allow to have a terminal as a buffer, but I think a lot of people just use `tmux` for that. In that case, you cannot enter the terminal with window movements, you have to use `tmux` bindings to switch to the terminal.
> 
> Personally, I wouldn't want to map window movements in a way that it allows to enter the terminal for this particular reason, but I can't speak for the general public in that regard.

martin-braun:
The question ultimately boils down to: Should we treat the terminal as a buffer or not. Personally I'd say no, but I care about the general consensus, since I will remap things for my needs anyways.

theol0403:
Yeah, not sure if I know to make the call either.

justinmk (approved):
> It depends on the user I guess. I think there are vim plugins that allow to have a terminal as a buffer, but I think a lot of people just use tmux for that. In that case, you cannot enter the terminal with window movements, you have to use tmux bindings to switch to the terminal.

Terminals are certainly just buffers in Nvim. But for these mappings, we don't need to be pedantic, just do whatever is most useful in the context of vscode.

It's a reversible decision, we can try these out and improve later.

martin-braun:
Let's decide in favor of treating terminals like buffers. This is the most reasonable thing to do, because terminals are part of the frame. The mentality is closer to an embedded buffer that function as terminal, especially since the file explorer goes past the terminal.

This becomes especially dominant when the user decides to put the terminal on the right side. The previous keymaps wouldn't allow to enter the terminal via window navigation mappings no matter what.

Anybody who want's a tmux-alike behavior can remap things properly.

martin-braun:
To anybody who wants to keep the terminal below the editor and wishes it to be excluded for window navigation (to uplift the `tmux` mentality) may change their bindings like so:

```json
    {
        "key": "ctrl+w up",
        "command": "-workbench.action.navigateUp",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w k",
        "command": "-workbench.action.navigateUp",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w down",
        "command": "-workbench.action.navigateDown",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w j",
        "command": "-workbench.action.navigateDown",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w up",
        "command": "workbench.action.focusAboveGroup",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w k",
        "command": "workbench.action.focusAboveGroup",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w down",
        "command": "workbench.action.focusBelowGroup",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w j",
        "command": "workbench.action.focusBelowGroup",
        "when": "!editorTextFocus"
    },
```

Thanks.

martin-braun:
Also: These window navigation bindings exclude panel navigation, i.e. in the Terminal. I mapped those in a way that reminds me of my `tmux` navigation bindings, I have no clue what reasonable defaults would be appropriate, but this is a different issue.

package.json
```diff
@@ -1269,6 +1269,65 @@
                 "command": "vscode-neovim.complete-selection-cmdline",
                 "key": "tab",
                 "when": "neovim.mode == cmdline_normal || neovim.mode == cmdline_insert || neovim.mode == cmdline_replace"
+            },
+            {
+                "key": "ctrl+w",
+                "command": "-workbench.action.switchWindow"
```
theol0403, line 1275: What is this for? I don't think this is a vscode default

theol0403 (approved):
Thanks for the PR! I think people have been asking for this for a while so it is good to have it as a default.

I think it would be worth it to add as many bindings that reflect the existing ones in `vim/vscode-window-commands.vim` as possible. For example, C-w+n/c/C-c/o/C-o/C-j/C-down/S-j/S-down. You could theoretically add bindings for everything, including splits/C-w+=.

My reasoning is that it would be nice to be able to have full access to C-w bindings regardless if a nvim-owned editor is focused or not.

justinmk:
This is definitely something I've missed (and too lazy to figure out myself), so thank you @martin-braun

package.json
martin-braun: Screenshot 2023-07-03 at 12 28 46

It definitely is and it intercepts.

theol0403: Strange, it isn't for me. Nothing is bound to switchWindow. Maybe a macos thing?

image

martin-braun:
I'm glad that you guys like it, but I can't take care of the entirety of `vim/vscode-window-commands.vim` at the moment. I consider these essential enough to be a must-have. Things are straight-forward and more organized if split into smaller PR imho.

This PR is about window navigation exclusively and I'd like to keep it this way, so it has a better chance to land without any conflicting issues. 

BUT: Resist the merge yet, there are more activity / side panels that the user needs the ability to escape from using `ctrl+w ctrl+w`, but we will get there.

package.json
martin-braun: @theol0403 Looks like it. [This gentleman](https://stackoverflow.com/a/52289040/1540350) also uses MacOS I think.

martin-braun:
@justinmk @theol0403 You can pretty much merge this. There are no other when-expressions I'm aware of that would help to improve the window navigation.

Issue #116, opened 2019-12-21
Weird behaviour after pasting a yanked line when selecting "vi{"


OfirTadmor:
Seems to add an extra "}" when pasting, 

![image](https://user-images.githubusercontent.com/13004405/71309616-ad13f900-2412-11ea-8e85-b19d67e65138.png)

![image](https://user-images.githubusercontent.com/13004405/71309608-99689280-2412-11ea-83e3-eb577a20d3f3.png)

asvetliakov:
This is serious and indicates that buffer in neovim side has messed up.  Do you remember steps you've done *before* pasting?

OfirTadmor:
this only happens when yanking a link  "yy", any other type of yank seems to work.
I was afraid it was something to do with my extensions in vscode or auto complete settings inside vscode, but after trying to disable most of them this problem stayed.
were you able to recreate this ?

asvetliakov:
Just reproduced, looking

asvetliakov:
Looks like upstream issue https://github.com/neovim/neovim/issues/11591

xiyaowong:
This issue has been fixed in nvim v0.10.0 (not yet released)

PR #1250, opened 2023-07-01
feat: collapse all and copy bindings for file explorer

martin-braun:
nvim-tree allows to copy files using `c` as well. Also, you can collapse all folders with `W` which is also a default on nvim-tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -651,6 +651,16 @@
                 "when": "filesExplorerFocus && !inputFocus"
             },
             {
+                "key": "c",
+                "command": "filesExplorer.copy",
+                "when": "explorerViewletVisible && filesExplorerFocus && !explorerResourceIsRoot && !inputFocus"
+            },
+            {
+                "key": "shift+w",
+                "command": "workbench.files.action.collapseExplorerFolders",
+                "when": "filesExplorerFocus"
+            },
+            {
                 "key": "tab",
                 "command": "togglePeekWidgetFocus",
                 "when": "inReferenceSearchEditor && neovim.mode == normal || referenceSearchVisible"
--- a/package.json
+++ b/package.json
@@ -651,12 +651,7 @@
                 "when": "filesExplorerFocus && !inputFocus"
             },
             {
-                "key": "c",
-                "command": "filesExplorer.copy",
-                "when": "explorerViewletVisible && filesExplorerFocus && !explorerResourceIsRoot && !inputFocus"
-            },
-            {
-                "key": "shift+w",
+                "key": "z shift+m",
                 "command": "workbench.files.action.collapseExplorerFolders",
                 "when": "filesExplorerFocus"
             },


martin-braun:
Unrelated `pasting line after vi{` test for #116 blocks my contribution. @asvetliakov

theol0403:
Don't worry about the tests, they are flaky.

martin-braun:
@theol0403 How to re-trigger them?

theol0403:
I can merge on a failing test. Just ignore the results.

package.json
```diff
@@ -650,6 +650,16 @@
                 "command": "explorer.newFolder",
                 "when": "filesExplorerFocus && !inputFocus"
             },
+            {
+                "key": "c",
+                "command": "filesExplorer.copy",
+                "when": "explorerViewletVisible && filesExplorerFocus && !explorerResourceIsRoot && !inputFocus"
```
justinmk, line 656: would `yy` make sense for this? why are we using nvim-tree as the source of conventions instead of more vim-like conventions?

package.json
```diff
@@ -650,6 +650,16 @@
...
+            },
+            {
+                "key": "shift+w",
+                "command": "workbench.files.action.collapseExplorerFolders",
+                "when": "filesExplorerFocus"
```
justinmk, line 661: `zM` would be more appropriate here. Or `zc` if this is cursor-relative.

package.json
martin-braun: It's the most popular file explorer for Neovim, it is not? We have a `y` mapping for copy already. `nvim-tree` allows to use either `y` or `c`, I'm used to `c` on the file explorer, so others might, too.

package.json
martin-braun: I wasn't aware of Neovim's fold mechanism. I agree about `zM` after reading it's documentation. For some reason `nvim-tree` uses `W`. Interested in supporting both maps? I could add `zM` to my PR.

package.json
justinmk: > It's the most popular file explorer for Neovim, it is not?

I have no idea, I've never used it. And before the NERDTree was the most popular for 10+ years, but that makes no difference.  Popular plugins come and go and `c` is the vim convention for "change", the complete opposite of yank.

package.json
justinmk: `W` seems random. Bad defaults chosen by plugins shouldn't leak into upstream projects.

martin-braun:
@justinmk I agree. Let's not waste more time here. Anybody is free to adjust maps, I just added `zM` for collapse, since this is reasonable.

justinmk (approved):

martin-braun:
@theol0403 I think it would be reasonable to exclude "flaky" tests if they randomly fail so many times, don't you think?

theol0403:
I am trying to fix the tests but it is proving to be very frustrating - especially since some of the failures come from genuine timing bugs in the extension. For now, it is what it is. It's not the same flaky tests every time.

theol0403:
Just so we don't only have this one "z" binding that sticks out - is there any other z<key> binding we could use that would make sense here? Even something like zz, zt for scrolling and any other fold+- bindings. I ask just because it feels awkward to introduce the z leader key to the possible tree bindings, but for there to only be one child.

theol0403:
Thanks for the PR btw! When I made the current bindings I picked something that seemed logical, but drawing inspiration from neotree or nvim-tree is in my opinion ideal.

martin-braun:
@theol0403 

> but drawing inspiration from neotree or nvim-tree is in my opinion ideal.

So, you actually support my initial commit of mapping `c` and `W`? I'm torn apart, I don't mind. On one hand it can't be bad to apply defaults from popular tree plugins, on the other hands I have to give @justinmk props for mentioning that plugins come and go. What if we have the ultimate tree plugin tomorrow that uses `c` for a different behavior?

Ultimately, I don't mind really.

> is there any other z binding we could use that would make sense here?

Yes, having only one z-binding is odd, but the result of the circumstances. Never used folding in Neovim, so lets elaborate together. 

Screenshot 2023-07-03 at 05 31 00

Vim is very feature rich when it comes to fold operations, VSCode doesn't offer enough `workbench.file.action` commands to substitute it properly. `z<CR>`/`z+`, `z.`/`zz`, `z-` would be great to have, but I doubt I can scroll the file explorer window without changing its selection. There is `list.focusPageDown` and `list.focusPageUp`, but it will scroll down / up with cursor movement, which is not the same.

@justinmk 

> and `c` is the vim convention for "change", the complete opposite of yank.

I realize this is a loose point, because `c` has a lot of meanings in Vim, btw. `:wincmd c` will close a window and not change it, so a `c` mapping to copy files wouldn't be as controversial as it seems. Just a reminder.

-----

All I wanted was to provide an alternative for copy and offer a collapse all (which is very useful to get clean on the files explorer) to play nice with `nvim-tree` compatibility.

Maybe we just drop it and let the user come up with their own, you decide.

theol0403:
I agree that is probably too minor of a topic to keep debating about. 

> So, you actually support my initial commit of mapping c and W? 

I'm pretty indifferent, but I agree with @justinmk on those specific points. Both c and W seem pretty random. All I was trying to say was "it is good that someone is looking at the different setups and validating the current design". In the same vein, any existing bindings that don't make sense could be deleted. 

I think something like c and W is something the user should add themselves :+1: 

> Never used folding in Neovim, so lets elaborate together.

I think we could add za/zc/zo, using list.toggle/collapse/expand. 

I think all we really need is for this pr to add za/zc/zo/zM, and *potentially* clean up any existing bindings that don't fit in with either vanilla nvim OR nvim-tree. 

Thank you for your efforts on this btw, sorry for going back and forth so much on something trivial.

martin-braun:
> I think we could add za/zc/zo, using list.toggle/collapse/expand.

I will look into it later, kk.

martin-braun:
@theol0403 `zm` vs `zM`? Also I can't replicate `zO`, `zC`, `zA`. Should they just not work or should they fallback to lowercase maps? Lastly, `zR` can't be mapped, because VSCode has no `list.expandAll`.

theol0403:
@martin-braun I suppose bind zm and zM to the same thing. Same with zO/zC/zA. All lowercase fallbacks. And yeah, seems like zR is not possible.

justinmk (approved):
Should have a one-line note about Explorer mappings in the readme.md. It's not super obvious.

theol0403:
I can merge when ready.

martin-braun:
Please review once more. If it's cool: Let's go. :)

README.md
```diff
@@ -341,16 +341,21 @@ These are the default commands and bindings available for file/scroll/window/tab
...
+| <kbd>h</kbd> / <kbd>l</kbd>                           | `list.collapse/select`          |
+| <kbd>Enter</kbd>                                      | `list.select`                   |
+| <kbd>gg</kbd>                                         | `list.focusFirst`               |
+| <kbd>G</kbd>                                          | `list.focusLast`                |
+| <kbd>o</kbd>                                          | `list.toggleExpand`             |
+| <kbd>C-u</kbd> / <kbd>C-d</kbd>                       | `list.focusPageUp/Down`         |
+| <kbd>z</kbd> <kbd>o</kbd> / <kbd>z</kbd> <kbd>O</kbd> | `list.expand`                   |
+| <kbd>z</kbd> <kbd>c</kbd>                             | `list.collapse`                 |
+| <kbd>z</kbd> <kbd>C</kbd>                             | `list.collapseAllToFocus`       |
+| <kbd>z</kbd> <kbd>a</kbd> / <kbd>z</kbd> <kbd>A</kbd> | `list.toggleExpand`             |
```
justinmk, line 356: To help with discovery, 1-2 line mention the readme is worth it. But the high-fidelity listing of every mapping is not worth it. Users should instead check vscode shortcuts UI to see exact mappings. The readme should just give a useful hint like:

> In the vscode File Explorer some familiar vim commands are configured by default:
> - for navigation: `j`, `k`, `gg`, `G`
> - for expanding/collapsing directories: `zo`, `zc`, etc.
> - run the `Preferences: Open Keyboard Shortcuts` vscode command to see the complete list of keybindings.

This is a more maintainable pattern that we can follow in all sections of the readme, which will result in a less noisy, more maintainable, and still useful overview.

martin-braun: @justinmk Would you mind doing this? I don't know which place would be the best to insert these. I will revert my `README.md` changes accordantly later.

justinmk: Ok, not a blocker

martin-braun:
Done!

README.md
theol0403: I do agree, but
1. until this is cleaned up, I'd rather include the list in the readme. I'll revert myself
2. unfortunately discovery in keybindings editor is difficult because there is no way to filter by extension

martin-braun:
Revert-Revert-Inception. You're welcome. :)

PR #1254, opened 2023-07-03
feat(main): add options.lua

theol0403:
This is the start of moving all vimscript to lua.

I was able to simplify the code a lot by reusing the "force options" function for both global and local functions.


hustrjh:
Perfect. I'm new to Neovim and it's a little difficult to write lua. Thanks a lot.

theol0403:
@hustrjh what? this has nothing to do with your config

hustrjh:
Aha, sorry for my misunderstanding. But is it possible to provide lua config example in the vault in the future?

theol0403:
@hustrjh you can use lua for your config just like you can with normal neovim

hustrjh:
Thanks. I get it. But it may be more friendly to users if there's some lua examples. I'm a little confused to call vscode command in lua. Here's example in vimscript below.
`xnoremap <C-S-P> <Cmd>call VSCodeNotifyVisual('workbench.action.showCommands', 1)<CR>`

theol0403:
There is no API for that yet. I hope to add one.

hustrjh:
Thanks for your patient reply and your excellent effort. I appreciate it a lot. 😊

justinmk:
> I'm a little confused to call vscode command in lua. Here's example in vimscript below.
> `xnoremap <C-S-P> <Cmd>call VSCodeNotifyVisual('workbench.action.showCommands', 1)<CR>`

vimscript functions can be called from Lua via `:help vim.fn`, e.g.:

    :lua vim.fn['VSCodeNotifyVisual']('workbench.action.showCommands', 1)

But given this PR, something like `require('vscode').notifyvisual(...)` would be nicer.

hustrjh:
@justinmk  Thanks for your advice. 😁

PR #1255, opened 2023-07-03
feat(mode): switch to `ModeChanged` for more detailed mode information

theol0403:
This gives us information for when the mode switches between kinds of visual modes. 

Based on https://github.com/vscode-neovim/vscode-neovim/pull/1254

### DIFF
--- a/src/mode_manager.ts
+++ b/src/mode_manager.ts
@@ -7,15 +7,15 @@ import { NeovimExtensionRequestProcessable } from "./neovim_events_processable";
 
 const LOG_PREFIX = "ModeManager";
 
-// a representation of the current mode. can be read in different ways using accessors. underlying type is full name as returned by `:help mode()`
+// a representation of the current mode. can be read in different ways using accessors. underlying type is raw name as returned by `:help mode()`
 class Mode {
-    public constructor(public full: string = "") {}
+    public constructor(public raw: string = "") {}
     // mode 1-char code: n, v, V, i, s, ...
     // converts ^v into v
     public get char(): string {
-        return this.full.charCodeAt(0) == 22 ? "v" : this.full.charAt(0);
+        return this.raw.charCodeAt(0) == 22 ? "v" : this.raw.charAt(0);
     }
-    // mode full name
+    // mode raw name
     public get long(): "insert" | "visual" | "normal" {
         switch (this.char.toLowerCase()) {
             case "i":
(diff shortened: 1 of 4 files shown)


theol0403:
My only concern with this is:
- am I parsing the modes (especially visual-block) correctly, or is there a more idiomatic way of doing this
- is there any special vim-mode that we would benefit from parsing correctly
- `mode_change` always fired on buffer init, wheras `ModeChanged` does not. Is `BufAdd` the best autocommand to ensure the mode is set on init? 

This is used for my upcoming visual selection PR, which needs to know the difference between cursor block etc.

src/mode_manager.ts
```diff
@@ -3,17 +3,45 @@ import { EventEmitter } from "events";
...
+                return "insert";
+            case "v":
+                return "visual";
+            case "n":
+            default:
+                return "normal";
+        }
+    }
+    // visual mode name
+    public get visual(): "single" | "line" | "block" {
```
justinmk, line 31: the usual names are charwise, linewise, block. `:help charwise`.

```suggestion
    public get visual(): "char" | "line" | "block" {
```

src/mode_manager.ts
```diff
@@ -3,17 +3,45 @@ import { EventEmitter } from "events";
...
+// a representation of the current mode. can be read in different ways using accessors. underlying type is full name as returned by `:help mode()`
+class Mode {
+    public constructor(public full: string = "") {}
+    // mode 1-char code: n, v, V, i, s, ...
+    // converts ^v into v
+    public get char(): string {
+        return this.full.charCodeAt(0) == 22 ? "v" : this.full.charAt(0);
+    }
+    // mode full name
+    public get long(): "insert" | "visual" | "normal" {
```
justinmk, line 19: how does this differ from `this.full`? could be a confusing interface.

```suggestion
    public get name(): "insert" | "visual" | "normal" {
```

justinmk:
> am I parsing the modes (especially visual-block) correctly, or is there a more idiomatic way of doing this

This is the normal way.

src/mode_manager.ts
theol0403: Thanks for the catch. I renamed this.full to `raw`, which is the shortcode returned by `mode()`.
`long` is the mode name. 

It was confusing indeed, this was why CI is failing lol.

theol0403: On second thought, I prefer your suggestion for `name`.

justinmk (approved):

PR #1257, opened 2023-07-04
feat(keybindings): global window resize/split

martin-braun:
Idea is to have global window split and resize keybindings outside of Neovim. Unfortunately, this doesn't work as it should be.

VSCode keybindings are locale aware. So instead of mapping the character, it requires to map the physical key, which is not what we want. Vim isn't aware of physical keys, so it reacts on characters which will result in different key strokes depending on the keyboard layout.

I'm using a German macOS keyboard layout (ISO whatever), VSCode won't let me map most of the keybindings to a locale-independent keymap: 

Screenshot 2023-07-04 at 08 55 55

Interestingly, the behavior we strive for seems to work for `=`: 

Screenshot 2023-07-04 at 08 57 03

So it seems to be possible, it's just not feature complete(?).


theol0403:
Aside from the localisation issues, is this ready for merge?

martin-braun:
@theol0403 I'd like to know if there is an easy fix for the location issues. I think the worse case is that it will only work partially, depending on your local.

If you decide to merge it, we should open an issue and keep track of it.

theol0403:
What if you tried `ctrl+w shift+.`?

martin-braun:
@theol0403 wdym? `shift+.` starts a command, even when followed by `ctrl+w` in Neovim. Why mapping this one? There are several bindings not working in my locale, not just one.

theol0403:
Sorry, I know nothing about your locale, but I thought maybe binding the key sequence would fix the problem. That assumes `shift+.` = `>` for you, however.

martin-braun:
@theol0403 This is my layout: 

German macOS keyboard layout

To make it work properly, I have to do: 

```json
    {
        "key": "ctrl+w =",
        "command": "workbench.action.evenEditorWidths",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w shift+-",
        "command": "workbench.action.toggleEditorWidths",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w [Backquote]",
        "command": "workbench.action.decreaseViewWidth",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w shift+[Backquote]",
        "command": "workbench.action.increaseViewWidth",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w +",
        "command": "workbench.action.increaseViewHeight",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w -",
        "command": "workbench.action.decreaseViewHeight",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w s",
        "command": "workbench.action.splitEditorDown",
        "when": "!editorTextFocus"
    },
    {
        "key": "ctrl+w v",
        "command": "workbench.action.splitEditorRight",
        "when": "!editorTextFocus"
    },
```

Hopefully you can see where this is going. VSCode wants physical keys, so that no matter which locale you have, the key to press is always at the same position. However, Vim wants logical keys, so the key position you have to press depends on your locale.

VSCode needs to allow to map a logical key. Unless they offer a flag or something like that, we cannot map Vim commands properly. Interestingly, this behavior doesn't happen for `z` or `y` which are swapped on the German keyboard.

xiyaowong:
LGTM

xiyaowong:
Looks good! Very convenient shortcuts

theol0403:
@xiyaowong can you rebase?

xiyaowong:
Wait a minute

xiyaowong:
Maybe change the script name `navigation `to `view`?

theol0403:
Sure, or something related to vscode bindings like vscode-integration.

PR #1268, opened 2023-07-05
build(deps): add dependabot

theol0403:
This makes it easier to update deps.

Should I group all deps into one PR for less noise?


justinmk (approved):

justinmk:
> Should I group all deps into one PR for less noise?

That risks not being able to "bisect" in case we find a regression. 

The "noise" will taper off after an initial deluge.

PR #1300, opened 2023-07-07
fix: colliding ctrl+w bindings in terminal

### LINKED ISSUE #1297: bug: neovim extension blocks ctrl+w binding for kill-word in terminal

### Neovim version (nvim -v)

v0.9.1

### Operating system/version

Windows

### Describe the bug

neovim vscode extension adds various "ctrl+w" bindings that have a "!editorTextFocus" when clause that cause "ctrl+w" in "terminalFocus" mode to match and block the standard bash "kill-word" binding for deleting words with "ctrl+w"

Attempting to add the clause doesn't seem to work for me.  Only disabling the extension makes a difference.

### Steps To Reproduce

1. Install Neovim VSCode Extension
2. Open a terminal in vscode
3. Type something (e.g. "abcd efgh")
4. Attempt to use "Ctrl+w" to delete the second word.
5. Notice a "Waiting for 1 second for additional chord key" message in the bottom status bar.

### Expected Behavior

Ctrl+w should be passed through to the terminal.

martin-braun:
Fixes #1297

Primary focus is to allow using `ctrl+w` to delete the last word. <s>Will introduce a new issue for now: We cannot escape the terminal (for now).</s>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1323,57 +1323,57 @@
             {
                 "key": "ctrl+w ctrl+w",
                 "command": "workbench.action.focusNextGroup",
-                "when": "!editorTextFocus && !(filesExplorerFocus || inSearchEditor || searchViewletFocus || replaceInputBoxFocus)"
+                "when": "!editorTextFocus && !terminalFocus && !(filesExplorerFocus || inSearchEditor || searchViewletFocus || replaceInputBoxFocus)"
             },
             {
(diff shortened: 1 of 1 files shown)


martin-braun:
@theol0403 I like the idea of using `<C-ESC>` to exit the terminal focus. I think this can be merged. In regards of the pane prefix discussion, that should be held at #1252 at a later stage.

PR #1305, opened 2023-07-09
fix(mode): allow escaping from replace mode

### LINKED ISSUE #1304: bug: cannot exit replace mode in Jupyter notebook

### Neovim version (nvim -v)

0.9.1

### Operating system/version

macOS 13.4.1

### Describe the bug

Entering replace mode in a Jupyter notebook cell and then pressing `<Esc>` doesn't go back to normal mode.

Following this, it's impossible to return to normal mode when re-entering the cell, until VSCode is restarted.

This bug doesn't happen in normal text files, only notebooks as far as I can tell.

Disabling VSCode's <Esc> keyboard shortcut for exiting the cell doesn't fix the bug (the cell is no longer exited, but it's still impossible to return to normal mode).

--------

Quick video (all VSCode extensions disabled and nvim config removed):

https://github.com/vscode-neovim/vscode-neovim/assets/122629585/d3f8a2ef-d85f-4ab8-a1d7-a418ed2ac6d7

### Steps To Reproduce

1. Create a new Jupyter notebook in VSCode.
2. Start editing a cell.
3. Press `R` to enter replace mode
4. Press `Esc`

### Expected Behavior

Return to normal mode.

penelopeysm:
This should close #1304; I've tried it out locally and everything seems to behave as expected. 🙂

(Force-pushed after linting.)

### DIFF
--- a/src/mode_manager.ts
+++ b/src/mode_manager.ts
@@ -16,7 +16,7 @@ export class Mode {
         return this.shortname.charAt(0).replace("\x16", "v");
     }
     // mode long name
-    public get name(): "insert" | "visual" | "cmdline" | "normal" {
+    public get name(): "insert" | "visual" | "cmdline" | "normal" | "replace" {
         switch (this.char.toLowerCase()) {
             case "i":
                 return "insert";
@@ -24,6 +24,8 @@ export class Mode {
                 return "visual";
             case "c":
                 return "cmdline";
+            case "r":
+                return "replace";
             case "n":
             default:
                 return "normal";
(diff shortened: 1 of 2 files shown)


theol0403:
I removed some unnecessary changes. Does that work for you?

theol0403:
The problem comes from https://github.com/vscode-neovim/vscode-neovim/blob/9479c650a3ff79382c6b86e84b16bc317fd3894f/package.json#L372

That reads as "in normal mode, send escape to neovim unless there are some available vscode escape commands such as dismissing popup, exiting cell notebook, etc" and "in non-normal (eg. insert) mode, always send escape to neovim". The problem was that it was treating replace mode as normal mode, and needed to treat it as non-normal mode.

penelopeysm:
Thanks for cleaning it up, and for the explanation; it works great on my end!

Issue #1331, opened 2023-07-21
feature: ^N/^P to navigate suggestions


martin-braun:
### Did you check the docs?

- [X] I have read all the vscode-neovim docs
- [X] I understand the fundamental limitations of the plugin and believe this request to be reasonable

### Is your feature request related to a problem? Please describe.

Navigating the suggestions in the code editor should be the Vim way. It's possible in VSCode using those bindings:

```json
[
  {
    "key": "ctrl+p",
    "command": "selectPrevSuggestion",
    "when": "suggestWidgetMultipleSuggestions && suggestWidgetVisible && textInputFocus"
  },
  {
    "key": "ctrl+n",
    "command": "selectNextSuggestion",
    "when": "suggestWidgetMultipleSuggestions && suggestWidgetVisible && textInputFocus"
  }
]
```

The latter ones are for the QuickFix list. Sadly, this can't work, because Neovim claims `ctrl+n` and `ctrl+p`, as we are in insert mode.

### Describe the solution you'd like

Ideally Neovim should initiate those navigation commands, but it would require to be able to parse the `When` expression, so it needs to know if a suggestion window is open.

### Describe alternatives you've considered

I have global vim bindings with one of the modifiers to trigger arrow keys, but I still think it would be nice to replicate what most LSPs offer.

martin-braun:
And issue in my configuration, it's already possible.

PR #1332, opened 2023-07-21
feat: quick fix navigation via ctrl+n/p

martin-braun:
Suggestions in Vim can be navigated with Ctrl+N/P, so I implemented those for QuickFix in this PR:

For navigation of the suggestions itself, it's a bit more complicated, see #1331

### DIFF
--- a/package.json
+++ b/package.json
@@ -1379,6 +1379,16 @@
                 "key": "ctrl+Escape",
                 "command": "workbench.action.focusActiveEditorGroup",
                 "when": "terminalFocus"
+            },
+            {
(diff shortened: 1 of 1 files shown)


theol0403:
Strange... for me on linux, ctrl-n/ctrl-p already works for BOTH suggestions and the quickfix menu.

martin-braun:
@theol0403 Okay, I got this. I send ctrl+n to Neovim and `vscode-neovim` does not as it seems. I should be able to revert those in my configuration, although it would limit `ctrl+n` and `ctrl+p` to VSCode exclusively.

Issue #1264, opened 2023-07-04
Issue Tracker: Refactor highlight provider


theol0403:
Currently, highlight provider is very bloated and buggy, especially to do with tabs in the file. Should be refactored.

- [x] https://github.com/vscode-neovim/vscode-neovim/issues/519
- [ ] https://github.com/vscode-neovim/vscode-neovim/issues/329
- [ ] https://github.com/vscode-neovim/vscode-neovim/issues/695
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/1004
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/1144
- [ ] https://github.com/vscode-neovim/vscode-neovim/issues/1205
- [ ] https://github.com/vscode-neovim/vscode-neovim/issues/1191

theol0403:
@justinmk would love ideas here.

## Problem 1: syntax off
Many neovim syntax-aware features (see linked issues) are broken, because `syntax off` is set. Ideally, we have `syntax on`.

Why is `syntax off``? Because we don't want to render neovim syntax highlights, which would look bad and spam the highlight provider. 

Why can't we ignore the syntax groups? Because many extensions will `hi Link` to syntax groups, so ignoring will break them. Also, even if they are ignored, they will clog up rpc because they are ignored on vscode side. 

### Potential Solutions

Ideally, we want a way to have `syntax on`, *without actually doing syntax highlighting*. This will support plugins linking to syntax groups, but avoid syntax highlighting and clogging rpc. 

However, I'm not sure what options are available to do this.

Another solution would be to somehow break all the `:hi link` to render as their own group. 

## Problem 1: syntax group ignoring
The way syntax group ignoring is currently done is messy and buggy. 

Lets follow an example. Say we want to ignore `MatchParen`. Neovim will create a highlight id containing `MatchParen`, and report it with `hl_attr_define` along with all decoration attributes. **Currently, vscode-neovim indexes highlights by group name, not id**. Thus, it will store `MatchParen` in a map, and also add the highlight id to a blacklist. 

The reason it still bothers storing the decorator is in the event that external buffers want to be used, with full syntax highlighting. 

Now, whenever `grid_line` reports a highlight id matching the blacklist, it will ignore it.

[...]

PR #1334, opened 2023-07-23
fact(highlight): ignore undesired highlights from vim side instead of vscode side

theol0403:
This attempts to simplify the highlight manager and improve the mechanism for ignoring syntax groups.

Instead of providing vscode options to do this, it instead relies on neovim configuration to effectively erase the syntax groups. 

Ref: https://github.com/vscode-neovim/vscode-neovim/issues/1264

### DIFF
--- a/runtime/lua/vscode/highlight.lua
+++ b/runtime/lua/vscode/highlight.lua
@@ -1,0 +1,98 @@
+local M = {}
+
+-- remove highlight groups that should never be shown
+vim.api.nvim_create_autocmd({ "BufEnter", "FileType", "ColorScheme" }, {
+    callback = function()
+        vim.api.nvim_set_hl(0, "Normal", {})
+        vim.api.nvim_set_hl(0, "NormalNC", {})
+        vim.api.nvim_set_hl(0, "NormalFloat", {})
+        vim.api.nvim_set_hl(0, "NonText", {})
+        vim.api.nvim_set_hl(0, "SpecialKey", {})
+        vim.api.nvim_set_hl(0, "TermCursor", {})
+        vim.api.nvim_set_hl(0, "TermCursorNC", {})
+        vim.api.nvim_set_hl(0, "Visual", {})
+        vim.api.nvim_set_hl(0, "VisualNOS", {})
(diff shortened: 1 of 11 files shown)


runtime/lua/vscode/highlight.lua
```diff
@@ -0,0 +1,17 @@
+-- disable matchparen because we don't need it
+vim.g.loaded_matchparen = 1
```
justinmk, line 2: I wonder if this kind of stuff should go in a single file like `vscode/defaults.lua`, so that it is easy to document, easy to discover, and leaves the door open for a future "skip" flag if users want it for some reason.

justinmk (approved):
Much less code = good.

The PR description sounds somewhat drastic. Does this disable all highlights from nvim or only some? Historically, highlight-driven nvim plugins such as https://github.com/justinmk/vim-sneak "mostly work" in vscode-neovim, although simplifying vscode-neovim may take precedence over continuing to support that (until a more robust, less complex mechanism is found).

theol0403:
@justinmk please see https://github.com/vscode-neovim/vscode-neovim/issues/1264#issuecomment-1646407147, this encompasses the solution for issue 2.

Before, vscode-neovim would ignore syntax groups on vscode side, using an inexact method because neovim reports syntax IDs with merged attributes, not syntax groups. This PR doesn't ignore any highlights on vscode side, but instead ignores on neovim side, simplifying the mechanism.

Plugins like sneak/leap should still work out of the box, and in fact more plugins should work. This is because I removed ignoring all syntax groups by default, because it didn't seem necessary (as syntax off is currently set), and ignored the highlight groups of certain plugins. 

I don't think this pr has any regressions.

theol0403:
> Does this disable all highlights from nvim or only some? 

Only the ones we want to ignore, by setting hi to NONE.

justinmk:
Ok, so the only problem is that the PR title is confusing, it sounds much more broad than it actually is.

theol0403:
That's fair, my bad.

justinmk (approved):

PR #1339, opened 2023-07-24
feat(lua): add lua api

theol0403:
This is a WIP for creating a lua API. For now, it just contains `notify` and `call`. I can't think of too many other options now, but the door is open in the future.

This attempts to properly use lua modules, though I am still unexperienced with best practices.

### DIFF
--- a/runtime/lua/vscode.lua
+++ b/runtime/lua/vscode.lua
@@ -1,28 +1,17 @@
-require("vscode.defaults")
-require("vscode.cursor")
-
--- used to execute vscode command
-local command_event_name = 'vscode-command'
--- used for extension communications
-local plugin_event_name = 'vscode-neovim'
+local api = require("vscode.api")
+local defaults = require("vscode.defaults")
+local cursor = require("vscode.cursor")
 
 local M = {}
 
-M.notify = function(command, ...)
-    return vim.rpcnotify(vim.g.vscode_channel, command_event_name, command, ...)
+M.notify = api.notify
+M.call = api.call
+M.notify_extension = api.notify_extension
+M.call_extension = api.call_extension
+
+M.setup = function()
+    defaults.setup()
+    cursor.setup()
 end
 
-M.call = function(command, ...)
-    return vim.rpcrequest(vim.g.vscode_channel, command_event_name, command, ...)
-end
-
-M.notify_extension = function(command, ...)
-    return vim.rpcnotify(vim.g.vscode_channel, plugin_event_name, command, ...)
-end
-
-M.call_extension = function(command, ...)
-    return vim.rpcrequest(vim.g.vscode_channel, plugin_event_name, command, ...)
-end
-
-_G.vscode = M
 return M
--- a/runtime/lua/vscode.lua
+++ b/runtime/lua/vscode.lua
@@ -6,8 +6,6 @@ local M = {}
 
 M.notify = api.notify
 M.call = api.call
-M.notify_extension = api.notify_extension
-M.call_extension = api.call_extension
 
 M.setup = function()
     defaults.setup()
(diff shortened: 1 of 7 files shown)


runtime/lua/vscode/cursor.lua
```diff
@@ -1,4 +1,5 @@
 local util = require("vscode.util")
+local vscode = require("vscode")
```
theol0403, line 2: This crashes because it requires vscode which requires cursor which requires vscode. What is the correct solution to this?

src/viewport_manager.ts
```diff
@@ -152,7 +152,7 @@ export class ViewportManager implements Disposable, NeovimRedrawProcessable, Neo
         }
         const viewport = this.gridViewport.get(gridId);
         if (viewport && startLine != viewport?.topline && currentLine == viewport?.line) {
-            this.client.executeLua("vscode.scroll_viewport(...)", [Math.max(startLine, 0), endLine]);
+            this.client.executeLua("vscode.util.scroll_viewport(...)", [Math.max(startLine, 0), endLine]);
```
theol0403, line 155: This fails because `util` is not a member of the vscode M export. Should I change this to a `require(vscode.util)`, or is there a better way for me to access the util module? I think the reason this currently works is `_G.vscode = vscode` in `vscode.lua`.

justinmk: `require('vscode.util')` is the normal idiomatic way.

I would drop `_G.vscode = vscode` unless there is a very good reason it's needed. It adds potential for conflicts with other plugins or user config, so better to avoid it if possible.

runtime/lua/vscode.lua
```diff
@@ -1,27 +1,27 @@
 require("vscode.defaults")
 require("vscode.cursor")
```
justinmk, line 2: generally a red flag if require() has side-effects. Modules should only return a table, so consumer modules use them like so:

local foo = require('vscode.cursor')

runtime/lua/vscode/cursor.lua
justinmk: Why does vscode require cursor? 

If the vscode.lua module is intended to be the thing that "gathers" all the necessary modules, then don't put any actual logic in vscode.lua.

justinmk (approved):

runtime/lua/vscode.lua
```diff
@@ -1,28 +1,17 @@
...
-    local current_height = vim.api.nvim_win_get_height(0)
-    local new_height = vscode_endline - vscode_topline + 1
-    -- resize height
-    if current_height ~= new_height then
-        vim.api.nvim_win_set_height(0, new_height)
-    end
+M.notify = api.notify
+M.call = api.call
+M.notify_extension = api.notify_extension
+M.call_extension = api.call_extension
```
justinmk, line 10: is notify_extension/call_extension a "public" API?

justinmk (approved):

src/viewport_manager.ts
```diff
@@ -152,7 +152,7 @@ export class ViewportManager implements Disposable, NeovimRedrawProcessable, Neo
         }
         const viewport = this.gridViewport.get(gridId);
         if (viewport && startLine != viewport?.topline && currentLine == viewport?.line) {
-            this.client.executeLua("vscode.scroll_viewport(...)", [Math.max(startLine, 0), endLine]);
+            this.client.executeLua("require('vscode.api').scroll_viewport(...)", [Math.max(startLine, 0), endLine]);
```
theol0403, line 155: This gets called on every single cursor move, is `require` performant enough in this case?

justinmk: require() is fully optimized

Issue #1263, opened 2023-07-04
Issue Tracker: Sync neovim options


theol0403:
We can connect buffer options according to https://github.com/vscode-neovim/vscode-neovim/pull/656#discussion_r912836951

- [ ] https://github.com/vscode-neovim/vscode-neovim/issues/175

PR #1340, opened 2023-07-24
feat(options): synchronize editor line number style

theol0403:
Fixes https://github.com/vscode-neovim/vscode-neovim/issues/1263, fixes https://github.com/vscode-neovim/vscode-neovim/issues/175, supersedes https://github.com/vscode-neovim/vscode-neovim/pull/656.

For now, this PR just syncs number/relative number, but lays the groundwork for syncing more options. This allows for smart relative numbers (https://github.com/jeffkreeftmeijer/vim-numbertoggle).

However, that specific plugin does not work, because it first checks if `nu` is set, but we force it to be off. 

A working snippet is 

```vim
augroup numbertoggle
  autocmd!
  autocmd BufEnter,FocusGained,InsertLeave,WinEnter * if mode() != "i" | set rnu   | endif
  autocmd BufLeave,FocusLost,InsertEnter,WinLeave   * if &nu           | set nornu | endif
augroup END
```

### DIFF
--- a/runtime/lua/vscode/force_options.lua
+++ b/runtime/lua/vscode/force_options.lua
@@ -56,3 +56,18 @@ forceoptions(vim.opt)
 vim.api.nvim_create_autocmd({ "BufEnter", "FileType" }, {
     callback = function() forceoptions(vim.opt_local) end,
 })
+
+-- send option changes to vscode, and then reset
+vim.api.nvim_create_autocmd({ "OptionSet" }, {
+    callback = function(ev)
+        vim.fn.VSCodeExtensionNotify("option-set", vim.fn.win_getid(), ev.match,
+            {
+                option_type = vim.v.option_type,
+                option_new = vim.v.option_new,
+                option_oldlocal = vim.v.option_oldlocal,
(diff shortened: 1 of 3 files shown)


theol0403:
Actually, this does not seem to work. It seems setting inside the `InsertLeave`/`InsertEnter` autocommand skips `OptionSet`.

justinmk:
> It seems setting inside the `InsertLeave`/`InsertEnter` autocommand skips `OptionSet`.

That is intentional; you can opt-in to recursion by using `++nested` (`:help autocmd-nested`).

But that means any users or nvim plugins that are creating things like your `augroup numbertoggle` example must specify `++nested` on their autocmd definitions.

There is also the `option_set` UI event (`:help ui-global`) which is emitted on the RPC channel, but it only covers certain options.

theol0403:
Yeah, that doesn't include number. I'm not sure if there is any way of doing this then.

justinmk:
The approach here addresses #175 , no? It sounds like people just want to do `:set nu` from a mapping or the cmdline, neither of which is a nested autocmd.

Issue #1266, opened 2023-07-04
Issue Tracker: Refactor document change manager (remove dot-repeat hack)


theol0403:
We want to refactor two-way sync.

- [x] https://github.com/vscode-neovim/vscode-neovim/pull/992
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/543
- [ ] if we send inputs using nvim_input, we can remove dot-repeat hack, composite escape, etc
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/543
- [ ] https://github.com/vscode-neovim/vscode-neovim/issues/625
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/644
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/1056
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/1068
- [x] https://github.com/vscode-neovim/vscode-neovim/issues/1187
- [ ] https://github.com/vscode-neovim/vscode-neovim/pull/948

theol0403:
Here are my new ideas for removing dot repeat. I had some old ones at https://github.com/vscode-neovim/vscode-neovim/pull/992#issuecomment-1228734888, but they mostly involved trying to send edits to nvim in a way that would be picked up by dot repeat. Directly writing to dot repeat is going to be added to nvim at some point according to @justinmk, but not yet iirc. 

My new idea takes inspiration from the way that many nvim plugins handle dot repeat: https://www.reddit.com/r/neovim/comments/wkqkzf/comment/ijownam/?utm_source=share&utm_medium=web2x&context=3

The challenge is that I need to construct a string to save in operatorfunc. 

There are two strategies:
1. bind vscode `type` event and record keystrokes before sending them back to vscode. This is a major design decision but the performance should be good with the new extension affinity settings. Whenever an arrow key or cursor is moved, dot repeat is reset.
2.  construct a set of typed changes by parsing the sequential editor changes produced by vscode, and trying to calculate what the user actually typed based on document changes. This is tricky however due to things like auto bracket pairs and discontinuous edits.

This should delete a lot of code.

theol0403:
I have narrowed it down to two options:

### bind vscode `type`

I have figured out how to not break completions, and the lag seems acceptable. 

#### The challenge
Edits come in both with `type` and `onDocumentChanged`. How can I merge them together? Ideas:

- when type, wait for document changed, and cancel it out surgically
[...]

PR #1347, opened 2023-07-27
feat(document_change): extract nvim_input from document change

theol0403:
Ref: https://github.com/vscode-neovim/vscode-neovim/issues/1266

This removes the need for the custom dot repeat hack and composite escape. For composite escape to work it currently needs https://github.com/max397574/better-escape.nvim.

I'm not sure if this is a viable option, just due to edge cases and bugs. However, worth investigating. 

Todo:

- [ ] handle moving cursor (like with auto parens)
- [ ] ignore document changes while in insert mode?
- [ ] consider what happens with pasting a line of code containing a composite escape
- [ ] try to remove need for better-escape (use vscode cursor instead, and ignore large movements)
- [ ] simplify code
- [ ] isolate situations where nvim_input is used

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ mode and editor commands, making the best use of both editors.
     -   [VSCode specific differences](#vscode-specific-differences)
     -   [Troubleshooting](#troubleshooting)
         -   [Performance problems](#performance-problems)
-    -   [Composite escape keys](#composite-escape-keys)
     -   [Jumplist](#jumplist)
     -   [Wildmenu completion](#wildmenu-completion)
     -   [Multiple cursors](#multiple-cursors)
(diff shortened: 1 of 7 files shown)


xiyaowong:
Will close this PR. It's a good idea, but it won't work:

1. Performance issues. Any modifications from vscode such as paste, formatting, revert, etc., are split into inputs, undoubtedly causing significant performance issues.
2. Edge cases:
   - What happens if a user pastes text containing a map key sequence in insert mode, how to handle it? (Very difficult)
   - What if the user has other input during insertion?
   - Dotrepeat no longer works correctly, undo stacks in nvim and vscode may be corrupted.
   - Multiple cursor editing in vscode.
   - Modifications from external sources, multiple documents being modified simultaneously, modified buffers not being the current buffer.
   - etc.

PR #1350, opened 2023-07-27
feat(main): Add option to pass startup command

nimaipatel:
This adds `vscode-neovim.neovimCmd`  to pass startup command using `--cmd`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,11 @@
                     "default": false,
                     "markdownDescription": "Equivalent to the startup option `--clean`."
                 },
+                "vscode-neovim.neovimCmd": {
+                    "type": "string",
+                    "default": "",
+                    "markdownDescription": "Command passed to startup option `--cmd`, see `:h --cmd`"
+                },
                 "vscode-neovim.NVIM_APPNAME": {
                     "type": "string",
                     "default": "",
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -38,6 +38,7 @@ export interface ControllerSettings {
     useWsl: boolean;
     customInitFile: string;
     clean: boolean;
+    neovimCmd: string;
     NVIM_APPNAME: string;
     neovimViewportWidth: number;
     neovimViewportHeightExtend: number;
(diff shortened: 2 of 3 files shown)


nimaipatel:
Please review for @theol0403, #1350

src/main_controller.ts
```diff
@@ -129,6 +130,9 @@ export class MainController implements vscode.Disposable {
         if (settings.clean) {
             args.push("--clean");
         }
+        if (settings.neovimCmd) {
+            args.push("--cmd", settings.neovimCmd);
```
theol0403, line 134: I don't know the answer to this, but what happens if `neovimCmd` contains other flags, like "--clean"... will it be properly escaped?

package.json
```diff
@@ -93,6 +93,11 @@
                     "default": false,
                     "markdownDescription": "Equivalent to the startup option `--clean`."
                 },
+                "vscode-neovim.neovimCmd": {
+                    "type": "string",
```
justinmk, line 97: This should be array, so that the args can be passed without dealing with quoting.

justinmk:
Reference the original issue and mention the use-case in the PR/commit message. 

The implementation here is fragile. The original use-case as I understood it is "I want to use settings.json to store any arbitrary Nvim config". If that use-case is truly worth supporting, then it should be done in a robust and general way, not "this fixes my immediate impulse, good enough".

src/main_controller.ts
nimaipatel: The argument after `--cmd` is always parsed and executed as vimL.

justinmk: So if someone wants more control (to pass multiple `-c`, `--cmd`, etc. args), we'll have to add another option. 

To avoid that, perhaps this option should be `nvimArgs` array. Then people can have settings like:

```
"nvimArgs" = [
  "--cmd", "set nowrap",
  "-c", "edit foo",
  ...
]
```

nimaipatel: I had originally considered this, but command-line options are already used to implement other settings. For example, if  `vscode-neovim.neovimClean` is set to true along and `nvimArgs = [ "--cmd" ]` then the final neovim command would be `nvim --cmd --clean` which would result in an error. 

The solution would be to append `nvimArgs` at the end after everything else if set. Perhaps we could even name it `nvimExtraArgs`.

justinmk: > The solution would be to append `nvimArgs` at the end after everything else is set.

Yes, of course. But even just one command can break stuff, depending on what is sent. 

The setting description can mention what the default args are.

Issue #1355, opened 2023-07-29
How to use VSCodeNotify() instead of VSCodeNotifyRange() after recent breaking changes with 'editor.action.reindentselectedlines' command


still-dreaming-1:
Previously when I called VSCodeNotifyRange had several parameters it accepted. Here is an example of how I used it before from my Lua config:

```
-- change o to also use VS Code to auto indent
normalMap('o', "o<Cmd>call VSCodeNotifyRange('editor.action.reindentselectedlines', line('.'), line('.'), 1)<CR>")
```

The idea with this command is to change the behavior of `o` to both start a new line and get VSCode to indent it.

Currently if I just change it to use `VSCodeNotify()` and only pass the command, it leaves the line unindented. If I then manually run that reindentselectedlines command, it does indent the line.

I also tried using `V` to visually select the entire line before calling `VSCodeNotify('editor.action.reindentselectedlines')`, but it still doesn't work.

On the new, unindented line, I can also manually run (via `:`) `call VSCodeNotify('editor.action.reindentselectedlines')` and it will indent the line.

theol0403:
Thanks for this. I didn't realize there was a usecase for the function without simply forwarding visual mode. I'll re-introduce it shortly.

theol0403:
Actually, this is already possible. I'll still re-add because of https://github.com/vscode-neovim/vscode-neovim/issues/1351.

The problem is that the cursor update from `o` isn't sent to vscode by the time that the command is sent. I think I'll also solve this.

theol0403:
@still-dreaming-1 in the linked PR, using `VSCodeNotify` in your example will now work.

still-dreaming-1:
If updating the cursor after `o` is possible, enabling the use of VSCodeNotify() for this use case, that sounds like the best solution.

theol0403:
That is what the PR does.

still-dreaming-1:
Is this supposed to be fixed in v0.4.1 of this extension? I am currently on that version, and my mapping is running the 'editor.action.reindentselectedlines' command on the line the cursor started on instead of the line it is on at the time the command is run. I am basing this conclusion on the fact that the cursor is at the very start of the new line. Also, if I intentionally mess up the indention of the line the cursor is one before pression `o`, the indentation of that line does get fixed. Here is the mapping I am using from my init.lua:

[...]

Issue #1351, opened 2023-07-27
0.4.0:  keep the `VSCodeNotifyRange` function


xiyaowong:
### Did you check the docs?

- [X] I have read all the vscode-neovim docs
- [X] I understand the fundamental limitations of the plugin and believe this request to be reasonable

### Is your feature request related to a problem? Please describe.

Some code actions, quickfixes, and refactoring operations enter the snippet mode after execution. These operations can be performed normally from the normal mode of nvim, but they do not enter the snippet mode. Please see the GIFs below.

Can't enter snippet mode
![demo](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/cfa07af1-568c-4834-aceb-bd23f91fcece)

My previous approach was to first save the selected range, then switch to insert mode, and then use that range to call `VSCodeNotifyRange`
```lua

local invoke = function(cmd)
  local mode = api.nvim_get_mode().mode
  local target
  if mode == 'V' then
    local startLine = fn.line('v')
    local endLine = fn.line('.')
    target = function() fn.VSCodeNotifyRange(cmd, startLine, endLine, true) end
  elseif mode == 'v' then
    local startPos, endPos = getpos()
    target = function()
      fn.VSCodeNotifyRangePos(cmd, startPos[2], endPos[2], startPos[3], endPos[3], true)
    end
  else
    target = function() fn.VSCodeNotify(cmd) end
  end
  xy.feedkeys('<esc><esc>')
  vim.defer_fn(function()
    vim.cmd.startinsert()
    target()
  end, 20)
end
```
</details>

![demo](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/531a064e-21fd-43e5-a943-064325e1863a)

 

### Describe the solution you'd like

Provide `VSCodeNotifyRange` function. Alternatively, this use case could be supported from within the extension itself.

### Describe alternatives you've considered

-

theol0403:
From how I understand your snippet, can't you just replace `VSCodeNotifyRange` with `VSCodeNotify`? 

Now, `VSCodeNotify` will automatically send the visual selection to vscode if you are in visual mode, and then you can do the same thing after where you escape and enter insert mode.

theol0403:
Oh I see the problem, I missed where you send the command after you enter insert mode. Maybe this is a useful enough reason to add.

doppio:
[...]

Issue #1345, opened 2023-07-26
bug: update broke line commenting


vlrevolution:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.4

### Operating system/version

WSL2

### Describe the bug

xmap gc  <Plug>VSCodeCommentary
nmap gc  <Plug>VSCodeCommentary
omap gc  <Plug>VSCodeCommentary
nmap gcc <Plug>VSCodeCommentaryLine

these binds are no longer working, please check

### Steps To Reproduce

put in init.vim and try 

### Expected Behavior

should comment line

hustrjh:
Have you checked the breaking change in this version?
![image](https://github.com/vscode-neovim/vscode-neovim/assets/69783980/f25b2940-94f2-4522-a4a4-774dc4f0bce1)

theol0403:
It is indeed removed.

If you want to replace `VSCodeCommentary`, call `VSCodeCall('editor.action.commentLine')` instead like https://github.com/vscode-neovim/vscode-neovim/blob/214acebf9dc3e748d875b21bf70f3025e86933be/vim/vscode-code-actions.vim#L5.

crash5:
> Have you checked the breaking change in this version? ![image](https://user-images.githubusercontent.com/69783980/256398688-f25b2940-94f2-4522-a4a4-774dc4f0bce1.png)

Sorry, but it does not say anything about breaking the VSCodeCommentary command. Only states that it removes the Visual and Range suffixes. (maybe if you know the internals of the vscode-neovim plugin internals it is clear for you, but for me it was not...)

theol0403:
My bad, forgot to document it.

crash5:
> https://github.com/vscode-neovim/vscode-neovim/blob/214acebf9dc3e748d875b21bf70f3025e86933be/vim/vscode-code-actions.vim#L5

Is it possible to make it work with motions? (Can you put  this snippet into the readme? I think there will be peoples looking for it soon :) )

theol0403:
Humm, no it isn't possible. The reason I removed it is because visual selections are now automatically sent. However, I did not think of operator-pending modes, I didn't know it supported it before.

Would it be a problem if you just used any of the vim/neovim comment plugins?

[...]

PR #1358, opened 2023-07-30
fix(cursor): re-add new VSCodeNotifyRange impl

theol0403:
Adds `VSCodeNotifyRange` but with a more concise and accurate impl. Fixes https://github.com/vscode-neovim/vscode-neovim/issues/1355, fixes https://github.com/vscode-neovim/vscode-neovim/issues/1351, fixes https://github.com/vscode-neovim/vscode-neovim/issues/1345.

Also re-adds `VSCodeCommentary` and fixes a bug with `VSCodeNotify`.

### DIFF
--- a/runtime/lua/vscode-neovim/api.lua
+++ b/runtime/lua/vscode-neovim/api.lua
@@ -23,6 +23,23 @@ function M.call_extension(command, ...)
     return vim.rpcrequest(vim.g.vscode_channel, plugin_event_name, command, { ... })
 end
 
+-- send command to vscode with range (line or char). [1, 1]-based.
+function M.call_range(command, line1, line2, leaveSelection, ...)
+    return M.call_extension('range-command', command, 'V', line1, line2, 1, 1, leaveSelection, { ... })
+end
+
+function M.notify_range(command, line1, line2, leaveSelection, ...)
+    return M.notify_extension('range-command', command, 'V', line1, line2, 1, 1, leaveSelection, { ... })
+end
+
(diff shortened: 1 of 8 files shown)


runtime/lua/vscode-neovim/api.lua
```diff
@@ -23,6 +23,23 @@ function M.call_extension(command, ...)
     return vim.rpcrequest(vim.g.vscode_channel, plugin_event_name, command, { ... })
 end
 
+-- send command to vscode with range (line or char). [1, 1]-based.
+function M.call_range(command, line1, line2, leaveSelection, ...)
+    return M.call_extension('range-command', command, 'V', line1, line2, 1, 1, leaveSelection, { ... })
+end
+
+function M.notify_range(command, line1, line2, leaveSelection, ...)
+    return M.notify_extension('range-command', command, 'V', line1, line2, 1, 1, leaveSelection, { ... })
```
justinmk, line 32: Seems overwrought to have all these dedicated variants. `call()` and `notify()` could cover all these variations with a "kwargs" style dict parameter. That would also simplify the documentation. It also makes the calls, and usage, much easier to understand, instead of having 4+ positional args.

The variadic args `...` could just be a list, there's no benefit to variadic args here: they are hard to visually parse vs the non-variadic args.

```lua
notify('foo', { arg1, arg2, … } {
  line1=…,
  line2=…,
  col1=…,  -- optional, treat as "line" if `nil`
  col2=…,  -- optional, treat as "line" if `nil`
}
```

PR #1403, opened 2023-08-29
fix(cursor): decrease delay for debounce

xiyaowong:
200ms is too long and will cause the cursor to be out of sync when moving quickly and switching modes.

Can reproduced by: `llllv`, `lllla`, `2lv`, `2jv`, `2la`...
cursor-out-of-sync

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -345,7 +345,7 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
                 await this.updateNeovimVisualSelection(editor, selection);
             }
         },
-        200,
+        20,
         { leading: false, trailing: true },
     );
 


theol0403:
I'm not sure if this is the real issue.

* those commands work for me with no problems, could it be your config? do you use a gj/gk mapping?
* that debounce is ONLY for vscode->neovim sync... when typing commands rapidly in neovim, there should be no input from vscode 
* a long debounce (could be shorter) is needed for using the mouse to create a visual selection, we don't want to spam neovim with cursor position updates

xiyaowong:
Oh, I mapped jk to vscode' cursorMove.

theol0403:
Yeah, so the issue is that cursorMove is handled by vscode, so there is indeed a delay in vscode updating neovim with the new position. 

I'm not sure what the best compromise is, the lag isn't ideal, but also we do need some debouncing for jitter/sync reasons.

xiyaowong:
Need for a more comprehensive approach

PR #1406, opened 2023-08-31
fix(cursor): optimized cursor synchronization debouncing

### LINKED ISSUE #1404: bug: Selection using mouse is not syncing fast enough

### Neovim version (nvim -v)

v0.9.1

### Operating system/version

macos 13.5.1

### Describe the bug

I am with an empty init.lua and a vscode with only neovim plugin.

I noticed that if I have the cursor somewhere, and select using the mouse some other chars, and tapping `x` to delete, the previously selected char will be deleted, which is wrong. If I do the same, wait for a second, and then press `x` then the new selection will be deleted. But doing it right after the mouseup, the wrong thing is deleted.

### Steps To Reproduce

I am with an empty init.lua and a vscode with only neovim plugin.

- I have a regular file with many lines.
- I click once, on say line 2 on any character, and the cursor is on that character.
- Then I click and drag on line 5 up to line 8 to select a bunch of characters/lines
- Immediately tap `x` to delete the selected chars/lines, but the character from line 2 that was previously on the cursor is deleted.

I noticed that if I wait for a second, and then press `x` then the chars/lines will be deleted. But doing it right after the mouseup, the wrong thing is deleted.

### Expected Behavior

The newly selected chars should be the ones being deleted always.

xiyaowong:
Resolve #1404

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -63,6 +63,12 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
     private debouncedCursorUpdates: WeakMap<TextEditor, DebouncedFunc<CursorManager["updateCursorPosInEditor"]>> =
         new WeakMap();
 
+    // Different change kinds use different debounce times
+    private debouncedApplySelectionChanged: Map<number, DebouncedFunc<CursorManager["applySelectionChanged"]>> =
+        new Map();
+    // A flag indicates that func still pending.
+    private previousApplyDebounceTime: number | undefined;
+
     public constructor(
         private logger: Logger,
         private client: NeovimClient,
@@ -195,16 +201,20 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
         if (!modeConf) {
             return;
         }
+        let style;
+        if (modeName == "visual") {
+            // in visual mode, we try to hide the cursor because we only use it for selections
+            style = TextEditorCursorStyle.LineThin;
+        } else if (modeConf.cursorShape === "block") {
+            style = TextEditorCursorStyle.Block;
+        } else if (modeConf.cursorShape === "horizontal") {
+            style = TextEditorCursorStyle.Underline;
+        } else {
+            style = TextEditorCursorStyle.Line;
+        }
         for (const editor of window.visibleTextEditors) {
-            if (modeName == "visual") {
-                // in visual mode, we try to hide the cursor because we only use it for selections
-                editor.options.cursorStyle = TextEditorCursorStyle.LineThin;
-            } else if (modeConf.cursorShape === "block") {
-                editor.options.cursorStyle = TextEditorCursorStyle.Block;
-            } else if (modeConf.cursorShape === "horizontal") {
-                editor.options.cursorStyle = TextEditorCursorStyle.Underline;
-            } else {
-                editor.options.cursorStyle = TextEditorCursorStyle.Line;
+            if (editor.options.cursorStyle !== style) {
+                editor.options.cursorStyle = style;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/cursor_manager.ts
```diff
@@ -195,16 +201,20 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
         if (!modeConf) {
             return;
         }
+        let style;
```
theol0403, line 204: Should probably have type annotation

src/cursor_manager.ts
```diff
@@ -195,16 +201,20 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
...
-                // in visual mode, we try to hide the cursor because we only use it for selections
-                editor.options.cursorStyle = TextEditorCursorStyle.LineThin;
-            } else if (modeConf.cursorShape === "block") {
-                editor.options.cursorStyle = TextEditorCursorStyle.Block;
-            } else if (modeConf.cursorShape === "horizontal") {
-                editor.options.cursorStyle = TextEditorCursorStyle.Underline;
-            } else {
-                editor.options.cursorStyle = TextEditorCursorStyle.Line;
+            if (editor.options.cursorStyle !== style) {
+                editor.options.cursorStyle = style;
```
theol0403, line 217: Probably don't need the comparison here and let vscode handle the update

src/cursor_manager.ts
```diff
@@ -302,34 +312,50 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
...
-                    `${LOG_PREFIX}: Skipping onSelectionChanged event since it was selection produced by doc change`,
-                );
-                return;
-            }
+    private getDebouncedApplySelectionChanged = (
+        kind: TextEditorSelectionChangeKind | undefined,
+    ): DebouncedFunc<CursorManager["applySelectionChanged"]> => {
+        const debounceTime =
+            // Should use same debounce time if previous debounced func still pending
+            this.previousApplyDebounceTime !== undefined
```
theol0403, line 325: Maybe this should be made more readable by using if statement, switch statement, extracting into function (then you can return guard the previous time), or map?

src/cursor_manager.ts
```diff
@@ -302,34 +312,50 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
...
+                ? this.previousApplyDebounceTime
+                : kind === TextEditorSelectionChangeKind.Mouse
+                ? 150
+                : 50;
+        this.previousApplyDebounceTime = debounceTime;
+
+        let func = this.debouncedApplySelectionChanged.get(debounceTime);
+        if (func) return func;
+        func = debounce(this.applySelectionChanged, debounceTime, { leading: false, trailing: true, maxWait: 10 });
```
theol0403, line 334: Doesn't maxWait: 10 ruin the whole purpose of this?

src/cursor_manager.ts
```diff
@@ -302,34 +312,50 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
...
+        this.debouncedApplySelectionChanged.set(debounceTime, func);
+        return func;
+    };
 
+    private applySelectionChanged = async (editor: TextEditor, kind: TextEditorSelectionChangeKind | undefined) => {
```
theol0403, line 339: Lint

src/cursor_manager.ts
```diff
@@ -365,8 +390,9 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
             `${LOG_PREFIX}: Updating cursor pos in neovim, winId: ${winId}, pos: [${pos.line}, ${pos.character}]`,
         );
         const vimPos = [pos.line + 1, pos.character]; // nvim_win_set_cursor is [1, 0] based
-        const request: [string, unknown[]][] = [["nvim_win_set_cursor", [winId, vimPos]]];
-        await callAtomic(this.client, request, this.logger, LOG_PREFIX);
+        await this.client.request("nvim_win_set_cursor", [winId, vimPos]); // a little faster
```
theol0403, line 393: Is this measurably faster, or you just think it will be because there are a little fewer stack calls?

theol0403 (changes requested):
Looks good! Nice solution for the problem.

src/cursor_manager.ts
xiyaowong: I did a simple test and it took 2ms less on average using request.

src/cursor_manager.ts
```diff
@@ -320,23 +313,29 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
...
+            debounceTime = 150;
+        } else {
+            debounceTime = 50;
+        }
         this.previousApplyDebounceTime = debounceTime;
 
         let func = this.debouncedApplySelectionChanged.get(debounceTime);
         if (func) return func;
-        func = debounce(this.applySelectionChanged, debounceTime, { leading: false, trailing: true, maxWait: 10 });
+        func = debounce(this.applySelectionChanged, debounceTime);
```
theol0403, line 330: Keep it explicit?
```suggestion
        func = debounce(this.applySelectionChanged, debounceTime, { leading: false, trailing: true });
```

src/cursor_manager.ts
```diff
@@ -320,23 +313,29 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
...
-                : kind === TextEditorSelectionChangeKind.Mouse
-                ? 150
-                : 50;
+        let debounceTime: number;
+        // Should use same debounce time if previous debounced func still in progress
+        // This avoid multiple cursor updates with different positions at the same time
+        if (this.previousApplyDebounceTime !== undefined) {
+            debounceTime = this.previousApplyDebounceTime;
+        } else if (kind === TextEditorSelectionChangeKind.Mouse) {
+            debounceTime = 150;
```
theol0403, line 322: To better solve the linked issue, maybe 100ms?

src/cursor_manager.ts
theol0403: Cool, delete commented lines

theol0403 (approved):
Thanks!

xiyaowong:
Done

BTBMan:
It's reproduced now in the latest version. @xiyaowong

xiyaowong:
@BTBMan https://github.com/vscode-neovim/vscode-neovim/issues/1409#issuecomment-1704471139

PR #1412, opened 2023-09-03
fix(commands): add delay to wait for other possible events

### LINKED ISSUE #1355: How to use VSCodeNotify() instead of VSCodeNotifyRange() after recent breaking changes with 'editor.action.reindentselectedlines' command

xiyaowong:
Currently not robust

Fixes #1355

### DIFF
--- a/src/custom_commands_manager.ts
+++ b/src/custom_commands_manager.ts
@@ -1,8 +1,9 @@
 import { commands, Disposable, window } from "vscode";
 
 import { Logger } from "./logger";
-import { NeovimCommandProcessable } from "./neovim_events_processable";
 import { MainController } from "./main_controller";
+import { NeovimCommandProcessable } from "./neovim_events_processable";
+import { wait } from "./utils";
 
 export class CustomCommandsManager implements Disposable, NeovimCommandProcessable {
     private disposables: Disposable[] = [];
@@ -19,8 +20,10 @@ export class CustomCommandsManager implements Disposable, NeovimCommandProcessab
     public async handleVSCodeCommand(command: string, args: unknown[]): Promise<unknown> {
         const editor = window.activeTextEditor;
         if (!editor) return;
+        // Give some time to other possible events  to weak up
+        // 10ms basically has no side effects, even `cursorMove` which may be called frequently
+        await wait(10);
         await this.main.cursorManager.waitForCursorUpdate(editor);
-        const res = await commands.executeCommand(command, ...args);
-        return res;
+        return commands.executeCommand(command, ...args);
     }
 }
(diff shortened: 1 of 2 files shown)


xiyaowong:
@theol0403 Perhaps we can simply make VSCodeNotify deferred, as it is itself an asynchronous API

theol0403:
@xiyaowong what is the problem with this PR, is it that adding the delay is not good for blocking user code? In that case I agree we can make it async, I thought it already was.

xiyaowong:
When run
```lua
 vim.api.nvim_feedkeys('o', 'n', false) -- redraw
 vim.fn.VSCodeNotify('editor.action.reindentselectedlines') -- command
```
The order of events is `command`, `redraw`.

When run
```lua
  vim.api.nvim_feedkeys('o', 'n', false) -- redraw
  vim.defer_fn(function () -- command
    vim.fn.VSCodeNotify('editor.action.reindentselectedlines')
  end, 0)
```
The order of events is `redraw`, `command`.

The issue is caused by the mechanism of nvim. Add delay on vscode side isn't reliable. Therefore, it's recommended to wrap the `VSCodeNotify` to a deferred function to avoid this issue in the root.

theol0403:
I see, thanks, that sounds good.

src/custom_commands_manager.ts
```diff
@@ -19,8 +20,10 @@ export class CustomCommandsManager implements Disposable, NeovimCommandProcessab
     public async handleVSCodeCommand(command: string, args: unknown[]): Promise<unknown> {
         const editor = window.activeTextEditor;
         if (!editor) return;
+        // Give some time to other possible events  to weak up
+        // 10ms basically has no side effects, even `cursorMove` which may be called frequently
+        await wait(10);
```
justinmk, line 25: This is fragile and not a good tradeoff. If the goal is to wait until input is "flushed" in the Nvim server, you can do that by calling a non-fast Nvim API function, e.g. `nvim_eval('')`.

Issue #498, opened 2021-01-01
BufferManager: nvim_win_set_cursor: Cursor position outside buffer


David-Else:
Using the latest 0.0.73 I got this error when I came back to an unsaved file in normal mode. The cursor was frozen, then after moving around some other files and returning then it worked again.

```
2021-01-01T20:38:46.607Z BufferManager: nvim_win_set_cursor: Cursor position outside buffer
```
Not sure how to reproduce, but I thought I would post it as an issue in case anyone else got it.

braxtons12:
I also get this happening, but I haven't been able to reduce this to a reliably reproducible situation, as it seems kind of random.

It _seems_ that most of the time this happens when changing focus (with a mouse click) from an open editor to another one or the explorer, etc. is what triggers this, but it's inconsistent. Sometimes I can go 4 hours without it ever happening, sometimes it happens every 20 minutes. I've also had it happen when opening vscode and trying to immediately type in one of the restored/reopened editors from the last session.

I'm currently using version 0.0.82 and have experienced it on both Windows 10 and Ubuntu 20.10 and 21.04.

achikin:
The first time it happened was when I did View -> Editor layout -> Split in Group and then it followed every split/unsplit action until I restarted VSCode and then it has gone. Can't reproduce. 
```
VSCode Neovim 0.0.8.2
------------------------
NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@BigSur

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/0.6.1/share/nvim"
-------------------------
MacOS 11.6
```

Ahmad-Zaklouta:
I am having the same issue all the time when I move between tabs.

wenhao-eb:
> I am having the same issue all the time when I move between tabs.

I have the issue in this same situation too.

coderj001:
I have same issue

```
NVIM v0.5.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #1416, opened 2023-09-03
fix(buffer): the output document should be modifiable

xiyaowong:
Fixed errors related to "output" in #498, but not sure if all scenarios are covered. Actually, I haven’t encountered any other situations either.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -170,7 +170,7 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
 
     public isExternalTextDocument(textDoc: TextDocument): boolean {
         // !Output should be modifiable, vscode treats it as a regular document.
-        // !When the option "modifiable" is set to false, nvim_buf_set_text will not work.
+        // !When the option "modifiable" is set to false, nvim_buf_set_text will not work. #498
         // !Don't remove this, cause it's a long time bug
         if (textDoc.uri.scheme === "output") {
             return false;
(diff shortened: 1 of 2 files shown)


theol0403:
I think hardcoding output to be true is so that syntax highlighting is enabled for the output buffer.

xiyaowong:
I don't think so. 

The output should not be treated as an external buffer, and the [code](https://github.com/vscode-neovim/vscode-neovim/blob/7094dc7208d7a4dca696bd131513a1a86b375fde/src/buffer_manager.ts#L228C1-L234C61) is also the same. Output is always created in vscode and always has `b:vscode_controlled`

src/buffer_manager.ts
```diff
@@ -169,8 +169,11 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
     }
 
     public isExternalTextDocument(textDoc: TextDocument): boolean {
+        // !Output should be modifiable, vscode treats it as a regular document.
```
justinmk, line 172:
```suggestion
        // !Output should be modifiable, vscode treats it as a regular document. #498
```

src/buffer_manager.ts
```diff
@@ -169,8 +169,11 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
...
+        // !When the option "modifiable" is set to false, nvim_buf_set_text will not work.
```
justinmk, line 173:
```suggestion
        // !When the option "modifiable" is set to false, nvim_buf_set_text will not work. #498
```

xiyaowong:
@theol0403  It's best to merge this PR as soon as possible, as an error will affect other debugging.

Issue #1251, opened 2023-07-02
bug: Combine status line items


martin-braun:
### Did you check the docs?

- [X] I have read all the vscode-neovim docs
- [X] I understand the fundamental limitations of the plugin and believe this request to be reasonable

### Is your feature request related to a problem? Please describe.

`q` recording is fully supported, but you can't see if you are recording or not.

### Describe the solution you'd like

It would be nice if you could see that you are recording a register in the output window of VSCode, as close as possible to the original behavior:

<img width="138" alt="Screenshot 2023-07-02 at 11 55 50" src="https://github.com/vscode-neovim/vscode-neovim/assets/24259245/666d0daf-f74c-427f-a82c-5daa170c152f">

### Describe alternatives you've considered

If you press `q` without seeing the prefix in the status bar, it means you were recording. I'm pressing `q` so I see the prefix in the status bar. I'm pressing `q` once more and now I know I'm recording to `@q`.

Everything seem to work, which is nice, but it still would be nice to see if I'm recording right now or not and to which register I'm recording.

justinmk:
Could handle the RecordingEnter/RecordingLeave events, but hopefully this is already reflected in the `:help ui` UI events.

theol0403:
This already exists for me... do you have a vim setting that possibly turns this off? 

![image](https://github.com/vscode-neovim/vscode-neovim/assets/16546293/9476b2a1-187c-408b-b3a2-e1d7f6e6fca2)

martin-braun:
@theol0403 Ha! Got it. Sometimes it likes to find itself on the far right, sometimes it does not: 

<img width="2048" alt="Screenshot 2023-07-03 at 06 06 04" src="https://github.com/vscode-neovim/vscode-neovim/assets/24259245/3b299938-5783-40a2-a568-dfff6a41cf82">

Still an issue, I expect it to be at the position where "Type :qa and press <Enter> to exit Nvim".

Both status bar widgets share the name "VSCode Neovim (Extension)". Unselecting it in the context menu will only hide the left one: 

<img width="242" alt="Screenshot 2023-07-03 at 06 11 35" src="https://github.com/vscode-neovim/vscode-neovim/assets/24259245/b6c7cf32-5416-4d82-a535-06c6bbed89b2">

But right clicking the right one will allow me to hide it as well: 

[...]

PR #1429, opened 2023-09-07
feat(statusline): combine status line items

xiyaowong:
Currentlly, we use `window.titleSeparator` as the separator. Of course, it is best to provide a separate configuration, whether in vscode or nvim. But the priority is not high.

Resolve: #1251


xiyaowong:
Currentlly, we use `window.titleSeparator` as the separator. Of course, it is best to provide a separate configuration, whether in vscode or nvim. But the priority is not high.

justinmk (approved):
Will be a nice quality of life improvement, thanks!

Issue #1374, opened 2023-08-06
bug: Weird virtual characters


ahmed-yasser1:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

NVIM v0.9.1

### Operating system/version

Pop!_OS 22.04 LTS x86_64

### Describe the bug

when I enter visual mode and select some lines I find some weird characters at the end of some lines.

as you see in the following picture

![bug](https://github.com/vscode-neovim/vscode-neovim/assets/99615534/e8c04dbd-794a-486f-8fd1-dbcf3994b69c)

__NOTE__ I did not face this bug in version `0.3.2`

### Steps To Reproduce

1. go to visual mode
2. just select some lines 
3. sometimes these characters appear and sometimes not, so try step 1 and 2 until you find them

### Expected Behavior

What I expect to find that when I go to visual mode and select some lines, I do not find these weird characters

PtravisCS:
I'm on Ubuntu 22.04.3, nvim v0.9.1 and am getting the same issue. In my case, the extra virtual character is always a repeat of the last character from the line it appears on.

I have noticed that when I do visual line (V) I sometimes see a virtual copy of every character in the line, but shifted 1 space to the right. I don't know if this is an unrelated bug or not, but thought I should mention it.

This bug started around 2 - 3 weeks ago for me after an update to the vscode-neovim plugin.

![image](https://github.com/vscode-neovim/vscode-neovim/assets/21698432/c69c6df4-63b0-4905-8c45-10dd6529d114)

Hopefully this proves helpful.

huangzl97:
Looks like they recently released a major version [v0.4.0](https://github.com/vscode-neovim/vscode-neovim/releases/tag/v0.4.0) which introduce several breaking changes and features related to cursor.

I tried rollback to 0.3.2 and everything works as usual. So i decided to use 0.3.2 until 0.4.x is stable.

Tesohh:
getting this issue too on macOS.
<img width="338" alt="Screenshot 2023-08-10 alle 23 37 16" src="https://github.com/vscode-neovim/vscode-neovim/assets/45905658/369528b8-6a6b-48cf-8898-3df92c7209ec">

Tesohh:
[...]

PR #1430, opened 2023-09-07
fix(highlights): fix highlights when emojis exist

xiyaowong:
Close: #1374 

https://github.com/vscode-neovim/vscode-neovim/issues/1374#issuecomment-1710047365


theol0403 (approved):
Awesome!

PR #1443, opened 2023-09-11
Implement external buffer document provider

### LINKED ISSUE #1432: Use VSCode document API for external buffers
I started a proof-of-concept for making external buffers use the [`TextDocumentContentProvider` API](https://code.visualstudio.com/api/references/vscode-api#TextDocumentContentProvider), which is what I assume this is referring to. I'm happy to work on a PR if that's the intended direction (and can discuss implementation details, design etc. on a new issue if that's the idea).

_Originally posted by @ian-h-chamberlain in https://github.com/vscode-neovim/vscode-neovim/issues/1262#issuecomment-1710272737_

ian-h-chamberlain:
Closes #1432 

This opens a document, with a proper buffer name and matches it to the appropriate neovim external buffer.

The document is "readonly" as far as VSCode is concerned, but it can still be edited by setting `modifiable`, so we still need to sync buffer changes whenever it updates. This can break highlighting, however. I'm not sure if there's a way to force `nomodifiable` permanently, but it is already set for all external documents by default

Some things I haven't tested yet (would be grateful if others who use these features could test for their use cases), as mentioned in #1432 :
- I think `:marks` shows up in the output panel so is unaffected by this change
- `q:`, `q/` etc I couldn't get to work even without my changes. Are these expected to be functional?
- lazy.nvim; I didn't get around to trying since I don't use it normally. I can try to test this though

----

Also ensure the buffer scrolls so the selection is visible. This makes `:help` scroll to the help topic immediately, much like neovim does natively.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -590,18 +590,25 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
         }
     };
 
-    private receivedBufferEvent = async (
+    private receivedBufferEvent = (
         buffer: Buffer,
         tick: number,
         firstLine: number,
         lastLine: number,
         linedata: string[],
         more: boolean,
-    ): Promise<void> => {
+    ): void => {
         this.onBufferEvent && this.onBufferEvent(buffer.id, tick, firstLine, lastLine, linedata, more);
-        const uri = this.buildExternalBufferUri(await buffer.name, buffer.id);
-        this.logger.debug(`${LOG_PREFIX}: received buffer event for ${uri}`);
-        this.bufferProvider.documentDidChange.fire(uri);
+        // Ensure the receivedBufferEvent callback finishes before we fire
+        // the event notifying the doc provider of any changes
+        (async () => {
+            const uri = this.buildExternalBufferUri(await buffer.name, buffer.id);
+            this.logger.debug(`${LOG_PREFIX}: received buffer event for ${uri}`);
+            this.bufferProvider.documentDidChange.fire(uri);
+            return uri;
+        })().then(undefined, (e) => {
+            this.logger.error(`${LOG_PREFIX}: failed to notify document change: ${e}`);
+        });
     };
 
     /**
@@ -823,8 +830,6 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
     }
 }
 
-type BufferChangeCallback = BufferManager["receivedBufferEvent"];
-
 /**
  * Implements the VSCode document provider API for external buffers from neovim.
  */
@@ -839,7 +844,7 @@ class BufferProvider implements TextDocumentContentProvider {
     public constructor(
         private logger: Logger,
         private client: NeovimClient,
-        private receivedBufferEvent: BufferChangeCallback,
+        private receivedBufferEvent: BufferManager["receivedBufferEvent"],
     ) {}
 
     async provideTextDocumentContent(uri: Uri, token: CancellationToken): Promise<string | undefined> {
@@ -861,10 +866,7 @@ class BufferProvider implements TextDocumentContentProvider {
             return;
         }
 
-        buf.listen("lines", (...args: Parameters<BufferChangeCallback>) => {
-            this.receivedBufferEvent(...args);
-            this.documentDidChange.fire(uri);
-        });
+        buf.listen("lines", this.receivedBufferEvent);
         await buf[ATTACH](true);
 
         return lines.join("\n");
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -737,7 +737,9 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
     }
 
     private buildExternalBufferUri(name: string, id: number): Uri {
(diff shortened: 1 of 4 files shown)


xiyaowong:
> but it can still be edited by setting modifiable

Unless user intentionally modifies the buffer, it remains unchanged in most cases. 

> :marks

By “marks,” I mean jumping across files, but it’s not directly related to this pull request. 

> lazy.nvim

lazy.nvim is just an example, the purpose is to support floating windows.

However, the above points should be completed in different PRs

justinmk:
> I'm not sure if there's a way to force `nomodifiable` permanently,

Not really. And it's generally a mistake anyways. If plugins or users want to mess around, let them.

src/buffer_manager.ts
```diff
@@ -808,3 +820,51 @@ export class BufferManager implements Disposable, NeovimRedrawProcessable, Neovi
         }
     }
 }
+
+type BufferChangeCallback = BufferManager["receivedBufferEvent"];
+
+/**
+ * Implements the VSCode document provider API for external buffers from neovim.
+ */
+class ExternalBufferProvider implements TextDocumentContentProvider {
```
justinmk, line 829: nit: "External" is over-explaining isn't it? Is there any other kind of "buffer provider"? 

```suggestion
class BufferProvider implements TextDocumentContentProvider {
```

src/buffer_manager.ts
```diff
@@ -37,6 +40,8 @@ const LOG_PREFIX = "BufferManager";
 
 const BUFFER_NAME_PREFIX = "__vscode_neovim__-";
 
+const BUFFER_SCHEME = "vscode-neovim";
```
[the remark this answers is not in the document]
ian-h-chamberlain: Yeah, I didn't look in too much detail at how it's used, since I wanted to focus on external buffers for this PR, but I think using VScode URI scheme is the more idiomatic way to work when possible.

Right now, the neovim buffer names have a couple possible formats, as far as I can tell:
* `__vscode_neovim__-file:///path/to/file`
* `__vscode_neovim__-untitled:Untitled-1`
* `__vscode_neovim__-output:extension-output-asvetliakov.vscode-neovim-%231-vscode-neovim`
* Just `file` (e.g. `motion.txt`) for external buffers (also always unlisted, I think)

That is just the buffer name in Neovim though, and gets some translation when brought into VSCode. It seems a bit convoluted how everything is wired up, but I think if a full custom editor was implemented it might make sense to have everything under the `vscode-neovim:` scheme. For now, there is still a need for the `file://`, `output:` etc. schemes to correlate from neovim -> VSCode.

I think some bigger refactors are probably needed, and if that results in a 1-to-1 translation from neovim buffer name <-> VSCode URI it will probably simplify a lot of the buffer manager logic.

src/buffer_manager.ts
ian-h-chamberlain: Not currently, but there could be other `CustomTextEditorProvider`s or similar for some of the other use cases mentioned. Since this provider is only used for "external buffers" coming directly from Neovim I thought it might be helpful to name it as such.

If you think it's overly repetitive I can definitely rename though.

justinmk: > but there could be other `CustomTextEditorProvider`s or similar for some of the other use cases mentioned.

When that happens, then verbose names can be used. But until then there's no need for a verbose name.

theol0403:
Take a look at the failing tests when you have a chance, the incsearch one isn't your fault, but the others are.

theol0403:
Also take a look at the URI errors in the logs, that's new.

ian-h-chamberlain:
Looks like Windows is still failing, probably due to some other file path issue that I didn't think of. Unfortunately I don't have a Windows machine to test on, so it might be tricky for me to iterate locally, but I'll keep looking into it and try to come up with a fix.

xiyaowong:
I’m also not sure why this test case keeps failing. I had no issues when testing it locally on Windows.

ian-h-chamberlain:
I remembered I actually did have a Windows install lying around, so after a bunch of trial and error I realized it had to do with `async` timing on the buffer event notification. I think it should pass now, but CI will tell.

Also, while testing locally I noticed something interesting - I'm not sure if it's true on only Windows, but some of the tests fail (and have comments indicating this) if the test runner is in the background. I had a lot better success by keeping it in the foreground while running test, maybe because of how Windows scheduling works or something?

theol0403 (approved):
Works for me. May as well address other kinds of external buffers in another PR.

theol0403:
Actually, I just noticed this in my output...
image

ian-h-chamberlain:
@theol0403 was that just from running a `:help` command? I noticed that when the help buffer first opens, it starts off empty and with no name, so the first attempt to open it results in an exception being thrown by `workspace.openTextDocument(uri)`. I think you'd normally see the `Skipping empty external buffer` log right before that...

I guess to match the previous logic, we could just wrap that in a `try`/`catch` and `return`, instead of letting the exception bubble up. I actually thought I had made that change already but I must have lost it while debugging the Windows stuff. I can push another change to handle "missing" documents like that as well.

theol0403:
Yeah it was from the help buffer.

PR #1451, opened 2023-09-15
fix(cursor, visual): fix sync mouse selection

### LINKED ISSUE #1450: bug: Selecting text with the mouse may enter visual line mode instead of the regular visual mode

### Neovim version (nvim -v)

NVIM v0.9.2

### Operating system/version

macOS 13.1

### Describe the bug

Selecting text with the mouse may enter visual line mode instead of the regular visual mode if you used visual line mode before. This causes what you selected with your mouse to be different from what's actually selected.

The same thing happens with visual block mode as well.

### Steps To Reproduce

1. Open a file with some text.
2. Press `V` then `<Esc>` to enter and exit visual line mode.
3. Use the mouse to select only a part of a line.
4. Nvim enters visual line mode so it selects the entire line instead of what you actually selected.

### Expected Behavior

Using the mouse should enter the regular visual mode regardless of what mode you've used before. That's the behavior of the original Nvim as well.

xiyaowong:
Fixes #1450

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -421,10 +421,10 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
         this.logger.debug(
             `${LOG_PREFIX}: Starting visual mode from: [${anchor.line}, ${anchor.character}] to [${active.line}, ${active.character}]`,
         );
-        await this.client.call("visualmode", [1]);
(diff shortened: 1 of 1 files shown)


xiyaowong:
They are same, but nvim_buf_set_mark is more stable.

theol0403 (approved):

PR #1455, opened 2023-09-17
feat: warn on empty `vscode-neovim.send` args

### LINKED ISSUE #1209: Warn on empty vscode-neovim.send args
In my vim configuration, I have `Ctrl+V` mapped to paste from the clipboard, so I use `Ctrl+Q` as an alternate keymapping to enter visual block mode. I attempted to configure my VSCode keybindings to remap `Ctrl+Q` from its normal (quit) function and send it to neovim instead by adding the following to my `keybindings.json`:

```json
    {
        "key": "ctrl+q",
        "command": "-workbench.action.quit"
    },
    {
        "key": "ctrl+q",
        "command": "vscode-neovim.send",
        "when": "editorTextFocus && neovim.mode == 'normal'"
    }
```

After doing so, I would get strange behavior in the editor after pressing `Ctrl+Q`, certainly not entering visual block mode. Using neovim's macro record feature, I was able to see that upon pressing `Ctrl+Q`, the key sequence that was actually getting passed to neovim was `undefined`. It makes sense that this would turn into a mess of successive commands and give strange behavior!

After enabling logging in neovim, I confirmed that the `key` argument to the `onSendCommand()` handler in `typing_manager.ts` is `undefined` instead of the key that was pressed. I therefore was able to work around the problem by explicitly specifying the key as the argument to the send command in my key bindings configuration:

```json
    {
        "key": "ctrl+q",
        "args": "<C-q>",
        "command": "vscode-neovim.send",
        "when": "editorTextFocus && neovim.mode == 'normal'"
    }
```

But it seems like this should work even without the `args` field.

xiyaowong:
Resolve #1209


theol0403:
I believe escape/enable/disable/toggle all don't require a parameter.

theol0403 (approved):

PR #1467, opened 2023-09-20
feat(command line): add a configurable completion timer delay

### LINKED ISSUE #1105: Configure Wildmenu completion delay
Currently, the wild menu delay is set to `1.5s`; it would be nice to make this setting configurable.

dyl10s:
This resolves #1105

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,12 @@
                     "title": "Neovim width",
                     "description": "Neovim viewport width. Increase if you're working with long lines and have horizontal scroll problems. Larger values will affect neovim performance"
                 },
+                "vscode-neovim.completionDelay": {
+                    "type": "number",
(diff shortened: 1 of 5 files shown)


theol0403:
Thanks for the PR! Can you modify the settings flow to match the rest of the extension, like https://github.com/vscode-neovim/vscode-neovim/blob/master/src/extension.ts#L20?

dyl10s:
> Thanks for the PR! Can you modify the settings flow to match the rest of the extension, like https://github.com/vscode-neovim/vscode-neovim/blob/master/src/extension.ts#L20?

Refactored it, let me know if there are any other standards that I am not following here that you would like me to adjust!

theol0403 (approved):

PR #1479, opened 2023-10-01
fix(cursor): update position after exiting insert mode

archilkarchava:
This PR fixes the issue where the cursor position is not updated after exiting the insert mode in the same cursor position it was entered. Previously, this would result in the incorrect cursor position.

### Before

https://github.com/vscode-neovim/vscode-neovim/assets/16840190/628d26af-0b67-4819-a8a7-8e4272557903

### After

https://github.com/vscode-neovim/vscode-neovim/assets/16840190/f7a24737-0a0c-4869-971a-8b682d08e988

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
             "devDependencies": {
                 "@types/glob": "^7.1.3",
                 "@types/lodash-es": "^4.17.9",
-                "@types/mocha": "^10.0.1",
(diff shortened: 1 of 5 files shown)


theol0403:
Thanks, looks like it broke some tests though.

archilkarchava:
@theol0403, I've fixed the tests

theol0403 (approved):

Issue #97, opened 2019-12-16
VSCode Multiple cursors


mizagog:
Not a bug realy, but is there plans to do multiple cursours with VS's cursours? it is extremely useful to create cursors for occurrences and then be able to do vim stuff for each cursor.

avstudio:
Yeah, I'm looking the way how to select in visual, and then with Ctrl+n add to selection other occurrences like in vim-multiple-cursors. Is it feasible right now? Btw. great work.

asvetliakov:
not yet, but definitely something i will look into. For now they're work in insert mode. In meantime```cgn``` (```.``` dot repeat from ```vim-repeat``` doesn't work though yet) or ```dgn``` may work for you

mizagog:
I will check it out

pumpkinlink:
What I did was to remove the Ctrl+d keybinding for "outdent line", now I can use VSCode's native keybinding for "add selection to next find match" (only on insert mode of course)

brentluna:
> What I did was to remove the Ctrl+d keybinding for "outdent line", now I can use VSCode's native keybinding for "add selection to next find match" (only on insert mode of course)
@pumpkinlink where did you make this change? I saw the setting is in a package.json in the extension code, but don't know how i would access that.

echaya:
wish we could support 'mg979/vim-visual-multi' natively but may need some 'vscode-easymotion' fork

yousefamar:
> In meantime `cgn` or `dgn` may work for you

@asvetliakov could you explain what you mean by this? Thanks!

theol0403:
> > In meantime `cgn` or `dgn` may work for you
> 
> @asvetliakov could you explain what you mean by this? Thanks!

http://vimcasts.org/episodes/operating-on-search-matches-using-gn/

pencilcheck:
> looking forward to the enhancement. For anyone interested in using `vim-visual-multi` as a temp solution, please check out: [#409 (comment)](https://github.com/vscode-neovim/vscode-neovim/issues/409#issuecomment-830796423)

Thanks! still buggy when trying to edit selections but at least C-n seems to select multiple regions now

AndroAmater:
Using vim-visual-multi and editing the selection doesn't give live feedback and in some cases doesn't work at all for me. 

It would be great if we could have this! It's probably the only vscode feature that I'm missing and I miss it quite a lot tbh. 

[...]

PR #1490, opened 2023-10-05
feat!: better integration of vscode multiple cursors

xiyaowong:
TODO
#97

### DIFF
--- a/runtime/lua/vscode-neovim/multi-cursors.lua
+++ b/runtime/lua/vscode-neovim/multi-cursors.lua
@@ -127,6 +127,14 @@ local function add_cursor(cursor)
   end)
 end
 
+---@param cursor Cursor
+local function goto_cursor(cursor)
+  return api.nvim_win_set_cursor(0, {
+    cursor.range.start.line + 1,
+    cursor.range.start.character,
+  })
+end
+
 local function reset()
   bufnr = api.nvim_get_current_buf()
   cursors = {}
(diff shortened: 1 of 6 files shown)


runtime/lua/vscode-neovim/multi-cursors.lua
```diff
@@ -0,0 +1,314 @@
...
+
+---@class Cursor
+---@field range LspRange
+---@field extmarks number[] cursor hilight and range(selection) highlight
+
+local CREATE_CURSOR_DEFER_TIME = 10
+local START_DEFER_TIME = 40
+local NOTIFY_DEFER_TIME = 20
+
+local ns = api.nvim_create_namespace("--- vscode multiple cursors ---")
```
justinmk, line 20: nit: strange to use punctuation in a namespace name.
```suggestion
local ns = api.nvim_create_namespace("vscode.multicursor")
```

xiyaowong:
Ready for review, but not ready for merge

theol0403:
Btw @xiyaowong if you want to set up GitHub sponsors I'd be happy to add you to the list for the project.

xiyaowong:
@theol0403 Thank you! Not for now, after all, I have only recently started contributing to the project.

xiyaowong:
Lower priority of this PR, convert to draft for now

theol0403:
@xiyaowong btw I have one week right now with no classes so I can be more involved this week and maybe even make PRs. I agree this feature is lower priority, but in general, if you want to work with me on PRs now is a good time.

xiyaowong:
Great!

xiyaowong:
@theol0403 This PR currently has the available functions, and you can test it.

https://github.com/vscode-neovim/vscode-neovim/assets/47070852/af31a3a5-dbef-4542-8955-4f45b58ba0d9

xiyaowong:
By using flash.nvim, it is convenient to create cursors. This should be very practical.

cursors

xiyaowong:
It's best as a built-in plugin. The current built-in multi-cursors is too rudimentary. And this PR does not change the current usage. However, a separate plugin is indeed more convenient to update.

theol0403:
This PR is entirely contained within neovim, so no disadvantage to having it external. Creates maintanance burden for extension. 

Also, this integrates with external plugins (like flash). Ideally, VSCode neovim has no notion of the external ecosystem, it should be pure and vanilla. 

This seems like a perfect usecase of an extension making use of the new API.

xiyaowong:
Should it be an "official" repo or my personal repo? Considering that this can be used as a use case of the new api and solve the existing issue, I prefer to be an "official" plugin. Users can improve this plugin together or create a better plugin.

theol0403:
> Should it be an "official" repo or my personal repo? Considering that this can be used as a use case of the new api and solve the existing issue, I prefer to be an "official" plugin. Users can improve this plugin together or create a better plugin.

Official is good. It can be under vscode-neovim.

PR #1502, opened 2023-10-07
chore(options): no need to force setting syntax

xiyaowong:


runtime/lua/vscode-neovim/defaults.lua
```diff
@@ -2,6 +2,7 @@
 local M = {}
 
 function M.setup()
+  vim.opt.syntax = "on"
```
theol0403, line 5: I think this still has to be syntax("on")

xiyaowong: Yes, I just realized something is wrong:)

theol0403 (approved):

Issue #1475, opened 2023-09-27
feature: option to disable the synchronization of mouse selections


BTBMan:
### Did you check the docs?

- [X] I have read all the vscode-neovim docs
- [X] I understand the fundamental limitations of the plugin and believe this request to be reasonable

### Is your feature request related to a problem? Please describe.

After v0.3.2, It syncs visual selections with VSCode and synchronizes mouse selections. but it's not very comfortable to use. It has some problems sometimes. E.g. The cursor position syncs very slowly when selecting using the mouse, The visual selection is not correct after selecting.

In the terminal, It doesn't synchronize mouse selections by default.

I know the VSCodeVim has this feature, And that's one of the reasons I didn't use it. It's hard to use for me.

### Describe the solution you'd like

I hope to add a VSCode option to switch the synchronization of mouse selections.

### Describe alternatives you've considered

I don't have.

xiyaowong:
> The visual selection is not correct after selecting.

Actually, this is caused by a synchronization problem with the viewport.

> In the terminal, It doesn't synchronize mouse selections by default.

As far as I know, a terminal is not a text editor and is not handled by vscode-neovim at all.

> I hope to add a VSCode option to switch the synchronization of mouse selections.

It is possible to add that option, and I also think that synchronizing the mouse selection is not very useful.

TODO:

- option to toggle mouse selection
- option to change the debounce time for synchronizing the mouse selection

justinmk:
Ideally we should avoid adding an option and just make a decision about the best builtin behavior.

xiyaowong:
If the user wants to disable a certain feature, it seems that the only option is to add a new option.

theol0403:
> Ideally we should avoid adding an option and just make a decision about the best builtin behavior.

Yeah, this was my original approach. I thought that since VSCode was a graphical program to start with, mouse support should continue to be first-class. Ignoring the mouse seemed to me like something only desirable in a TUI.

Additionally, the code that syncs the mouse, isn't actually specific to the mouse at all. Instead, it syncs vscode selections to nvim.

The way the sync works is that whenever a vscode selection is updated, it waits a bit for debouncing, then compares with the exisitng visual mode in nvim, and sends nvim into visual mode if there is a mismatch.

If people really want to ignore the mouse, you'd have to *add* logic that ignores the vscode selection changed event when the source is from a mouse and the selection range > 1.

Issue #1409, opened 2023-09-01
bug: Handling of 'o'/'O' broken going from 0.3.2 to 0.4.X (cursor position synchronization)


scienceplease:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.9.1

### Operating system/version

Fedora Linux 37

### Describe the bug

Bug first appeared in 0.4.0 and persists in all 0.4.X versions.

When inserting a new line before/after the current line with 'O'/'o', the displayed cursor position (in VSCode) and actual cursor position (in neovim) become out of sync. For example, when a newline is inserted with 'o', a new line is _always visually_ displayed by VSCode and the cursor _appears_ on the new line, but performing an action like 'dd' (delete current line) _sometimes_ effects the line above the displayed cursor position.

The nondeterministic/timing-based nature of the bug hints towards something asynchronous/event loop related where VSCode is inserting a new line but Neovim still thinks the cursor is on the prior line.

What changed with respect to the implementation of 'O'/'o' from 0.3.2 to 0.4.0? Was something that was previously synchronous made asynchronous? Is there some implicit assumption being made about the completion order of two or more asynchronous actions?

### Steps To Reproduce

After lots of experimenting I found a repro that I can perform with a fairly high success rate on v0.4.3. The faster you can complete the sequence below the higher probability of the bug:

Start with line of text with an empty line above and below it.

1) In NORMAL mode, while on the line of text, insert a new line above 'O' and then immediately insert another new line below 'o' your current position (this effectively inserts two empty lines above your line of text).

2) Quickly move to the empty line below the line of text 'jj' and then delete it with 'dd'. The line with text above your cursor will be deleted instead of the empty line that your cursor was on.

### Expected Behavior

 Neovim's cursor position matches VSCode after 'O'/'o'.

theol0403:
Related: https://github.com/vscode-neovim/vscode-neovim/issues/1355

I don't quite understand the repro steps, if you use o/O, you are in insert mode, so how can you use dd?

theol0403:
And yeah, the cursor sync was refactored to support visual mode, but I tried to be careful about bugs like this.

scienceplease:
> Related: #1355
> 
[...]

PR #1504, opened 2023-10-08
feat: add options to disable or change selection debounce time

xiyaowong:
new settings

- `vscode-neovim.normalSelectionDebounceTime`
- `vscode-neovim.mouseSelectionDebounceTime`

Resolve: #1475

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,6 +199,31 @@
                             "backgroundColor": "theme.editor.selectionBackground"
                         }
                     }
+                },
+                "vscode-neovim.normalSelectionDebounceTime": {
+                    "type": "integer",
+                    "minimum": 20,
+                    "maximum": 100,
+                    "default": 50,
+                    "description": "Debounce time used to synchronize selection from VSCode to Nvim, typically referring to cursor movement"
+                },
+                "vscode-neovim.mouseSelectionDebounceTime": {
+                    "default": 100,
+                    "enum": [
+                        0,
+                        50,
+                        100,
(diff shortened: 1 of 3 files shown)


theol0403 (approved):

theol0403:
@xiyaowong out of curiosity, does disabling the mouse also disable cursor move on mouse click? I have a feeling that might still be nice to have even if selections don't enter visual mode.

xiyaowong:
I did a simple test at that time and didn’t find any issues.

justinmk:
To avoid the need for 2 options this could instead be a factor or multiple, like 1x/1.5x/2x/3x. Or from a user's perspective it could be a "dial" or slider, i.e. a range from 1 to 100. Which then would be converted by vscode-neovim to appropriate values internally. No need for users to tweak the exact millisecond values.

xiyaowong:
1. These two options are completely unrelated.

2. At present, I tend to solve the problem now and improve the document later.

3. There is now a limited range of values for these two options.

justinmk:
How are they unrelated? They both control the sync window between client/server. The fact that they control different internals is an implementation detail. When would a user adjust one of these values but not the other?

xiyaowong:
About `normal selection`: #1403, #1409
About `mouse selection`: #1475

The current values range and the candidates are both good. Most users don’t even need to change them, but if they do, they will naturally discover these configuration options. Like VSCode, it has a lot of hidden configurations.

Issue #714, opened 2021-07-29
Edamagit support


jtpavlock:
Hello, and thank you for the great extension!

[Edamagit ](https://marketplace.visualstudio.com/items?itemName=kahole.magit)is an extension to mimic the functionality of magit in emacs. Currently, they have [some tips for working with the Vim extension](https://github.com/kahole/edamagit#vim-support-vscodevim), and I was wondering if similar settings were possible with neovim?

Fraznist:
I think I found a (very tedious) solution:
`vscode-neovim` introduces a lot of `Neovim: send key` keybindings, which aren't deactivated when the user is in the `edamagit` window. Neovim bindings end up taking precedence, as a result edamagit bindings do not activate. Adding `editorLangId != 'magit'` to the `when` expressions of `Neovim: send key` keybinds lets me use the edamagit keybindings.

For instance, `<Tab>` wasn't folding or unfolding any of my changes in the magit status page. Adding the following to `keybindings.json` restores the `<Tab>` functionality to edamagit:
```json
{
  "key": "tab",
  "command": "extension.vim_tab",
  "when": "editorFocus && vim.active && !inDebugRepl && vim.mode != 'Insert' && editorLangId != 'magit'"
}
```
Downside to this approach is we would probably need to add all `vscode-neovim` keybinds that `edamagit` uses manually.

giulioungaretti:
Maybe it's me doing something wrong :D but the above from @Fraznist  do not seem to work. 
For now I went:
 {
      "command": "-vscode-neovim.send",
      "key": "tab"
  }
But it's a rather aggressive solution :D

robertfw:
I also couldn't get the solution from @Fraznist to work,the solution from  @giulioungaretti  did do the job, though I haven't used it enough to see if anything else has broken.

Has anyone come up with an ideal solution for this?

xiyaowong:
@robertfw We can provide a configuration like `vscode-neovim.keybinds.editorLangIdBlacklist` and then add `editorLangId not in neovim.editorIdBlacklist` to the when expressions of all keybinds.
https://code.visualstudio.com/api/references/when-clause-contexts#in-and-not-in-conditional-operators

trkoch:
Please try to add `magit` to `editorLangIdExclusions`. See https://github.com/vscode-neovim/vscode-neovim#disable-keybindings.

> To disable keybindings defined by this extension in certain filetypes, you can use the `editorLangIdExclusions`
[...]

PR #1508, opened 2023-10-09
refactor: use scripts to maintain keybindings

xiyaowong:
Beneficial for maintenance. There are too many key bindings, and there will be more in the future. It’s a bit difficult to manually maintain them, especially since there is a lot of duplicate content.


theol0403:
Ooooh I was thinking of doing this years ago but never got around to it/knew if it was possible, you are a lifesaver 🙏 .

However, I think we should take our time by optimizing this.
- since the last time package.json was touched, parenthesis were added to when clauses, which can simplify some logic
- most keybindings have an identical structure, and should be algorithmically generated
- for example, all the Ctrl-R bindings could really be created with a few lines of code 
- for example, we could simplify most of the mode-specific bindings into “list of shortcuts we want” -> keybinds
- I don’t think this is the best approach, but we could have the best configurability and power if we moved the keybindings from package.json into creating them dynamically at runtime. I think there are too many cons with this approach though.

xiyaowong:
@theol0403  Yes, my plan is exactly that, but I don't intend to add those logics in this PR. It would make it difficult to trace back in case of issues.

xiyaowong:
My current thoughts are:
1. Automatically generate the same "when" clause.
2. Predefine all the Ctrl+* shortcuts, and let users enable specific shortcuts through configuration, for example, `vscode-neovim.keybinds.ctrlKeysInNormal: ["a", "b", "<Up>"]`.
3. Additionally, include functionality like `vscode-neovim.keybinds.editorLangIdBlacklist` mentioned in https://github.com/vscode-neovim/vscode-neovim/issues/714#issuecomment-1728614617

xiyaowong:
@theol0403 Another idea I have is to change `vscode-neovim.*` to `neovim.*`. I'm curious to know what you think about it.

theol0403:
Awesome, that sounds great. 

For this PR, is there a reason why some of the bindings are rearranged, other than a product of the auto generation? I ask because the order does matter, and I'll have to take a closer look to make sure nothing is broken. 

Also, a useful bit of information is that the bindings are evaluated bottom to top :)

theol0403:
> @theol0403 Another idea I have is to change `vscode-neovim.*` to `neovim.*`. I'm curious to know what you think about it.

Yes that's fine, but not sure if it's worth the breaking change.

xiyaowong:
Currently, there are only a few configuration items, and the impact on users is not significant.

theol0403 (approved):

xiyaowong:
> Also, a useful bit of information is that the bindings are evaluated bottom to top :)

Yes, I know this. I rearranged them based on my own thoughts. The number at the beginning of the file name represents the priority.

scripts/keybindings/5 widgets.cjs
justinmk, line 1: Why is there whitespace in the filename?

package.json
```diff
@@ -1286,25 +1296,16 @@
...
         ]
     },
     "scripts": {
         "lint": "eslint --ext .ts src",
         "format": "prettier --ignore-path .gitignore --ignore-path .prettierignore --write .",
+        "keybind": "node ./scripts/keybindings/index.cjs && prettier --write ./package.json",
         "test": "node ./out/test/runTest.js",
         "pretest": "npm run test-compile",
-        "vscode:prepublish": "webpack --mode production",
+        "vscode:prepublish": "npm run keybind && webpack --mode production",
```
justinmk, line 1308: Could we commit the autogenerated result, so that it's visible in the git tree? That makes things more obvious/explicit.

xiyaowong: > Or developers need to run keybind manually?

Yes. 

I added it to prepublish for convenience.

justinmk: That needs to be documented somewhere, it's probably time we added a CONTRIBUTING.md.

But ideally it shouldn't be needed at all, instead the generated result should be committed to the repo so that the repo state matches what would be published.

theol0403: I'm  actually in the middle of writing a CONTRIBUTING.md!

PR #1516, opened 2023-10-09
docs(contributing): add CONTRIBUTING.md

theol0403:
@justinmk @xiyaowong 

This is a great opportunity for us to write down everything we know about this extension so that we reduce the barrier of entry for new contributors. 

I am open to any additions or changes to what I've written so far.

Additionally, @xiyaowong since you've recently learned by yourself how this extension works, you might have some useful things to write down that would be hepful to newcomers. 

Feel free to add commits to this branch.


xiyaowong:
As a side note, due to my average English proficiency, I have spent a lot of time correcting English expressions, sometimes even longer than the time spent on writing the PR itself 🥲. This is also a reason why I rarely improve documentation at the moment.

theol0403:
@justinmk let me know if you can think of any process/nvim APIs I've missed and I'll merge

theol0403:
@xiyaowong if there is any documents, tools, or processes you use that would be helpful to write down, just tell them to me in a list and I can write about them

theol0403:
Maybe would be good for me to also include a bit about the debugging process/console logs, tests, and the vscode api.

trkoch:
Well written!

xiyaowong:
Maybe we can refer to other projects
https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md
https://github.com/microsoft/vscode/blob/main/CONTRIBUTING.md
https://github.com/gitkraken/vscode-gitlens/blob/main/CONTRIBUTING.md
etc.

It's too detailed now and focuses on explaining internal implementation details. Most of these contents are time-sensitive and are more suitable to be written as an article(how vscode-neovim works) or as documentation comments.

CONTRIBUTING.md
```diff
@@ -0,0 +1,271 @@
+# Contributing
+
+Thank you for your interest in contributing to the project! This project is the result of the passion and hard work of
+many contributors, and we appreciate your help! To get started, please read this document to familiarize yourself with
+the project.
+
+-   [Contributing](#contributing)
```
justinmk, line 7: github autogenerates a ToC so you don't need this
image

README.md
```diff
@@ -27,6 +27,8 @@ mode and editor commands, making the best use of both editors.
     -   [VSCode specific differences](#vscode-specific-differences)
```
justinmk, line 27: could remove this ToC. people need to learn to use github's autogenerated ToC

CONTRIBUTING.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Run Tests
+
+1. Open the repo in VSCode.
+2. Go to debug view and click `Extension Tests` (F5).
+3. To run individual tests, modify `grep: ".*"` in `src/test/suite/index.ts`.
+
+## Design Principles
+
+-   **Focus on removing code**. As it is, the project is already too large and complex, full of workarounds due to
```
justinmk, line 72: 💯 less is more.

CONTRIBUTING.md
```diff
@@ -0,0 +1,271 @@
...
+### Neovim UI Protocol
+
+VSCode-Neovim is built around neovim's ui protocol [:help ui](https://neovim.io/doc/user/ui.html). This protocol allows
+vscode to listen to events emitted by neovim, such as `mode_change`, `hl_attr_define`, and `win_viewport`. This API
+allows neovim to externalize the _rendering_ of various components of the editor, such as the cursor, highlights,
+buffers, cmdline, etc. When possible, this protocol should be used as the source of truth for the state of the editor.
+
+However, the UI protocol [is incomplete](https://github.com/neovim/neovim/issues/9421), and there are many holes in the
+API. For example, there is no way to get the type of visual mode or the current visual selection, `win_viewport` is not
+triggered on horizontal scrolling, there are no events for window splits and tabs, etc. As the nvim UI extension API is
```
justinmk, line 152: is there an upstream issue for the win_viewport horiz scrolling? https://github.com/neovim/neovim/issues?q=is%3Aissue+is%3Aopen+win_viewport

CONTRIBUTING.md
```diff
@@ -0,0 +1,271 @@
...
+type of visual mode, which is not available in the UI protocol. In that case, the autocommand should forward the event
+to vscode for custom handling.
+
+### Neovim Lua API
+
+To send commands to neovim, and to (worst-case scenario) ask it for additional information, the
+[neovim lua api](https://neovim.io/doc/user/api.html) can be used. As a last resort, the
+[builtin](https://neovim.io/doc/user/builtin.html) API can be used.
+
+Due to the number of RPC round trips required, the logic using the lua or vim APIs should be moved to a custom lua
```
justinmk, line 171:
```suggestion
To avoid RPC round trips, the logic using the lua or vim APIs should be moved to a custom lua
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,271 @@
...
+function running in nvim.
+
+## Managers
+
+VSCode-Neovim is structured using a manager pattern. Each manager is responsible for syncing a specific aspect of the
```
justinmk, line 176: should we mention the eventbus `.on` ? https://github.com/vscode-neovim/vscode-neovim/blob/c554c065a375f2bd32346aabdb6526ece8bb4189/src/eventBus.ts#L176

CONTRIBUTING.md
theol0403: @justinmk WinScrolled is currently used as a workaround, but the point here is to illustrate that there are gaps

theol0403:
> It's too detailed now and focuses on explaining internal implementation details. Most of these contents are time-sensitive and are more suitable to be written as an article(how vscode-neovim works) or as documentation comments.

I do agree, but I don't think it hurts. It's good to give people a map of how the extension is built without needing for them to dig through it. The implementation details should ideally be updated with notes as it evolves over time.

PR #1521, opened 2023-10-11
feat: Add option `statusLineSeparator`

xiyaowong:
Separator used by statusline

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,7 +227,7 @@
                 },
                 "vscode-neovim.statusLineSeparator": {
                     "type": "string",
-                    "default": " ",
+                    "default": "|",
                     "description": "Separator used by statusline"
                 }
             }
--- a/src/config.ts
+++ b/src/config.ts
@@ -143,7 +143,7 @@ export class Config implements Disposable {
         return this.cfg.get("logOutputToConsole", false);
     }
     get statusLineSeparator() {
-        return this.cfg.get("statusLineSeparator", " ");
+        return this.cfg.get("statusLineSeparator", "|");
     }
 
     get normalSelectionDebounceTime() {


package.json
```diff
@@ -224,6 +224,11 @@
                         "200ms"
                     ],
                     "description": "Debounce time for synchronizing mouse selection. Set to 0 to disable synchronization"
+                },
+                "vscode-neovim.statusLineSeparator": {
+                    "type": "string",
+                    "default": " ",
```
justinmk, line 230: Default should be `|`

theol0403: Yeah, we may as well pick a good default

theol0403: @xiyaowong The problem is that a big focus of this extension is to reduce the lines of code and simplify, because currently it is difficult to maintain. This means thinking critically of features that don't bring much value. In this case, having an option provides small value. 

If it would be better to simply have spaces for everyone, that would be fine. 

We can merge it, but we should avoid adding configuration options that make little impact.

xiyaowong: Many users have different preferences for UI, such as me. The maintenance difficulty is not directly related to the new features and the number of lines of code. I agree that our primary goal is to reduce the code complexity as much as possible, but it is by no means relying on reducing the code and avoiding new configuration items.

src/status_line.ts
```diff
@@ -1,31 +1,28 @@
...
         this._modeText.length && items.push(this._modeText);
         this._statusText.length && items.push(this._statusText);
         this._msgText.length && items.push(this._msgText);
-        if (items.length) {
-            this.statusBar.text = items.join(this._seperator);
-            this.statusBar.show();
-        } else {
-            this.statusBar.hide();
-        }
+        // Always show the status line
```
justinmk, line 24: +1 for this also. It was weird that the statusbar disappears.

xiyaowong: I think we need to adjust the priority. It looks strange right now. It should be displayed in the middle so that it doesn't cause other items to move around.

image

xiyaowong: Should use -1

PR #1522, opened 2023-10-11
refactor: add eventBus

xiyaowong:


src/cursor_manager.ts
```diff
@@ -137,51 +139,32 @@ export class CursorManager implements Disposable, NeovimRedrawProcessable, Neovi
         }
     }
 
-    public handleRedrawBatch(batch: [string, ...unknown[]][]): void {
```
theol0403, line 140: I suppose these can be private now

theol0403 (approved):
Typescript wizardry!

justinmk:
Nice. Can you explain the idea in the commit message, and possibly also in CONTRIBUTING.md or maybe an `architecture.md` doc. Or at least in `src/eventBus.ts`.

theol0403:
@justinmk I believe there is little documentation due to the effort it takes for them to write english, but I think this PR is fairly straightforward - instead of having each manager implement an interface for handling events, which gets called by the main controller, now each manager subscribes to an event broadcast.

src/eventBus.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export interface Event<T extends keyof EventsMapping = keyof EventsMapping> {
+    name: T;
+    data: EventsMapping[T];
+}
+
+export type EventBusData<T extends keyof EventsMapping> = EventsMapping[T];
+
+class EventBus implements Disposable {
+    private readonly emitter = new EventEmitter<Event>();
```
justinmk, line 113: Is the idea that all Nvim events are dispatched by this single `EventEmitter` ?

theol0403: Yeah I believe so. This will also let you write lua handlers for nvim events with the new api RFC

justinmk:
> but I think this PR is fairly straightforward 

PRs in isolation are often easily digestible, but for maintenance / other contributors it's worth having a basic high-level description somewhere, which can be augmented over time.

In this case, just mentioning this one-sentence explanation somewhere is much better than nothing:

> instead of having each manager implement an interface for handling events, which gets called by the main controller, now each manager subscribes to an event broadcast.

That explanation could live in the module itself as well as an "architecture" section in CONTRIBUTING.md. Doesn't need to be fancy, the main thing is just to get something started so we have a place to add notes along the way.

xiyaowong:
Merge?

src/eventBus.ts
```diff
@@ -0,0 +1,155 @@
...
+};
+
+export interface Event<T extends keyof EventsMapping = keyof EventsMapping> {
+    name: T;
+    data: EventsMapping[T];
+}
+
+export type EventBusData<T extends keyof EventsMapping> = EventsMapping[T];
+
+class EventBus implements Disposable {
```
justinmk, line 112: Here's what I was hoping for (and also a note in CONTRIBUTING.md):

```
/**
 * All Nvim events are dispatched by this single EventEmitter.
 * Components can subscribe to event broadcasts using `EventBus.on()`.
 */
```

PR #1525, opened 2023-10-12
fix: position statusBarItem in the middle

xiyaowong:
If you don’t like this change, how about adding a new configuration option? :(


justinmk (approved):
Looks fine to me. Definitely doesn't need to be configurable.

I'd call this a "fix" or "feat", not "chore". "chore" is pretty much never a useful prefix.

xiyaowong:
Oh, I didn’t pay attention to modifying the commit message. Should I force push?

Issue #1524, opened 2023-10-11
Text selection breaks the rendering and scrambles the selected text 


mzdravkov:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

NVIM v0.9.0-dev

### Operating system/version

Windows (also in WSL)

### Describe the bug

Hello!

When I select text in visual mode the selected text gets scrambled in a weird way. Here's a video, because I find it a bit hard to explain:

https://github.com/vscode-neovim/vscode-neovim/assets/1406355/48631559-19c5-4edc-addc-91b4a0b18b8e

The problem can still be reproduced after disabling all other extensions and removing my `init.lua` config.
I tried using different versions of the vscode-neovim extension and the issue appears in version `0.7.0`. In version `0.6.1` it works without issues.

I can also see that there are some errors in the vscode-neovim output window. Those appear with the loading of the extension and not when I try to select:

```
Error detected while processing pre-vimrc command line..script /mnt/c/Users/mihai/.vscode/extensions/asvetliakov.vscode-neovim-0.9.0/vim/vscode-neovim.vim:
line  116:
E5108: Error executing lua ...ode-neovim-0.9.0/runtime/lua/vscode-neovim/highlight.lua:73: attempt to call field 'nvim_exec2' (a nil value)
stack traceback:
	...ode-neovim-0.9.0/runtime/lua/vscode-neovim/highlight.lua:73: in function 'setup_syntax_groups'
	...ode-neovim-0.9.0/runtime/lua/vscode-neovim/highlight.lua:122: in main chunk
	[C]: in function 'require'
	...liakov.vscode-neovim-0.9.0/runtime/lua/vscode-neovim.lua:5: in main chunk
	[C]: in function 'require'
	[string ":lua"]:1: in main chunk
Error detected while processing command line..script /mnt/c/Users/mihai/.vscode/extensions/asvetliakov.vscode-neovim-0.9.0/runtime/lua/vscode-neovim/force-options.lua:
E5113: Error while calling lua chunk: vim/_meta.lua:0: no such option: 'statuscolumn'
stack traceback:
	[C]: in function 'nvim_get_option_info'
	vim/_meta.lua: in function '__index'
	vim/_meta.lua: in function ''
[...]

PR #1527, opened 2023-10-13
feat: improved the nvim version check

xiyaowong:


theol0403 (approved):

src/main_controller.ts
```diff
@@ -301,31 +301,51 @@ export class MainController implements vscode.Disposable {
...
+        const requirements = {
+            options: ["statuscolumn"],
+            functions: ["nvim_exec2"],
+        };
+        // check version
+        {
+            const [, info] = await this.client.apiInfo;
+            const { major, minor, patch } = info.version;
+            const currVersion = [major, minor, patch];
+            if (minVersion.some((v, i) => v > currVersion[i])) {
```
justinmk, line 316: This will reject version 0.10.1 if minVersion is 0.9.2

src/main_controller.ts
```diff
@@ -301,31 +301,51 @@ export class MainController implements vscode.Disposable {
...
+                } else if (expr.startsWith("*")) {
+                    missFunctions.push(expr.substring(1));
+                }
+            }
+        });
+        const errMsgs = [
+            "Your nvim does not support the following features. Please check and update your nvim to the [latest stable version](https://github.com/neovim/neovim/releases/tag/stable). ",
+        ];
+        if (missOptions.length) errMsgs.push("Miss options: " + missOptions.join(", ") + ". ");
+        if (missFunctions.length) errMsgs.push("Miss functions: " + missFunctions.join(", ") + ". ");
```
justinmk, line 346: s/Miss/missing

Not sure this message is helpful though. It's a roundabout way of saying "you need version X".

src/main_controller.ts
xiyaowong: My bad

src/main_controller.ts
xiyaowong: It should be useful and users using the dev version can understand this information.
For example, the 0.9.0-dev version used in #1524

justinmk: > For example, the 0.9.0-dev version used in #1524

But `0.9.0-dev` should fail a check against 0.9.0. That's defined by semver.

xiyaowong: What if -dev contains the features we need?

justinmk: Why is it worth having complicated code to support that edge case?

xiyaowong: make sense

PR #1528, opened 2023-10-13
refactor: use event bus for vscode-neovim events

xiyaowong:
Remove `NeovimExtensionRequestProcessable` in favor of `eventBus.on`

~~Not sure if it's better than before.~~

### DIFF
--- a/src/commands_controller.ts
+++ b/src/commands_controller.ts
@@ -1,10 +1,10 @@
-import vscode, { Disposable } from "vscode";
+import vscode, { Disposable, commands } from "vscode";
 
 import { config } from "./config";
+import { eventBus } from "./eventBus";
 import { MainController } from "./main_controller";
-import { NeovimExtensionRequestProcessable } from "./neovim_events_processable";
 
-export class CommandsController implements Disposable, NeovimExtensionRequestProcessable {
+export class CommandsController implements Disposable {
     private disposables: Disposable[] = [];
 
     private get client() {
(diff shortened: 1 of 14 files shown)


src/mode_manager.ts
```diff
@@ -57,6 +57,35 @@ export class ModeManager implements Disposable, NeovimExtensionRequestProcessabl
     private isRecording = false;
     private eventEmitter = new EventEmitter();
 
+    constructor() {
```
theol0403, line 60: Should be moved to `handleModeChanged`.

src/document_change_manager.ts
```diff
@@ -95,7 +95,13 @@ export class DocumentChangeManager implements Disposable, NeovimExtensionRequest
     public constructor(private main: MainController) {
         this.main.bufferManager.onBufferEvent = this.onNeovimChangeEvent;
         this.main.bufferManager.onBufferInit = this.onBufferInit;
-        this.disposables.push(workspace.onDidChangeTextDocument(this.onChangeTextDocument));
+        this.disposables.push(
+            workspace.onDidChangeTextDocument(this.onChangeTextDocument),
+            eventBus.on("insert-line", ([type]) => {
```
theol0403, line 100: `insert-line` is actually no longer used so we can delete it here and elsewhere. However, this might be better to have in a different document-change-manager cleanup PR (I've already cleaned it up in a few different PRs I was unable to merge :upside_down_face: )

vim/vscode-neovim.vim
```diff
@@ -23,24 +23,22 @@ function! VSCodeNotify(cmd, ...)
...
-    call rpcrequest(g:vscode_channel, s:vscodePluginEventName, a:cmd, a:000)
-endfunction
-
 function! VSCodeExtensionNotify(cmd, ...)
     call rpcnotify(g:vscode_channel, s:vscodePluginEventName, a:cmd, a:000)
 endfunction
 
 function! VSCodeCallRange(cmd, line1, line2, leaveSelection, ...) abort
-    call VSCodeExtensionCall('range-command', a:cmd, 'V', a:line1, a:line2, 1, 1, a:leaveSelection, a:000)
+    " FIXME: breaking
```
theol0403, line 31: fixme

theol0403 (approved):
Overall, I like this change, and there is a code reduction.

src/document_change_manager.ts
xiyaowong: I know this. It's very rough at the moment and needs to be refactored in several PRs.

vim/vscode-neovim.vim
xiyaowong: In theory, this is a breaking change, but in practice, it has no impact. It needs to be addressed and improved in the PR for redesigning the API

theol0403: We will need to maintain the vimscript API because many people's configs rely on it. I agree the internal implementation can be redesigned though.

xiyaowong: Yes, the old VimScript API is worth keeping. Mainly the internal implementation and Lua API need to be rewritten.

src/commands_controller.ts
```diff
@@ -13,19 +13,23 @@ export class CommandsController implements Disposable, NeovimExtensionRequestPro
...
+            commands.registerCommand("vscode-neovim.ctrl-b", () => this.scrollPage("page", "up")),
+            commands.registerCommand("vscode-neovim.ctrl-d", () => this.scrollPage("halfPage", "down")),
+            commands.registerCommand("vscode-neovim.ctrl-u", () => this.scrollPage("halfPage", "up")),
+            commands.registerCommand("vscode-neovim.ctrl-e", () => this.scrollLine("down")),
+            commands.registerCommand("vscode-neovim.ctrl-y", () => this.scrollLine("up")),
+            eventBus.on("reveal", ([at, updateCursor]) => this.revealLine(at, !!updateCursor)),
+            eventBus.on("move-cursor", ([to]) => this.goToLine(to)),
+            eventBus.on("scroll", ([by, to]) => this.scrollPage(by, to)),
+            eventBus.on("scroll-line", ([to]) => this.scrollLine(to)),
+            eventBus.on("insert-line", async ([type]) => {
```
justinmk, line 26: So the main change in this PR is to remove `NeovimExtensionRequestProcessable` in favor of `eventBus.on` ? It's helpful to mention that kind of key idea in the PR description / commit message.

src/eventBus.ts
```diff
@@ -110,6 +142,10 @@ export interface Event<T extends keyof EventsMapping = keyof EventsMapping> {
 export type EventBusData<T extends keyof EventsMapping> = EventsMapping[T];
 
 class EventBus implements Disposable {
+    /**
+     * All Nvim events are dispatched by this single EventEmitter.
+     * Components can subscribe to event broadcasts using `EventBus.on()`.
+     */
```
justinmk, line 148: Thanks :)

src/main_controller.ts
```diff
@@ -227,13 +218,7 @@ export class MainController implements vscode.Disposable {
...
         if (method === "vscode-neovim") {
             const [command, args] = events;
-            extensionCommandManagers.forEach((m) => {
-                try {
-                    m.handleExtensionRequest(command, args);
-                } catch (e) {
-                    logger.error(`${command} failed, args: ${JSON.stringify(args)} error: ${(e as Error).message}`);
-                }
-            });
+            eventBus.fire(command as any, args);
```
justinmk, line 221: Definitely looks like some actual simplifications were gained.

P.S. I don't see any logging in eventBus, should there be (e.g. on errors)?

justinmk (approved):

src/main_controller.ts
xiyaowong: I have also thought about this question. I think errors should be handled in the handler rather than the event bus.

Issue #434, opened 2020-11-23
[feature request] leave insert mode onFocusChange event


vdegenne:
It would be very useful to have a way to leave insert mode on `onFocusChange` (specifically `onWindowChange`).
When I code a web application I often quickly switch from window to window using alt+shift or similar shortcut to check the result and go back to visual studio code for fine tuning the page. Sometimes (a lot of time) I forget that I didn't press escape to leave insert mode before switching the window so I start to press `j` or any other navigation key but because I am still in insertion mode I have `jjjjjjjjjj` that is written.. and because I can't escape the mode and press `u` to undo because the last modification will also get lost, I have to manually remove the unwanted `jjjjjjj` character on the line which is double time a pain in the... hands.

Please if you could implement this option as soon as you can, I am sure many people will also like that.
Or maybe just telling me if you think of a quick adjustment in VSC to allow this effect.

Thanks

martin-braun:
I made the same suggestion over at VsVim when I first started learning Vim. Now returning from Neovim for some benefits of VSCode by using `vscode-neovim` I noticed this issue doesn't bother me anymore and here is why:

I got used to keyboard movements and thus I got used to ctrl+c before attempting the navigation to a different window/buffer. This is pretty doable if you take the time to remap VSCode bindings as well as add custom bindings to pass-through vim keymaps to VSCode via `VSCodeCall` or `VSCodeNotify`.

Still I want to say that this convenience would be useful for those that are still tied to work-flows with a mouse. Even in the optimal keymaps ... VSCode is sometimes more convenient using a mouse.

theol0403:
@martin-braun if you have any keybindings that you think are general enough to include in the extension (including vscode-specific bindings), feel free to PR!

martin-braun:
@theol0403 Well, my nvim-tree mappings are slightly customized (so are my VSCode equivalents) as are my terminal navigations as well, since I use tmux outside of VSCode for that with a different prefix as well. So my bindings are slightly off the norm, but it works the best for me.

What is probably useful in this project are my general window movements `^w ...` as those are original, but still missing by vscode-neovim, but sorely implementing those will be half-baked. You will have the ability to leave the file explorer, but you will not have the ability to enter it.

theol0403:
@martin-braun what about entering the file explorer by modifying the existing c-w bindings to move between ui panes (including explorer) instead of just editor windows? I believe there is a command that does that.

martin-braun:
[...]

PR #1532, opened 2023-10-14
feat!: Lua API

xiyaowong:


.eslintrc.js
```diff
@@ -54,6 +54,7 @@ module.exports = {
                 ],
                 "@typescript-eslint/explicit-function-return-type": "off",
                 "@typescript-eslint/no-non-null-assertion": "off",
+                "@typescript-eslint/no-explicit-any": "off",
```
xiyaowong, line 57: This rule is very inconvenient, developers can try to avoid using “any” as much as possible on their own.

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,170 @@
...
--- send command to vscode with range (line or char). [1, 1]-based.
-function M.call_range(command, line1, line2, leaveSelection, ...)
-  return M.call_extension("range-command", command, "V", line1, line2, 1, 1, leaveSelection, { ... })
-end
+------- Requests -------
 
-function M.notify_range(command, line1, line2, leaveSelection, ...)
-  return M.notify_extension("range-command", command, "V", line1, line2, 1, 1, leaveSelection, { ... })
-end
+local REQUEST_SATE = {
```
theol0403, line 5: State

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,170 @@
...
+---@param id number callback id
+---@param result any resutl
+---@param is_error boolean is error
+function M.invoke_callback(id, result, is_error)
+  vim.schedule(function()
+    local callback = REQUEST_SATE.callbacks[id]
+    REQUEST_SATE.callbacks[id] = nil
+    if callback then
+      if is_error then
+        callback(result, nil)
```
theol0403, line 26: I think it should be (res, err), to match js (resolve, reject).

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,170 @@
...
-
-  local top_line = vim.fn.line("w0")
-  local diff = top_line - vscode_topline
-
-  if diff ~= 0 and (vscode_topline > 0) then
-    vim.fn.winrestview({
-      topline = vscode_topline,
-    })
+  if opts.callback then
+    opts.callback = add_callback(opts.callback)
```
theol0403, line 64: Would it be better practice to separately store the id rather than overwriting the function?

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,170 @@
...
+  })
+  opts = opts or {}
+  vim.validate({
+    ["opts.callback"] = { opts.callback, "nil" },
+    ["opts.args"] = { opts.args, "t", true },
+    ["opts.range"] = { opts.range, "t", true },
+    ["opts.line_range"] = { opts.line_range, "t", true },
+    ["opts.leave_selection"] = { opts.leave_selection, "b", true },
+  })
+  opts._ = 1
```
theol0403, line 96: What does this do?

runtime/lua/vscode-neovim/autocmds.lua
```diff
@@ -0,0 +1,7 @@
+vim.api.nvim_create_autocmd("InsertLeave", {
```
theol0403, line 1: I think we should make every file a lua module. Require should not produce any action or side effects, it should only import the module, and setup should be called. This way in the future, we can support exposing data through the module that user code can import (without registering another auto command).

runtime/lua/vscode-neovim/default-options.lua
```diff
@@ -0,0 +1,13 @@
+vim.cmd.syntax("on")
```
theol0403, line 1: Same here, I don't think we want to remove the module pattern.

src/main_controller.ts
```diff
@@ -138,23 +139,15 @@ export class MainController implements vscode.Disposable {
...
 
-    private setClientInfo() {
-        readFile(path.posix.join(this.extensionPath, "package.json"))
-            .then((buffer) => {
-                const versionString = JSON.parse(buffer.toString()).version as string;
-                const [major, minor, patch] = [...versionString.split(".").map((n) => +n), 0, 0, 0];
-                this.client.setClientInfo("vscode-neovim", { major, minor, patch }, "embedder", {}, {});
-            })
-            .catch((err) => console.log(err));
+    public fireNvimEvent(event: string, ...args: VimValue[]): void {
```
theol0403, line 142: Where is this being used other than init?

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,170 @@
...
+  local args = { ... }
+  vim.schedule(function()
+    vim.tbl_map(function(cb)
+      cb(unpack(args))
+    end, EVENT_STATE[event] or {})
+  end)
 end
 
+------- VSCode settings integration -------
+function M.get_config() end
```
theol0403, line 167: Todo

theol0403 (commented):
Looks great!

runtime/lua/vscode-neovim/api.lua
xiyaowong: Writing..

src/main_controller.ts
xiyaowong: I struggled for a long time, not knowing where to put it.😂

runtime/lua/vscode-neovim/autocmds.lua
xiyaowong: Ok

runtime/lua/vscode-neovim/api.lua
xiyaowong: Empty table will be considered an array. I have modified it but haven’t pushed it yet.

runtime/lua/vscode-neovim/api.lua
xiyaowong: I hope users will pay attention to error handling, just like pcall.

runtime/lua/vscode-neovim/api.lua
xiyaowong: Functions cannot be passed in RPC

runtime/lua/vscode-neovim/api.lua
xiyaowong: done

runtime/lua/vscode-neovim.lua
```diff
@@ -1,30 +1,28 @@
...
+local numbers = require("vscode-neovim.numbers")
+local autocmds = require("vscode-neovim.autocmds")
 
-M.notify = api.notify
-M.call = api.call
-M.call_range = api.call_range
-M.notify_range = api.notify_range
-M.call_range_pos = api.call_range_pos
-M.notify_range_pos = api.notify_range_pos
+default_optons.setup()
```
theol0403, line 10: We still need to expose setup as part of this module. We don't want to call setup every time vscode-neovim is required. Setup should be once on init, like it was before.

vim/vscode-neovim.vim
```diff
@@ -110,5 +110,5 @@ augroup END
 
 
 lua << EOF
-require("vscode-neovim").setup()
```
theol0403, line 113: I think this is still needed

theol0403 (changes requested):

runtime/lua/vscode-neovim.lua
xiyaowong: The setup needs to be reserved for users to configure features.

Lua will not reload an already loaded module unless the user manually modifies `package.loaded`...

xiyaowong: Improved

runtime/lua/vscode-neovim.lua
```diff
@@ -1,30 +1,28 @@
...
+}
 
-vim.api.nvim_create_autocmd("InsertLeave", {
-  callback = function()
-    vim.fn.VSCodeNotify("hideSuggestWidget")
-    vim.fn.VSCodeNotify("closeParameterHints")
-    vim.fn.VSCodeNotify("editor.action.inlineSuggest.hide")
-  end,
-})
+_G._vscode = vscode
```
theol0403, line 26: I thought we agreed this is an anti-pattern :p

runtime/lua/vscode-neovim/api.lua
theol0403: Is that worth the strange ordering though? Most users will be writing simple keybindings, and don't really care about error handling.

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,205 @@
...
+  vim.validate({
+    name = { name, "string" },
+    opts = { opts, "table", true },
+  })
+  opts = opts or {}
+  vim.validate({
+    ["opts.callback"] = { opts.callback, "f", true },
+    ["opts.args"] = { opts.args, "t", true },
+    ["opts.range"] = { opts.range, "t", true },
+    ["opts.line_range"] = { opts.lien_range, "t", true },
```
theol0403, line 56: line_range, not lien_range

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,205 @@
...
+  end)
 end
 
-function M.notify_range_pos(command, line1, line2, pos1, pos2, leaveSelection, ...)
-  return M.notify_extension("range-command", command, "v", line1, line2, pos1, pos2, leaveSelection, { ... })
+---Run action, this function is asynchronous
+---@param name string The action name, generally a vscode command
+---@param opts? table Optional options table, all fields are optional
+---            - args: (table) Optional arguments for the action
+---            - range: (table) Specific range for the action. In visual mode, this parameter is generally not required.
```
theol0403, line 38: Before merge, we (I can) need to document this in the readme, especially an example to specify range. 

Also, I'm not totally sure about the range format. I think lsp range is useful, but in reality I'm not sure how often it will be used. The line_range example shows that. 

My suggestion:

```lua
range = {start: {int | {line: int, character: int}, end: {int | {line: int, character: int}} | [int | {line: int, character: int}, int | {line: int, character: int}]

range = {vim.fn.line("."), vim.fn.line("v")}
range = {{line=vim.fn.line("."), character=vim.fn.col(".")}, {line=vim.fn.line("v"), character=vim.fn.col("v")}}
--- still supports lsp range
```

runtime/lua/vscode-neovim/default-options.lua
```diff
@@ -17,4 +14,4 @@ function M.setup()
   vim.g.loaded_matchparen = 1
 end
 
-return M
```
theol0403, line 20: Why this pattern here, but not in filetype.lua?

src/main_controller.ts
```diff
@@ -7,7 +7,9 @@ import { attach, NeovimClient } from "neovim";
 import vscode from "vscode";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { transports as loggerTransports, createLogger as winstonCreateLogger } from "winston";
+import { VimValue } from "neovim/lib/types/VimValue";
```
theol0403, line 10: lint

src/main_controller.ts
```diff
@@ -20,10 +22,9 @@ import { HighlightManager } from "./highlight_manager";
 import { createLogger } from "./logger";
 import { ModeManager } from "./mode_manager";
 import { MultilineMessagesManager } from "./multiline_messages_manager";
-import { NeovimCommandProcessable } from "./neovim_events_processable";
 import { StatusLineManager } from "./status_line_manager";
 import { TypingManager } from "./typing_manager";
-import { findLastEvent } from "./utils";
+import { findLastEvent, wait } from "./utils";
```
theol0403, line 27: lint

theol0403 (changes requested):
I'm not sure about the hooks functionality. I think it is code we don't need. I had thought the `vscode.on()` was for subscribing to `eventBus`, which also is probably not useful.

`init` can be observed by `VimEnter` or something similar.

For https://github.com/vscode-neovim/vscode-neovim/issues/434, we should add specific code that calls `doau FocusGained` etc, and let users use autocommands. We should not reinvent any mechanisms we don't need to.

runtime/lua/vscode-neovim/default-options.lua
xiyaowong: They are same

theol0403: Then keep it the same, there is value in consistency so a reader knows what to expect.

runtime/lua/vscode-neovim.lua
xiyaowong: Easy to test, I will delete it.

runtime/lua/vscode-neovim/api.lua
xiyaowong: If you use simple keibindings, you won't use the callback parameter.

xiyaowong:
> I'm not sure about the hooks functionality. I think it is code we don't need. I had thought the `vscode.on()` was for subscribing to `eventBus`, which also is probably not useful.
> 
> `init` can be observed by `VimEnter` or something similar.
> 
> For #434, we should add specific code that calls `doau FocusGained` etc, and let users use autocommands. We should not reinvent any mechanisms we don't need to.

I don't think implicit doautocmd is better. Users can do it themselves.

Using on can handle bidirectional synchronization more conveniently, such as editor.options(&ts, &sw)).

This mechanism is useful and common. Many nvim plug-ins have some hooks

runtime/lua/vscode-neovim/api.lua
xiyaowong: We can provide some convenient functions to create range, in fact, we can use vim.lsp.util directly.

runtime/lua/vscode-neovim/api.lua
xiyaowong: If you use simple keibindings, you won't use the callback parameter.

xiyaowong:
Deleted comments about lint.

runtime/lua/vscode-neovim/api.lua
theol0403: Either we provide a flexible API to accept many different types of ranges, or we create external helpers. Not a mix of both. If we want to make helpers, then we should remove the `line_range` as it will be redundant.

What we need, either provided by the API or by helpers:
* lsp range
* convenient/concise char-wise range
* convenient/concise line-wise range
* easy conversion from `vim.fn.getpos(".")` to char or line-wise range

xiyaowong: > Before merge, we (I can) need to document this in the readme

I will write the wiki so that more people can improve it together. Write examples at the same time

theol0403:
> I don't think implicit doautocmd is better. Users can do it themselves.

I'm not sure what you mean, currently `FocusLost` is never triggered. I think the correct solution to https://github.com/vscode-neovim/vscode-neovim/issues/434 is not to provide a custom hook that users can use, but to make `FocusLost` function as expected so existing nvim code works automatically. 

> Using on can handle bidirectional synchronization more conveniently, such as editor.options(&ts, &sw)).

Can you explain further? Sure, I agree that hooks are useful, but I think all the hooks implemented so far can be better done using built-in neovim mechanisms that work out of the box with other nvim code.

xiyaowong:
As a demonstration, I will use hooks to support this #1530

> > I don't think implicit doautocmd is better. Users can do it themselves.
> 
> I'm not sure what you mean, currently `FocusLost` is never triggered. I think the correct solution to #434 is not to provide a custom hook that users can use, but to make `FocusLost` function as expected so existing nvim code works automatically.

Doesn't necessarily work as expected. When the window state changes, the internal state of vscode may also change. But I'm not sure, but I think there will be a lot of edge cases.


```lua
vscode.on("window_state_changed", function(focused)
  vim.cmd.doautocmd(focused and "FocusGained" or "FocusLost")
end)
```


> 
> > Using on can handle bidirectional synchronization more conveniently, such as editor.options(&ts, &sw)).
> 
> Can you explain further? Sure, I agree that hooks are useful, but I think all the hooks implemented so far can be better done using built-in neovim mechanisms that work out of the box with other nvim code.

As a demonstration, I will use hooks to support this #1530

runtime/lua/vscode-neovim/api.lua
xiyaowong: > Either we provide a flexible API to accept many different types of ranges, or we create external helpers. Not a mix of both. If we want to make helpers, then we should remove the `line_range` as it will be redundant.
> 
> What we need, either provided by the API or by helpers:
> 
> * lsp range
> * convenient/concise char-wise range
> * convenient/concise line-wise range
> * easy conversion from `vim.fn.getpos(".")` to char or line-wise range

Makes sense

theol0403: Maybe all we need is `range = vscode.util.mark_to_range(".", linewise=true)`, `range = vscode.util.linewise(3, 4)`, `range = vscode.util.charwise({3, 0}, {4, 1})`.

xiyaowong:
Example that supports modeline, and synchronize all editor options(tab options and number style) https://github.com/xiyaowong/vscode-neovim/compare/feat/vscode-action...xiyaowong:vscode-neovim:tmp/support-modeline

runtime/lua/vscode-neovim.lua
```diff
@@ -1,30 +1,34 @@
...
+  action = api.action,
+  call = api.call,
+  -- hooks
+  on = api.on,
+  -- vscode settings
+  has_config = api.has_config,
+  get_config = api.get_config,
+  update_config = api.update_config,
+  -- notifications
+  notify = api.notify,
```
theol0403, line 28: We might need to make an overload that somehow produces an error if people try to use the old `notify` function.

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,302 @@
...
-  return M.call_extension("range-command", command, "V", line1, line2, 1, 1, leaveSelection, { ... })
+---Run action, this function is asynchronous
+---@param name string The action name, generally a vscode command
+---@param opts? table Optional options table, all fields are optional
+---            - args: (table) Optional arguments for the action
+---            - range: (table) Specific range for the action. In visual mode, this parameter is generally not needed.
+---                     Three formats supported (All values are 0-indexed):
+---                        - [start_line, end_line]
+---                        - [start_line, start_character, end_line, end_character]
+---                        - {start = { line = $start_line , character = $start_character}, end = { line = $end_line , character = $end_character}}
```
theol0403, line 44: Why the dollar sign?

runtime/lua/vscode-neovim/api.lua
```diff
@@ -1,133 +1,302 @@
...
+  --- Display a notification only one time.
+  ---
+  --- Like |vscode.notify()|, but subsequent calls with the same message will not
+  --- display a notification.
+  ---
+  ---@param msg string Content of the notification to show to the user.
+  ---@param level integer|nil One of the values from |vim.log.levels|.
+  ---@param opts table|nil Optional parameters. Unused by default.
+  ---@return boolean true if message was displayed, else false
+  function M.notify_once(msg, level, opts)
```
theol0403, line 292: What is the expected behavior with `notify_once("one"), notify_once("two"), notify_once("one")`? For it to ignore all duplicate messages, or only ones that are in a row?

runtime/lua/vscode-neovim/highlight.lua
theol0403, line 1: Can you fix the namespace name here (line 6), there shouldnt be symbols. `vscode-buffer-highlights`

theol0403 (approved):

theol0403:
> vim.notify = vscode.notify
vim.notify_once = vscode.notify_once

Consider making this the default.

runtime/lua/vscode-neovim/api.lua
xiyaowong: `vim.notify_once`

runtime/lua/vscode-neovim/api.lua
xiyaowong: means variable, can remove it

runtime/lua/vscode-neovim.lua
xiyaowong: This is completely consistent with the format of vim.notify, and users decide whether to replace it or not. This is the same as the existing notification plugin usage in nvim community.

xiyaowong:
> > vim.notify = vscode.notify
> > vim.notify_once = vscode.notify_once
> 
> Consider making this the default.

It should be decided by the users themselves.Or provide a configuration item, which is obviously unnecessary.
https://github.com/rcarriga/nvim-notify
https://github.com/folke/noice.nvim

runtime/lua/vscode-neovim/api.lua
theol0403: Not sure you understood - what is the default behavior of vim.notify_once?

xiyaowong: https://github.com/neovim/neovim/blob/0892c1cd94672e0dffe3b97fc4a85ea517b05598/runtime/lua/vim/_editor.lua#L620

runtime/lua/vscode-neovim/api.lua
xiyaowong: > Maybe all we need is `range = vscode.util.mark_to_range(".", linewise=true)`, `range = vscode.util.linewise(3, 4)`, `range = vscode.util.charwise({3, 0}, {4, 1})`.

It seems that these functions are not necessary. Firstly, specifying the range is rarely needed. Secondly, the range parameter is flexible enough, users just need to remember to use 0-indexed values.

runtime/lua/vscode-neovim/api.lua
theol0403: Alright. If the neovim implementation calls vim.notify, we don't need to override it right? Only override vim.notify.

xiyaowong:
@theol0403 "leave_selection" seems ambiguous. How about changing it to "restore_selection"?

justinmk:
> `window_state_changed` same effect as FocusLost/FocusGained

I haven't looked at the details but agree with the theme of @theol0403 's comments in this PR: it's very important to avoid creating an [inner platform](https://en.wikipedia.org/wiki/Inner-platform_effect). The `on()` mechanism should not be an alternative to Nvim's (and VSCode's) existing event system, that creates confusion and adds unnecessary surface area for both users and contributors to understand.

If there are bugs with Nvim's event system (like a missing Focus event) then they should be fixed, rather than adding another layer. vscode-neovim has suffered from having too many "features" which were fragile. We need to continue to add only things that are strongly justified, not "nice to have".

theol0403:
> It should be decided by the users themselves.Or provide a configuration item, which is obviously unnecessary.

Generally, I agree, but vscode-neovim sits somewhere between being a GUI and a distribution. Since there are only two possible implementations of notify that works with vscode (output window, or vscode notification), we should provide the best default. In that case, I think connecting it to vscode notification is the best default.

theol0403:
> @theol0403 "leave_selection" seems ambiguous. How about changing it to "restore_selection"?

Maybe `select_range`?

xiyaowong:
Is it good or bad to have a theoretically fragile feature that rarely encounters problems? Currently, it seems difficult to avoid "fragile" situations because vscode-neovim is not a GUI client for NVIM, and the implementation of many features needs to consider both VSCode and NVIM.

xiyaowong:
It is possible to use nvim_exec_autocmds, but it is not convenient to use.

theol0403:
> It is possible to use nvim_exec_autocmds, but it is not convenient to use.

I think we should try.

xiyaowong:
> > It is possible to use nvim_exec_autocmds, but it is not convenient to use.
> 
> I think we should try.

That means we have to consider the issue of "autocmd-nested". For example, in [the implementation of feature A](https://github.com/xiyaowong/vscode-neovim/compare/feat/vscode-action...xiyaowong:vscode-neovim:tmp/support-modeline), when options are changed, autocmds without the "nested" flag will not be executed.

theol0403:
> > > It is possible to use nvim_exec_autocmds, but it is not convenient to use.
> > 
> > I think we should try.
> 
> That means we have to consider the issue of "autocmd-nested". For example, in
> 
> https://github.com/xiyaowong/vscode-neovim/blob/27b7714b51ad801e7d25ad047de119c00403c524/runtime/lua/vscode-neovim/sync-options.lua#L108-L114
> 
> when options are changed, the OptionSet autocmds without the "nested" flag will not be executed.

I think your example of using the hooks for option sync is a good use of hooks, and we can keep that.

runtime/lua/vscode-neovim.lua
xiyaowong: There should be few users using the old Lua API. If we consider compatibility for this, our maintenance cost will be much higher compared to the cost for users to execute `:%s/vscode.notify/vscode.action/g`

image
image

runtime/lua/vscode-neovim/api.lua
xiyaowong: Oh, right. I didn’t realize that.

xiyaowong:
> > @theol0403 "leave_selection" seems ambiguous. How about changing it to "restore_selection"?
> 
> Maybe `select_range`?

Maybe still 'restore_selection' is better, because the selection may not equal the 'range' after executing some vscode commands.

xiyaowong:
> * We should look into triggering FocusGained/Lost by default

Sure

runtime/lua/vscode-neovim.lua
justinmk: > There should be few users using the old Lua API.

Yeah that api is very new so breakage is acceptable. Just mention it in release notes.

src/actions.ts
```diff
@@ -111,7 +111,9 @@ class ActionManager implements Disposable {
 
     private initHooks() {
         this.disposables.push(
-            window.onDidChangeWindowState((e) => this.fireNvimEvent("window_state_changed", e.focused)),
+            window.onDidChangeWindowState((e) =>
+                this.client.command(`doautocmd ${e.focused ? "FocusGained" : "FocusLost"}`),
```
theol0403, line 115: Would it be better to use `nvim_exec_autocommand`?

theol0403 (approved):

theol0403:
I am happy with this PR, pending readme documentation and ideally examples.

README.md
```diff
@@ -268,24 +267,24 @@ See gif in action:
...
-| <ul><li>`VSCodeNotifyRange(command, line1, line2, leaveSelection, ...)` </li><li>`VSCodeCallRange`</li><li>`require("vscode-neovim").notify_range`</li><li>`require("vscode-neovim").call_range`</li></ul>                          | Produce linewise VSCode selection from `line1` to `line2` and invoke VSCode command. Setting `leaveSelection` to 1 keeps VSCode selection active after invoking the command. Line is 1-based. |
-| <ul><li>`VSCodeNotifyRangePos(command, line1, line2, pos1, pos2, leaveSelection ,...)`</li><li>`VSCodeCallRangePos`</li><li>`require("vscode-neovim").notify_range_pos`</li><li>`require("vscode-neovim").call_range_pos`</li></ul> | Produce characterwise VSCode selection from `line1.pos1` to `line2.pos2` and invoke VSCode command. Pos is \[1,1\]-based.                                                                     |
+There are a few helper functions that are used to invoke VSCode commands from Neovim.
```
theol0403, line 274: Need to include VSCodeCall variants. Needed for bindings like visual-mode comment selection that exits visual mode after.

README.md
```diff
@@ -268,24 +267,24 @@ See gif in action:
 
 ![multicursors](/images/multicursor.gif)
 
-### Invoking VSCode actions from neovim
+### API
```
theol0403, line 270: Make API a ## heading, with emoji

README.md
```diff
@@ -309,6 +308,195 @@ nnoremap ? <Cmd>call VSCodeNotify('workbench.action.findInFiles', { 'query': exp
...
+      end
+    end,
+  })
+end
+
+-- Format current document
+vscode.action("editor.action.formatDocument")
+
+do -- Comment the three lines below the cursor
+  local lnum = vim.fn.line(".")
```
theol0403, line 387: Careful for 1-indexed lines, would be better to have the example show the vim.fn.line(.) - 1

theol0403 (approved):

README.md
xiyaowong: ⚡️

README.md
```diff
@@ -384,12 +386,22 @@ end
 vscode.action("editor.action.formatDocument")
 
 do -- Comment the three lines below the cursor
-  local lnum = vim.fn.line(".")
+  local curr_line = vim.fn.line(".") - 1 -- just want you remember using 0-indexed value for the range :)
```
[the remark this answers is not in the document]
xiyaowong: 🤣🤣

README.md
```diff
@@ -268,18 +267,20 @@ See gif in action:
...
-| <ul><li>`VSCodeNotifyRangePos(command, line1, line2, pos1, pos2, leaveSelection ,...)`</li><li>`VSCodeCallRangePos`</li><li>`require("vscode-neovim").notify_range_pos`</li><li>`require("vscode-neovim").call_range_pos`</li></ul> | Produce characterwise VSCode selection from `line1.pos1` to `line2.pos2` and invoke VSCode command. Pos is \[1,1\]-based.                                                                     |
+There are a few helper functions that are used to invoke VSCode commands from Neovim.
+
+-   `VSCodeNotify(command, ...)` `VSCodeCall(command, ...)` Invoke VSCode command with optional arguments.
+-   `VSCodeNotifyRange(command, line1, line2, leaveSelection, ...)`
+    `VSCodeCallRange(command, line1, line2, leaveSelection, ...)` Produce linewise VSCode selection from `line1` to
+    `line2` and invoke VSCode command. Setting `leaveSelection` to 1 keeps VSCode selection active after invoking the
+    command. Line is 1-based.
+-   `VSCodeNotifyRangePos(command, line1, line2, pos1, pos2, leaveSelection ,...)`
+    `VSCodeNotifyRangePos(command, line1, line2, pos1, pos2, leaveSelection ,...)` Produce characterwise VSCode
```
theol0403, line 282: duplicate

README.md
```diff
@@ -268,18 +267,20 @@ See gif in action:
...
-| <ul><li>`VSCodeNotifyRange(command, line1, line2, leaveSelection, ...)` </li><li>`VSCodeCallRange`</li><li>`require("vscode-neovim").notify_range`</li><li>`require("vscode-neovim").call_range`</li></ul>                          | Produce linewise VSCode selection from `line1` to `line2` and invoke VSCode command. Setting `leaveSelection` to 1 keeps VSCode selection active after invoking the command. Line is 1-based. |
-| <ul><li>`VSCodeNotifyRangePos(command, line1, line2, pos1, pos2, leaveSelection ,...)`</li><li>`VSCodeCallRangePos`</li><li>`require("vscode-neovim").notify_range_pos`</li><li>`require("vscode-neovim").call_range_pos`</li></ul> | Produce characterwise VSCode selection from `line1.pos1` to `line2.pos2` and invoke VSCode command. Pos is \[1,1\]-based.                                                                     |
+There are a few helper functions that are used to invoke VSCode commands from Neovim.
+
+-   `VSCodeNotify(command, ...)` `VSCodeCall(command, ...)` Invoke VSCode command with optional arguments.
```
theol0403, line 276: Maybe more concise to say "- `VSCodeNotify(command, ...)`/`VSCodeCall`: invoke VSCode command with optional arguments"

xiyaowong: Originally planned to write this way, but changed it later…

xiyaowong:
Ready to merge⚡️

theol0403:
Thank you for your hard work!

theol0403:
@xiyaowong looking forward to seeing the settings sync PR

Issue #1145, opened 2023-01-30
cmd+d to do multiselect / multicursor and then apply neovim commands to each occurrence does not work


55Cancri:
When I press `viw` to select a word, then press `cmd+d` to select the other instances, it first activates Match Case and Match Whole Word settings, then starts highlighting the other occurrences.

However, when I then press something like `ciw` to change the words while all the occurrences are highlighted, only the first one changes.
1. How do I begin selecting occurrences immediately, without first activating those settings?
2. How do I apply the neovim keys to each currently selected occurrence instead of just the first one?

theol0403:
Multi cursor not supported in visual mode

55Cancri:
> Multi cursor not supported in visual mode

Normal vscode vim with jumpy2 plugins works wonders

xiyaowong:
Can provide a function that encapsulates editor.action.addSelectionToNextFindMatch to support this feature
<img src="https://github.com/vscode-neovim/vscode-neovim/assets/47070852/65366365-07f6-4f7a-9949-9dbeed7076b0" width="400px" />

theol0403:
That would be nice, if we wrapped all the muticursor commands in normal mode like "select all word under cursor" and all the other ones, that would nicely transition you into insert mode.

ghry5:
It's a little janky but seems to work. This is what I came up with:
```
  vim.keymap.set("v", "<D-d>", function()
    local sel_start = vim.fn.getpos("v")
    local sel_end = vim.fn.getpos(".")
    vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<esc>", true, false, true), "v", false)
    vim.defer_fn(function()
      vim.cmd.startinsert()
      vscode.notify_range_pos("editor.action.addSelectionToNextFindMatch", sel_start[2], sel_end[2], sel_start[3],
        sel_end[3], true)
    end, 20)
  end, opts)
```

To work it needs to forward the cmd+d keystroke from vscode:
```
  {
    "command": "vscode-neovim.send",
    "key": "cmd+d",
    "when": "editorTextFocus && neovim.mode != insert",
    "args": "<D-d>"
  },
  {
    "key": "cmd+d",
    "command": "editor.action.addSelectionToNextFindMatch",
    "when": "editorFocus && neovim.mode != 'visual'"
  },
  {
    "key": "cmd+d",
    "command": "-editor.action.addSelectionToNextFindMatch",
    "when": "editorFocus"
[...]

PR #1546, opened 2023-10-17
feat: wrappers for vscode multi cursor commands

xiyaowong:
- [x] `editor.action.addSelectionToNextFindMatch`
- [x] `editor.action.addSelectionToPreviousFindMatch`
- [x] `editor.action.selectHighlights`
- [x] `selectAllSearchEditorMatches`

Example:
Add selection to next find match:
- Invoke action in nvim: `vscode.action('editor.action.addSelectionToNextFindMatch')`
- Execute command in vscode: `neovim:editor.action.addSelectionToNextFindMatch`

keybindings.json
```json
  {
    "command": "neovim:editor.action.addSelectionToNextFindMatch",
    "key": "ctrl+d",
    "when": "editorFocus"
  }
```

Resolve: #1145


xiyaowong:
@archilkarchava The approach is the same, but it is more convenient and concise in vscode, and it is easier to handle edge cases, such as cursor position jitter.

theol0403:
I think it would be good to make this more general. We should provide a generic "execute this command from normal mode that moves into insert mode and creates a selection.

Then, the API could be (rough example):

```
{
command: "neovim:wrap-cursor-command",
args: "addNextSelection"
}
```

theol0403:
The reason is that there are a lot of multi cursor commands that would be good to wrap without maintenance burden.

src/cursor_manager.ts
```diff
@@ -528,6 +534,58 @@ export class CursorManager implements Disposable {
...
+
+        if (this.main.modeManager.isInsertMode) {
+            await run();
+            return;
+        }
+
+        if (editor.selection.isEmpty) {
+            await run();
+            const selections = editor.selections;
+            this.main.client.input("<Esc>"); // Correcting the position
```
theol0403, line 552: I don't love the delays. Maybe you can await the input and then await a cursor update.

xiyaowong:
> I think it would be good to make this more general. We should provide a generic "execute this command from normal mode that moves into insert mode and creates a selection.
> 
> Then, the API could be (rough example):
> 
> ```
> {
> command: "neovim:wrap-cursor-command",
> args: "addNextSelection"
> }
> ```

I don't think this is better. With prefix "neovim:", the user can understand at a glance what the command is for. Where the maintenance cost is increased, the command is automatically registered and unified.

xiyaowong:
Even if users don't know these functions, users can directly use `editor.action.addSelectionToNextFindMatch`, which is automatically replaced, which is also the design intention of Action in PR lua api.

src/cursor_manager.ts
xiyaowong: > await the input

I just discovered that "input" can be awaited. :smile:

theol0403:
What if you want to wrap `editor.action.smartSelect.grow` (random example), or maybe a command from another vscode extension. You would have to make a PR here every time. Is there not a clean way to make a generic wrapper for any vscode command that you want to transition into insert mode?

xiyaowong:
These wrappers themselves are special cases, and the few built-in commands are enough.

xiyaowong:
This is a general example in VSCode, hoping to be helpful to those in need

```lua
vim.keymap.set({ 'n', 'i', 'x' }, '<C-d>', function()
  local vscode = require('vscode-neovim')
  local mode = vim.api.nvim_get_mode().mode

  local select = function(cb)
    vscode.action('editor.action.addSelectionToNextFindMatch', { callback = cb })
  end

  if mode == 'i' then return select() end

  if mode:lower() == 'v' then
    vim.api.nvim_input('<ESC>a')
    vim.schedule(function()
      local range = vim.lsp.util.make_given_range_params(nil, nil, 0, 'utf-16').range
      vscode.action('noop', {
        range = range,
        callback = select,
        restore_selection = false,
      })
    end)
    return
  end

  if mode == 'n' then select(function()
    vim.api.nvim_input('<ESC>')
    select()
  end) end
end)
```

theol0403:
> This is a general example in nvim, hoping to be helpful to those in need

Again, is there any way we can provide this as part of the API or included as a helper function?

src/cursor_manager.ts
```diff
@@ -495,6 +501,57 @@ export class CursorManager implements Disposable {
...
+
+        await cursorReady();
+
+        if (this.main.modeManager.isInsertMode) return run();
+
+        if (editor.selection.isEmpty) {
+            await run();
+            const selections = editor.selections;
+            await this.main.client.input("<Esc>"); // Correcting the position
+            await wait(100); // Compensate for debounce time
```
theol0403, line 519: Can you pull this constant in, instead of magic number?

xiyaowong:
> > This is a general example in nvim, hoping to be helpful to those in need
> 
> Again, is there any way we can provide this as part of the API or included as a helper function?

There should be, but there should be more things to consider in implementation. 

Provides a simple action to switch from normal/visual to insert mode, but keeps the selections of vscode unchanged.

xiyaowong:
These commands can be implemented in nvim. All crowded into the project will really make the project bloated. Closed this pr.

xiyaowong:
@theol0403 
It seems like I made several serious mistakes. I tried to click the sync button in vscode as usual, but I found out that it actually pushed the commits to the master branch. Is there any way to revert these commits? I am truly sorry. 😭 

image

theol0403:
@xiyaowong you can revert them manually, or make a force push as a last resort. You may have to temporarily disable branch protection rules to allow that.

PR #1565, opened 2023-10-23
chore: rename extension description

theol0403:
This should improve discoverability by ranking higher in the vscode search, because of the "vim" keyword.


theol0403:
@xiyaowong not sure, worth a try. I think since this extension affects so much, it is a good example of an extension that should load as soon as possible.

Issue #1068, opened 2022-10-27
Error occurs when shifting (<) nested blocks (overlapping ranges)


linw789:
Running command: `<` + `11` + `j` gives error: _DocumentChangeManager: Error applying neovim edits, error: Overlapping ranges are not allowed!_

`>` + `11` + `j` works however.

```c
    if (foo == bar)
    {
        foo += 1;
        if (baz == bar)
        {
            baz += 1;
        }
    }
    else if (foo == 0)
    {
        bar == 0;
    }
```

theol0403:
Thanks for this! I've occasionally seen that issue, but haven't been able to find a reliable reproduction. I can reproduce this though!

This will likely be fixed if/when I find time to rewrite the nvim -> vscode sync code.

FangLin2001:
I encounter this problem too. Just short nested codes can cause this.

spmacdonald:
Just encountered this myself.  Any suggested work around?

MateuszSadowski:
I'm encountering the same issue.

PR #1567, opened 2023-10-24
fix(sync): fallback to replacing text on failed edits

xiyaowong:
This PR does not fix issue #1068. It only falls back to directly replacing text when there is an error. And the original error will still be displayed normally so that users can notice and provide feedback. 

Through debugging, it seems that this portion of the code is problematic:

https://github.com/vscode-neovim/vscode-neovim/blob/f2f57f82f1b9d7b16f52fd7977b7fb3d1e447280/src/document_change_manager.ts#L332C1-L333C65

### DIFF
--- a/src/document_change_manager.ts
+++ b/src/document_change_manager.ts
@@ -322,13 +322,14 @@ export class DocumentChangeManager implements Disposable {
                         logger.debug(`No visible text editor for document, skipping`);
                         continue;
                     }
-                    let oldText = doc.getText();
                     const eol = doc.eol === EndOfLine.CRLF ? "\r\n" : "\n";
-                    let newText = newLines.join(eol);
+                    const oldText = doc.getText();
+                    const newText = newLines.join(eol);
                     // add few lines to the end otherwise diff may be wrong for a newline characters
-                    oldText += `${eol}end${eol}end`;
-                    newText += `${eol}end${eol}end`;
-                    const diffPrepare = diffLineToChars(oldText, newText);
+                    const oldTextForDiff = oldText + `${eol}end${eol}end`;
+                    const newTextForDiff = newText + `${eol}end${eol}end`;
+                    const diffPrepare = diffLineToChars(oldTextForDiff, newTextForDiff);
+                    // FIXME: Find the root cause of the error: Overlapping ranges are not allowed!
                     const d = diff(diffPrepare.chars1, diffPrepare.chars2);
                     const ranges = prepareEditRangesFromDiff(d);
                     if (!ranges.length) {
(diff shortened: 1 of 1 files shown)


theol0403:
A couple years ago I tried refactoring the nvim -> vscode buffer sync. It is possible, but I ended up running out of time to get it right. 

https://github.com/vscode-neovim/vscode-neovim/pull/992#issuecomment-1228734888

Basically, the way it currently works is nvim sends line updates, then vscode-neovim constructs an internal text representation of the buffer using those line updates. Then, it hashes each line, and performs a custom diff with an internal text representation of the old buffer. Then, knowing the lines that change, it decides to replace those lines, or does another diff to create more surgical changes (deleting/adding/replacing within the line).

I'm not sure why this is all needed. Instead, I want to take the nvim line updates, and directly send them to vscode changes. Maybe do a quick diff to simplify things. There is this tool fast-meyers-diff which not only provides an extremely fast diff, but gives you the results in a form you can directly send to the vscode change builder.

I guess we can merge this PR, but I'd rather not have to add more code to this... instead either fix the bug or totally simplify the sync code.

xiyaowong:
I started to pay attention to this project when you released 0.4.0, so I don't know much historical information.

I have tested the time-consuming of replacing text directly and using diff. My conclusion is that direct replacing text is very fast and diff is very slow.

Since this is a problem that has been discussed before, it is best to solve it once and for all.

xiyaowong:
> directly send them to vscode changes. 
> There is this tool fast-meyers-diff

It's worth looking into. By the way, have you tested it to verify its feasibility?

theol0403:
From some early testing, fast-meyers-diff seems to be very fast, and might be fast enough for charwise diffing of the entire buffer. 

However, diff might not even be necessary if we can directly apply changes from nvim. We could also investigate using nvim lsp to produce changes.

xiyaowong:
> We could also investigate using nvim lsp to produce changes.

I've thought about this, too😀

theol0403:
This was some of my work a while ago: https://github.com/theol0403/vscode-neovim/commit/c6e6c7e94d5a3b2f1cb2ddebe97823285011bc0a

xiyaowong:
Amazing!
image

xiyaowong:
> This was some of my work a while ago: [theol0403@c6e6c7e](https://github.com/theol0403/vscode-neovim/commit/c6e6c7e94d5a3b2f1cb2ddebe97823285011bc0a)

@theol0403 
The modified code has been greatly simplified, making the synchronization logic much easier to understand.

In a nutshell, this solution should be better than the previous one in terms of code simplicity, performance, and stability (addressing existing issues).

I noticed that this was a commit made a year ago. Is it possible that there were many potential issues with this plan that prevented you from continuing the research?

theol0403:
@xiyaowong no major issues other than time. There were some bugs and edge cases, including desyncs. I didn't have time to complete it. It should definitely be possible.

xiyaowong:
@theol0403 How about I pick your commit and continue working? We can identify problem points and then solve them by generating a large number of test cases.

theol0403:
Feel free

xiyaowong:
> However, diff might not even be necessary if we can directly apply changes from nvim. 

So far, I'm optimistic about this approach

theol0403:
Sounds good! One note is that *some* sort of diff is useful (like in my commit) for changes that aren't line add/delete because then the vscode builder can be used to make surgical changes (and not flash syntax hl and such).

xiyaowong:
In general, it is rare to have a file with 50,000 lines. However, directly replacing the document at this time can meet normal editing needs. If optimization is performed, the effect may be better, although this approach is a bit crude, it is stable and super simple. Of course, using "fast-myers-diff" has absolute advantages in terms of speed, but it is mainly reflected in very large texts, and the speed advantages in small texts are not noticeable in terms of experience. However, this approach is theoretically the most elegant, but we may need to consider some edge cases.

I have completely copied the content you originally committed, and everything is working fine👍. What is probably most needed at the moment is a large amount of testing.

theol0403:
Replacing the entire content (or even replacing an entire line) is not good. VSCode has to re-compute LSP/syntax hl for the entire buffer. The advantage of fast-myers-diff is that it can produce a "smallest number of changes to go from state a to state b", and then surgically perform those in vscode. So for example, deleting a word will actually tell vscode to delete a word. 

I think the viable options include replaying changes from nvim (and doing a diff on modified lines), doing a diff on the entire buffer, or marking lines as "dirty" (through logic or diffing a hash) and then doing a diff on those lines only.

theol0403:
I think one possible reason of the prior implementation is that nvim might send multiple updates at once, and so it bundled them together to reduce the number of work vscode was doing. Not sure if that behavior is worth keeping (albeit with simpler code).

xiyaowong:
> ...reduce the number of work vscode was doing. Not sure if that behavior is worth keeping

Basic edits such as ciw, diw, diw, x, etc., only trigger one update. When using plug-ins like surround, there may be a slight increase, but the quantity is also negligible. Apart from that, there doesn’t seem to be many other editing needs :eyes: . At least, these are the ones I mostly use in my usual work.

theol0403:
What about vim commands liks %s, %g, macros, `99.`, etc.

xiyaowong:
In a 20,000-line file, replace "this" with "that" using "%s/this/that/g" in 3500 places.

Number of edits: 23549
Time consuming: 63869ms 😂 

Current solution: < 1s

theol0403:
Yikes, maybe there is a relatively easy solution to move the document.edit(builder) outside of the loop and build the builder incrementally. The only problem is that nvim might give document positions that assume the prior edit has been applied.

theol0403:
Maybe also try benchmarking:
* old text-based accumulation code
* fast-myers-diff after accumulation
* execute diff operations

That will still reduce a lot of code.

xiyaowong:
1. Merge the changes as now, using fast-myers-diff to replace the existing diff
2. Use multiple solutions, such as directly using text replacement when the number of changes reaches a certain level.

xiyaowong:
> Maybe also try benchmarking:
> 
> * old text-based accumulation code
> * fast-myers-diff after accumulation
> * execute diff operations
> 
> That will still reduce a lot of code.

Perhaps we can add an additional alternative solution, such as using a replacement approach when the changed content is very disorganized and there are a large number of changes.

theol0403:
I don't think using heuristics is the best approach - it varies based on computer speed, and it is hard to decide magic numbers, and the number of changes is not well-known. Either we:
* accumulate direct (save nvim events) edits, and then apply them at once
* accumulate internal (apply events to internal record), then use clever diffing (current solution, or switch to fast-myers) to efficiently replay changes

theol0403:
> Merge the changes as now, using fast-myers-diff to replace the existing diff

I think this is what we should try first, but we have to be careful to benchmark fast-myers-diff with the old approach (which keeps hashes of each line and does a linewise hash diff first to get changed lines).

xiyaowong:
Additionally, accumulating changes is necessary. For example, using 'cc' to modify the current line and then returning to normal mode, nvim will send multiple change events to remove the indentation. I guess this is exactly the problem the original author encountered initially.

xiyaowong:
> > Merge the changes as now, using fast-myers-diff to replace the existing diff
> 
> I think this is what we should try first, but we have to be careful to benchmark fast-myers-diff with the old approach (which keeps hashes of each line and does a linewise hash diff first to get changed lines).

I will give it a simple try, but since I have no experience in text processing, if there are too many issues, I won't delve too deep into it. It's because it can be time-consuming, especially when it comes to edge cases and testing🤧. Moreover, it currently works 99.9% of the time😌.

xiyaowong:
Diff the entire text directly, and it seems there is no problem.

https://github.com/xiyaowong/test-myers-diff/actions/runs/6655042639

xiyaowong:
> eol

I have unified it, and eol should be a minor problem.

theol0403:
Why no ubuntu?

Also, you should add a test where:
* EOL gets changed from LF to CRLF and vice-versa
* Tabs get converted to spaces and vice-versa
* The entire file gets `ggVGJ` compressed into a single line
* The entire file gets deleted, 0 characters

xiyaowong:
> Why no ubuntu?

Always occurs error: ozone_platform_x11.cc(247)] Missing X server or $DISPLAY, and no time to fix it.

Only testing the diff, there are no platform differences.

> 
> Also, you should add a test where:
> 
> * EOL gets changed from LF to CRLF and vice-versa
> * Tabs get converted to spaces and vice-versa
> * The entire file gets `ggVGJ` compressed into a single line
> * The entire file gets deleted, 0 characters

I don't see any major issues. The test cases even include changes to binary content.

xiyaowong:
It failed, but is it possible that the test was too strict? :joy:

https://github.com/xiyaowong/test-myers-diff/actions/runs/6655774024/job/18086786762

theol0403:
> ozone_platform_x11.cc(247)] Missing X server or $DISPLAY,

Ah yes, this is why https://github.com/vscode-neovim/vscode-neovim/blob/476502604a5fcfef89c528532de17b391036aed6/.github/workflows/build_test.yml#L39 is used.

xiyaowong:
> > ozone_platform_x11.cc(247)] Missing X server or $DISPLAY,
> 
> Ah yes, this is why
> 
> https://github.com/vscode-neovim/vscode-neovim/blob/476502604a5fcfef89c528532de17b391036aed6/.github/workflows/build_test.yml#L40
> 
> is used.

I accidentally deleted it without reading carefully. Haha.

theol0403:
What happens when the user switches the EOL from CRLF to LF? Your tests don't catch that because you normalize the line endings before doing the diff.

xiyaowong:
We all use one EOL for synchronization and convert it back when applying

xiyaowong:
image

theol0403:
I knew what I was doing when I wrote that code! :smile:

theol0403:
Awesome. I am a little worried about the time it takes to make large diffs, I don't know how it compares to the old system.

It is possible we may need to do a two-pass algorithm, where we first do a line-wise diff (ideally using a hashing technique instead of comparing each line charwise), where we insert and remove chunks, and then do a char-wise diff on the changed lines.

theol0403:
Also, it may be possible to improve the performance of the calcDiffWithPosition function, it does a lot of string manipulation and might be better to do some sort of pure maths instead.

theol0403:
> Also, it may be possible to improve the performance of the calcDiffWithPosition function, it does a lot of string manipulation and might be better to do some sort of pure maths instead.

I think the solution is to create an array containing the length of every line, and then summing those lengths using the same algorithm instead of concatenating the strings.

xiyaowong:
Well, it's just a test for accuracy now. The next step is to avoid unnecessary operations. By the way, I should mention that the test cases are very stringent and this kind of situation is unlikely to occur in daily use, such as using ggVGJ in an 8k lines file... When performing such operations, VSCode can also become slow.

xiyaowong:
I can test tomorrow. I have to go to sleep now, it’s already 1 am in my time zone.

xiyaowong:
version 2 has the best performance

theol0403:
I tried making a version that lazy-loads the character accumulation, but I think the difference is minor. In fact, in my testing, version 1 to version 3 have the same performance. I think the computation is nothing compared to fast-myers-diff.

## Version 3
```
class LazyLineLengths {
  private lines: string[];
  private cache: number[] = [];
  private cacheLength = 0;
  private generator: Generator<number>;

  constructor(text: string) {
    this.lines = text.split("\\n");
    this.cache = new Array(this.lines.length);
    this.generator = this.accumulateLengths();
  }

  private *accumulateLengths(): Generator<number> {
    let cumulativeLength = 0;
    for (let i = 0; i < this.lines.length; i++) {
      cumulativeLength += this.lines[i].length + 1;
      this.cache[i] = cumulativeLength;
      this.cacheLength++;
      yield cumulativeLength;
    }
  }

  getLengthAt(line: number): number {
    while (this.cacheLength <= line) {
      this.generator.next();
    }
    return this.cache[line];
  }

  // given an array of cumulative line lengths, find the line number given a character position after a known start line.
  // return the line as well as the number of characters until the start of the line.
  findLine(pos: number, startLine: number): [number, number] {
    let low = startLine,
      high = this.lines.length - 1;
    while (low < high) {
      const mid = low + Math.floor((high - low) / 2); // can adjust pivot point based on probability of diffs being close together
      if (this.getLengthAt(mid) <= pos) {
        low = mid + 1;
      } else {
        high = mid;
      }
    }
    const char = low > 0 ? this.getLengthAt(low - 1) : 0;
    return [low, char];
  }
}

// fast-myers-diff accepts a raw 1D string, and outputs a list of operations to apply to the buffer.
// However, we must compute the line and character positions of the operations ourselves.
// Assuming the operations are sequential, we can use a binary search to find the line number given a character position,
// with the search space being the cumulative line lengths, bounded on the left by the last line.
// Then, given a character position, we can start counting from the cursor to find the line number, and the remainder is the character position on the line.
function* calcDiffWithPosition(
  oldText: string,
  newText: string
): Generator<DocumentChange> {
  const patch = calcPatch(oldText, newText);
  // generate prefix sum of line lengths (accumulate the length)
  const lineLengths = new LazyLineLengths(oldText);
  let lastLine = 0;
  for (const [start, end, text] of patch) {
    const [lineStart, charToLineStart] = lineLengths.findLine(start, lastLine);
    const [lineEnd, charToLineEnd] = lineLengths.findLine(end, lineStart);
    const charStart = start - charToLineStart;
    const charEnd = end - charToLineEnd;
    const range = new Range(
      new Position(lineStart, charStart),
      new Position(lineEnd, charEnd)
    );
    lastLine = lineEnd;
    yield {
      range,
      text,
      rangeLength: end - start,
    };
  }
}
```

PR #1569, opened 2023-10-25
fix(keybinding): fix `Ctrl+c`

xiyaowong:
This PR is mainly to fix invalid binding, but I made some adjustments:

First of all, it doesn't make much sense to bind <kbd>ctrl+c</kbd>, so <kbd>ctrl+c</kbd> is not bound by default. If <kbd>ctrl+c</kbd> is enabled, the default binding is the escape command. If users still have needs, they should adjust keybindings themselves.


theol0403:
It is important that the mapping stays the same as it used to be https://github.com/vscode-neovim/vscode-neovim/blob/5e5d11fca303806795e9ce07cd60cdada9a18ae4/package.json#L328
* it should map to escape by default, because that is default in nvim and also some people use it as muscle memory
* the when clauses should stay the same, as it reads "ctrl-c should close open menus only if we are in normal mode, otherwise ctrl-c should enter normal mode from insert/replace/visual"

Also there is a typo in the when clause

xiyaowong:
> It is important that the mapping stays the same as it used to be
> 
> https://github.com/vscode-neovim/vscode-neovim/blob/5e5d11fca303806795e9ce07cd60cdada9a18ae4/package.json#L328
> 
> * it should map to escape by default, because that is default in nvim and also some people use it as muscle memory

Alright

> * the when clauses should stay the same, as it reads "ctrl-c should close open menus only if we are in normal mode, otherwise ctrl-c should enter normal mode from insert/replace/visual"

I copied and pasted the original directly.

theol0403:
No, your mode check is inverted.

xiyaowong:
Do you mean this?
image

theol0403:
Yes. There has to be one condition that checks for normal and menus aren't open, along with a binding that checks for not normal.

theol0403:
I feel like this overcomplicates things. It is best to keep them separate and next to each other so we can see the logic. I think you should just fix the ctrlKeys for the old format.

xiyaowong:
alright, I made some changes

xiyaowong:
Or directly let ctrl+c not allow enabling and disabling through configuration. If the user wants to change it, delete it directly in keybindings. ✋

theol0403:
I prefer it like this. Users expect ctrlKeysInsert to apply to all ctrl keys bound by the extension.

xiyaowong:
Added a little comment so new contributors won't be confused looking at this

PR #1570, opened 2023-10-27
feat(api): add `vscode.to_op`

### LINKED ISSUE #1422: feature: make vscode commands dot-repeatable and work in operator-pending mode
It would be nice to make the code used for VSCodeCommentary to be modular for any command.

https://github.com/vscode-neovim/vscode-neovim/issues/1345#issuecomment-1658129459

xiyaowong:
feat: refactor `==` and `gc`, support motions and dotrepeat

close #1422

### DIFF
--- a/runtime/lua/vscode-neovim/api.lua
+++ b/runtime/lua/vscode-neovim/api.lua
@@ -1,3 +1,6 @@
+local api = vim.api
+local fn = vim.fn
+
 local M = {}
 
 ------------------------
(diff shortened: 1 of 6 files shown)


runtime/plugin/code_actions.lua
```diff
@@ -0,0 +1,46 @@
+local vscode = require("vscode-neovim")
+
+local k = function(mode, lhs, rhs)
```
theol0403, line 3: I don't think this helper provides enough value to keep users from seeing the full details in one line.

theol0403 (approved):

runtime/plugin/code_actions.lua
xiyaowong: This is concise and common

theol0403: It's just if users are supposed to use the file as a reference for the API, it should be simple and straightforward instead of needing to also copy a helper. It is nice if users can just copy/paste directly.

xiyaowong: This is the most common operation, and users can see the most critical "expr" parameter at a glance. There is no misleading anywhere. A user who just wants to copy and paste will not look at these files at all.

runtime/lua/vscode-neovim/api.lua
```diff
@@ -291,14 +291,6 @@ do
   local op_func_id = 0
 
   ---@see map-operator
-  --- vim.keymap.set(
```
theol0403: I think this usage is good to have somewhere, maybe the readme?

xiyaowong: I don't know what to write.. This function is simply encapsulated, and the details are in the vim document.
Maybe more examples can be added later.

theol0403 (approved):

PR #1571, opened 2023-10-27
fact(document_change): simplify nvim to vscode change sync

theol0403:
This continues the work from https://github.com/vscode-neovim/vscode-neovim/pull/1567. I will probably not have time to test this, so @xiyaowong feel free to take over this PR and add your own commits. 

Fixes https://github.com/vscode-neovim/vscode-neovim/issues/1068

Should also speed up editing? 

Ideas: don't use generator, don't use text.split.


theol0403:
Unfortunately, while this is still very fast, on VERY large files, it seems as though the old implementation is much faster at applying diffs caused by %s. 

Not sure if this is acceptable, or if we should improve the algorithm.

xiyaowong:
Some tests have been done, and in the case of large text and a lot of updates, at least it is faster than the terminal nvim (not clean).I tried to do some optimization, but I found that the bottleneck was in the process of diff.

xiyaowong:
> Ideas: don't use generator, don't use text.split.

They are all very cheap.

justinmk:
Even if this is a bit slower for some cases, having simpler code could encourage more improvements in the future.

xiyaowong:
I think this speed is acceptable. It takes no more than 3 seconds to use %s/this/that/g to replace 3000 places in a 20,000-line file. It takes about 200ms to perform simple addition, deletion and surrounding operations in a 300,000-line file.

I know this has to do with device performance. But these differences are only noticeable in **VERY** large files, but in this case there is no difference between very slow and slower

theol0403:
Good to merge then. Do you think the extra boilerplate for `LazyLineLengths` is worth it? The simple implementation of pre-accumulating the line lengths for the entire buffer instead of on-demand (for binary search) does have less code, and I don't know how much of a performance difference it makes.

xiyaowong:
> Good to merge then. Do you think the extra boilerplate for `LazyLineLengths` is worth it? 

At the beginning of the discussion, I originally wanted to say that I would prefer version 2, because version 3 brings little performance improvement, but the code becomes obviously "verbose". But seeing that you created this PR quickly, I withdrew my comment.

xiyaowong:
> little performance improvement

Can actually be ignored

theol0403:
I have no problem deleting unnecessary code.

xiyaowong:
In any case, this PR should be the biggest improvement in recent times👍

trkoch:
> … while this is still very fast, on VERY large files

How did you test/measure and what do you mean by very large?

(My test may be a little bit extreme and for sure unrealistic, but applying neovim edits to `sqlite3.c` [amalgamation](https://www.sqlite.org/download.html) is still >30s. If you're looking for a large file, look no further 😀 )

trkoch:
❤️

image

trkoch:
@theol0403 Verified https://github.com/vscode-neovim/vscode-neovim/issues/829 does not regress. Thanks for keeping that in mind!

PR #1574, opened 2023-10-28
chore: use "onStartupFinished" for activation

xiyaowong:
I think we should use "onStartupFinished" instead of "*"

1. Better user experience.
2. vscode-neovim needs to perform some synchronization work during initialization, which may cause issues when VSCode is restoring the workspace state(not sure, but it seems possible).

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
         "Keymaps"
     ],
     "activationEvents": [
-        "*"
+        "onStartupFinished"
     ],
     "main": "./dist/extension.js",
     "contributes": {


justinmk:
LGTM. Did the git history give any hint about why `*` was chosen?

xiyaowong:
It may be because when this extension was first created, there was no 'onStartupFinished'

https://code.visualstudio.com/updates/v1_46#_onstartupfinished-activation-event

theol0403 (approved):

Issue #1555, opened 2023-10-21
bug: ghost text in editor window when using vim search


floooh:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.9.4

### Operating system/version

macOS 14.0

### Describe the bug

Quite often when using vim text search via '/', text rendering artefacts appear in the editor when typing the search text, the artefacts are purely visual, the actual content of the edited file is not altered.

<img width="447" alt="Screenshot 2023-10-21 at 15 15 46" src="https://github.com/vscode-neovim/vscode-neovim/assets/1699414/5f39cdf9-93a4-46eb-947f-105a3bf0844a">

Interestingly, when it works as expected, the editor jumps to the first matching occurance, but this 'jumping to search location' doesn't seem to happen when it's broken, instead those artefacts show up rendered in seemingly random places.

Here's another example. Strarting to type `_sg_wgpu_` first highlighted matches that were already visible in the current editor location, but once the search text was for a match outside the currently visible section, the artefacts appear instead of jumping the function `_sg_wgpu_create_buffer` further down in the file.

The single `e` after the semicolon seems to be the last character of the search text btw and changes while typing:

<img width="643" alt="Screenshot 2023-10-21 at 15 27 30" src="https://github.com/vscode-neovim/vscode-neovim/assets/1699414/06f7a846-6a36-4693-9495-4a12077b6d9a">

If it helps, this is a pretty large source file (> 18kloc).

Here's another particularly wild example:

<img width="404" alt="Screenshot 2023-10-21 at 15 38 11" src="https://github.com/vscode-neovim/vscode-neovim/assets/1699414/fdd84ee4-457b-4073-b133-f94f3c2f10bd">

The common theme seems to be that at one point, 'jump to search location' stops working and then the artefacts appear instead.

### Steps To Reproduce

I have not yet found a reliable way to reproduce this unfortunately, neither with other VSCode extensions or not. My neovim config is vanilla. It happens often enough to be annoying though ;)

The bug is already present for a couple of months at least.

### Expected Behavior

No text rendering artifacts when searching text.

xiyaowong:
[...]

Issue #1149, opened 2023-02-04
Search replace with confirmation :%s/foo/bar/gc does not scroll automatically


trahman1318:
Having the same issue as https://github.com/vscode-neovim/vscode-neovim/issues/999

MacBook Pro (non-M1)
MacOS Monterey 12.6.3
empty init.vim
empty settings.json
no vscode extensions
NVIM v0.8.3
vscode v1.75.0
extension version v0.0.95

Tried running `%s/get_option/get_option_new/gc` on this [file](https://github.com/mongodb/mongo/blob/450ffbb63d5b265d2e17c6ec4556f030099b9f60/SConstruct) with my cursor at the top. I'm able to confirm replacements, but there is no scrolling which makes it difficult to use. Does anyone else have this issue? Or is there a workaround?

ctstacey:
Sorry, I don't have answers or workarounds at this time (others might?), but here are some clues to help trouble shoot this issue...

I can replicate with a minimal file.

Windows_NT x64 10.0.19044
VSCode 1.75.0
NVIM 0.8.1
VSCode-Neovim 0.0.97

To replicate:
1. open new file (in VSCode with VSCode-Neovim extension enabled)
2. add the letter `a` on line 2
3. add the letter `a` on line 92
4. move to line 1
5. use `%s/a/b/gc` to substitute globally (with scroll and confirm) and notice it will not scroll to the second `a` (after replacing, or not, the first `a`)

If you are successful, try varying the number of lines between the two `a`s until it fails to scroll to the second `a` (after replacing, or not, the first `a`).

It did not make any difference how much text was in the lines between the two `a`s.

I also found that moving to around line 65 meant when I performed the substitute, the extension would not scroll to the first `a` (to make the first substitution on confirmation).

LawrenceLiu023:
I have the same problem. The cursor does not follow the matches and the match waiting to be confirmed is highlighted in the same colour as the other matches, making it difficult to find the right one. Even with an empty init.vim, the problem remains unsolved.

VSCode 1.75.1
Neovim 0.8.2
VSCode-Neovim 0.0.97
Windows11 10.0.22000.1516

xiyaowong:
Similar to https://github.com/vscode-neovim/vscode-neovim/issues/1555#issuecomment-1773966732

jpmorris:
[...]

PR #1575, opened 2023-10-28
fix(incsearch): viewport is not updating during searching

xiyaowong:
- Resolve: #1149
- Resolve: #1555

### DIFF
--- a/src/viewport_manager.ts
+++ b/src/viewport_manager.ts
@@ -5,7 +5,7 @@ import { config } from "./config";
 import { EventBusData, eventBus } from "./eventBus";
 import { createLogger } from "./logger";
 import { MainController } from "./main_controller";
-import { disposeAll } from "./utils";
+import { ManualPromise, disposeAll } from "./utils";
 
 const logger = createLogger("ViewportManager");
 
@@ -27,6 +27,8 @@ export class ViewportManager implements Disposable {
      */
     private gridViewport: Map<number, Viewport> = new Map();
 
+    private syncViewportPromise?: ManualPromise;
+
     private get client() {
         return this.main.client;
     }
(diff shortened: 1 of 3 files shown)


src/viewport_manager.ts
```diff
@@ -87,6 +93,27 @@ export class ViewportManager implements Disposable {
                         view.line = curline;
                         view.col = curcol;
                     }
```
justinmk, line 95: off-topic, line 83 (89) looks weird, is it intentional?

                        for (const [grid, , topline, botline, curline, curcol] of args) {

xiyaowong: Yes, currently the second item(win) is not needed

theol0403:
This feels like a hack, and I don't really like the added code. Is there any other possible fix that is more generic?

xiyaowong:
Neovim did not send the corresponding grid information during incsearch, and no relevant instructions were found in the neovim documentation. @justinmk  

This 'hack' is currently the simplest one among the solutions I have tried.

theol0403:
It used to work before your highlight manager refactor... do you know why it broke?

xiyaowong:
No, this is a problem that has always existed. I tested 0.3.2 💆🏻‍♂️

It is fundamentally unrelated to highlighting.

https://github.com/vscode-neovim/vscode-neovim/issues/1149#issuecomment-1427386558

theol0403 (approved):

xiyaowong:
This problem can only be easily reproduced in clean mode. Most people(including me) don't encounter this problem because they use a search plugin or execute nohl after searching

xiyaowong:
https://github.com/vscode-neovim/vscode-neovim/pull/1453

xiyaowong:
https://github.com/neovim/neovim/pull/27858

Issue #1534, opened 2023-10-15
bug: fast typing impossible due to race condition / performance


ogwhic:
Edit: The "race condition"/"poor performance" in the title is just my best guess at what's going on. I really have no idea...

'### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.9.2

### Operating system/version

Windows 11

### Describe the bug

Unable to determine if this is a vscode-neovim specific issue, code completion or combination of the two.

I have both NeoVim and code completion/intellisense/whatever via the C# extension enabled. I'm not sure exactly what's happening but when typing quickly the autocomplete UI appears and disappears and some combination of NeoVim action and autocomplete action both trigger in response to pressing return or esc or some other key.

The result is often the code I've just typed being inserted twice with some corrupt attempt at autoformatting or just random code from the autocomplete being vomited into the text editor.

I'm spending more time correcting the damaged caused than actual coding.

Because it only happens when I'm typing quickly it's difficult to figure out exactly what's going wrong. Just code vomit everywhere.

It "feels" like some kind of race condition/fight is taking place between neovim and the autocomplete, with both being triggered by the same key press (e.g., return to start a new line being hijacked by autocomplete incorrectly, or an internal buffer of neovim being dumped multiple times into the document).

### Steps To Reproduce

1. Install VSCode
2. Install NeoVim
3. Install vscode-neovim
4. Install C# Dev Kit extension
5. Disable all but the 'C#' extension among those installed by step 4 (optional)
6. Try touch typing code in a C# file
7. Observe a combination of random code and duplicate copies of what you just typed being dumped into the editor as you type

### Expected Behavior

Text I type and intentional autocomplete/snippets/whatever should be inserted into the editor. Random code and duplicate copies of what I type should not.

theol0403:
Not sure what's going on... vscode-neovim should really not be doing anything during insert mode. My only guess is somehow the document change sync is stepping on your typing.

c0z0:
I'm also seeing this. Sometimes changes made in insert mode are duplicated. I couldn't manage to find a way to reliably reproduce it though. It happens about once every five inserts.

fpiemontese-dku:
[...]

PR #1581, opened 2023-10-30
fix(dotrepeat): incorrect order of deletion and addition

### LINKED ISSUE #1580: bug: dotrepeat not work when deleting before adding
dotrepeat

xiyaowong:
- Resolve #1580
- Refactor using lua: 
	- More concise
	- Faster

- Might have alleviated #1534 as it's faster now


theol0403 (approved):
Hooray for deleting bad code 🎉

PR #1596, opened 2023-11-07
fix(blockwise-visual): handle double-width chars and tabs

xiyaowong:
improvement: the selection direction is consistent with nvim

Issue - Fixed

### DIFF
--- a/runtime/lua/vscode-neovim/internal.lua
+++ b/runtime/lua/vscode-neovim/internal.lua
@@ -115,4 +115,51 @@ do
   end
 end
 
+---display-col to col
+---@param line string
+---@param discol number
+local function discol_to_col(line, discol)
+  if discol == 0 then
+    return 0
+  end
+
+  local col = 0
+  local max_col = vim.fn.strcharlen(line)
+
+  local min_distance = math.huge
+  local nearest_col = 0
+
+  while col <= max_col do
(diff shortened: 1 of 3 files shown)


trkoch:
@xiyaowong Not sure I can follow the GIF. The issue you're fixing is that some characters are not selected which should?

theol0403:
Isn't there a built in vim function to convert display col to col? Could this PR not be simplified to just converting the column before sending it to VS code? This code seems a little complex.

xiyaowong:
> Isn't there a built in vim function to convert display col to col? 

Oh! I forgot to search for the built-in function. There is `virtcol2col`

xiyaowong:
@theol0403 Not sure if `virtcol2col` can simplify the code, after all, we still need to convert byte-index to char-index. And this function depends on a window

theol0403:
Isn't there also a function to convert char to byte char?

xiyaowong:
vim.str_utfindex

I'll refactor it when I have time. But, using multiple built-in functions is not necessarily more concise.

theol0403:
What about https://neovim.io/doc/user/builtin.html#charidx()

xiyaowong:
It seems almost the same

vim.str_utfindex is widely used in nvim-lsp. Maybe it's better?. I'm not sure. I haven't used any of these functions.

PR #1597, opened 2023-11-07
fix(visual): optimized simulation of VisualChanged event

xiyaowong:
Rework "viw" on last symbol

Discovered this minor issue while debugging the tests of the previous PR and thought of this idea. So far, it seems to be working fine.

### DIFF
--- a/runtime/lua/vscode-neovim/cursor.lua
+++ b/runtime/lua/vscode-neovim/cursor.lua
@@ -1,4 +1,4 @@
-local api = vim.api
+local api, fn = vim.api, vim.fn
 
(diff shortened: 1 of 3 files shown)


theol0403:
This might be better to fix as a simple PR to neovim to add visual changed. However, we can merge this for now. 

This also means we can ignore the visual highlight.

xiyaowong:
I have no experience with Neovim's internal code, and it will take a lot of time to contribute. Hope someone is willing to do this part of the work. At this time, it seems that this is not a priority for the Neovim team...

PR #1601, opened 2023-11-11
fix: rework `:tabnew`

### LINKED ISSUE #1548: Feature request: `tabnew` support
### Did you check the docs?

The vim command `tabnew` always opens a new blank document, regardless of working directory.

e.g. I'm in `~/dev/dwm` (verified that neovim agrees with this by running `:pwd`), and I type `:tabnew dwm.c` (a file which already exists). vscode-neovim will open a new, blank document instead of the file that I'm expecting to edit.

I would prefer if the behavior were similar to how vim/nvim works on the command line. Namely, that it opens the document I indicate.

I understand that it could be an issue, since vim is happy to have a buffer open with a handle to a file that doesn't exist, and vscode might not be, but I'm sure there's a way to bodge around that. A dirty hack to catch files which aren't already touched is preferable to not being able to use `:tabnew` at all.

n/a

### Additional context

_No response_

ElamC:
Fixes #1548

Rework `tabnew` to correctly replicate behaviour. For now an active workspace is required to provide a directory if the file is saved.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -9,6 +9,7 @@ import {
     Disposable,
     EndOfLine,
     EventEmitter,
+    NotebookDocument,
     Selection,
     TextDocument,
     TextDocumentContentProvider,
@@ -251,7 +252,7 @@ export class BufferManager implements Disposable {
     private async handleOpenFile(data: EventBusData<"open-file">) {
         const [fileName, close] = data;
         const currEditor = window.activeTextEditor;
-        let doc: TextDocument | undefined;
+        let doc: NotebookDocument | TextDocument | undefined;
         try {
             if (fileName === "__vscode_new__") {
                 doc = await workspace.openTextDocument();
@@ -261,9 +262,10 @@ export class BufferManager implements Disposable {
                 try {
                     await workspace.fs.stat(uri);
                 } catch {
-                    uri = Uri.from({ scheme: 'untitled', path: uri.path });
+                    uri = Uri.from({ scheme: "untitled", path: normalizedName });
+                    doc = await workspace.openNotebookDocument(uri);
                 }
-                doc = await workspace.openTextDocument(uri);
+                doc ??= await workspace.openTextDocument(uri);
             }
         } catch (error) {
             logger.error(`Error opening file ${fileName}, ${error}`);
@@ -276,7 +278,7 @@ export class BufferManager implements Disposable {
             viewColumn = currEditor.viewColumn;
             await commands.executeCommand("workbench.action.revertAndCloseActiveEditor");
         }
-        await window.showTextDocument(doc, viewColumn);
+        await window.showTextDocument(<TextDocument>doc, viewColumn);
         if (close === "all") {
             await commands.executeCommand("workbench.action.closeOtherEditors");
         }
@@ -718,7 +720,7 @@ export class BufferManager implements Disposable {
 
     private findPathFromFileName(name: string): string {
         const folders = workspace.workspaceFolders;
-        if (folders) {
+        if (folders && folders.length > 0) {
             return path.resolve(folders[0].uri.fsPath, name);
         } else {
             return name;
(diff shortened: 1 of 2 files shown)


theol0403 (commented):
Does this suppress the error that the path is incorrect?

ElamC:
> Does this suppress the error that the path is incorrect?

No path errors unless the command is used outside of a workspace. Also, saving the file doesn't prompt for a path [it just saves it](https://github.com/microsoft/vscode/issues/93441#issuecomment-604870486)

xiyaowong:
@ElamC  Good, but if you have time, please fix the failed tests.

src/buffer_manager.ts
```diff
@@ -251,14 +252,20 @@ export class BufferManager implements Disposable {
...
             } else {
                 const normalizedName = fileName.trim();
-                const filePath = this.findPathFromFileName(normalizedName);
-                doc = await workspace.openTextDocument(filePath);
+                let uri = Uri.parse(this.findPathFromFileName(normalizedName));
+                try {
+                    await workspace.fs.stat(uri);
+                } catch {
+                    uri = Uri.from({ scheme: "untitled", path: normalizedName });
+                    doc = await workspace.openNotebookDocument(uri);
```
xiyaowong, line 266: Why notebook?

ElamC: [Limitations](https://github.com/microsoft/vscode/issues/197836) with `TextDocument`, specifically when there is no active workspace. `openNotebookDocument` prompts for a file path when the document is saved and while it returns a `NotebookDocument` it can still be [used](https://github.com/microsoft/vscode/blob/e1a8566a298dcced016d8e16db95c33c270274b4/src/vs/vscode.d.ts#L10764) as a `TextDocument`

vim/vscode-tab-commands.vim
```diff
@@ -12,7 +12,7 @@ function! s:gotoEditor(...) abort
 endfunction
 
 command! -complete=file -nargs=? Tabedit if empty(<q-args>) | call VSCodeNotify('workbench.action.quickOpen') | else | call VSCodeExtensionNotify('open-file', expand(<q-args>), 0) | endif
-command! -complete=file -nargs=? Tabnew call VSCodeExtensionNotify('open-file', '__vscode_new__', 0)
+command! -complete=file -nargs=? Tabnew call VSCodeExtensionNotify('open-file', empty(<q-args>) ? '__vscode_new__' : expand(<q-args>), 0)
```
justinmk, line 15: q-args will add quotes to the name, should this be `<args>` ?
```suggestion
command! -complete=file -nargs=? Tabnew call VSCodeExtensionNotify('open-file', empty(<q-args>) ? '__vscode_new__' : expand(<args>), 0)
```

src/buffer_manager.ts
justinmk: That's quite a hack. The code should have a brief comment explaning this, with a URL to the upstream issue.

src/buffer_manager.ts
```diff
@@ -251,14 +252,20 @@ export class BufferManager implements Disposable {
...
-        let doc: TextDocument | undefined;
+        let doc: NotebookDocument | TextDocument | undefined;
         try {
             if (fileName === "__vscode_new__") {
                 doc = await workspace.openTextDocument();
             } else {
                 const normalizedName = fileName.trim();
-                const filePath = this.findPathFromFileName(normalizedName);
-                doc = await workspace.openTextDocument(filePath);
+                let uri = Uri.parse(this.findPathFromFileName(normalizedName));
```
xiyaowong, line 261:
```suggestion
                let uri = Uri.from({ scheme: "file", path: this.findPathFromFileName(normalizedName) });
```

xiyaowong (changes requested):

src/buffer_manager.ts
xiyaowong: Uri.parse cannot correctly recognize Windows paths, and the drive letter will be treated as a schema.

PR #1603, opened 2023-11-13
refactor(blockwise-visual): use builtin functions

xiyaowong:


xiyaowong:
The testing of Windows always fails inexplicably, but it runs normally when tested locally. 🤔🤨😕

theol0403:
You can't await lua code when processing regular cursor movements, it adds too much lag. Only await in visual mode. The cursor position should still be pushed to viewport_manager, not pulled from getpos.

xiyaowong:
I know this. I've tested it. There's no difference. A single request takes only 1 ms

theol0403:
It felt to me like there was more jitter than usual when the computer was under load.

theol0403 (approved):

PR #1605, opened 2023-11-13
refactor: Add "actions.lua" for executing internal lua functions

xiyaowong:


src/actions.ts
```diff
@@ -59,6 +59,15 @@ class ActionManager implements Disposable {
         this.client.executeLua('require"vscode-neovim.api".fire_event(...)', [event, ...args]);
     }
 
+    /**
+     * Execute a function from the Lua module `vscode-neovim.internal`.
+     * @param fname the internal function name
+     * @param args arguments
+     */
+    public async internalLua<T = any>(fname: string, ...args: VimValue[]): Promise<T> {
```
justinmk, line 67: "internal" to what? can we just call it `lua` ?
```suggestion
    public async lua<T = any>(fname: string, ...args: VimValue[]): Promise<T> {
```

theol0403 (approved):

PR #1620, opened 2023-11-15
docs: declutter keybindings reference

justinmk:
Problem:
The docs over-explain key bindings and command aliases, which
1. is a maintenance burden
2. makes it harder to find useful info in the readme

Solution:
- Show users how to use vscode features to find keybindings instead of explaining the full details of every mapping.
- Commands like `:write` and `:split` are expected to work as well as possible, there's no need to explain the low-level details of each case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,12 @@
 <a href="https://gitter.im/vscode-neovim/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge"><img src="https://badges.gitter.im/vscode-neovim/community.svg"></a>
 </p>
 
-[Neovim](https://neovim.io/) is a fork of VIM to allow greater extensibility and integration. This extension uses a
-fully embedded Neovim instance, no more half-complete VIM emulation! VSCode's native functionality is used for insert
+[Neovim](https://neovim.io/) is a fork of Vim to allow greater extensibility and integration. This extension uses a
+fully embedded Neovim instance, no more half-complete Vim emulation! VSCode's native functionality is used for insert
 mode and editor commands, making the best use of both editors.
 
--   🎉 Almost fully feature-complete VIM integration by utilizing Neovim as a backend.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -654,15 +614,15 @@ Enabled by `ctrlKeysForInsertMode`
 
 Default: `["a", "d", "h", "j", "o", "r", "t", "u", "w"]`
 
-Refer to VIM's manual for their use.
+> 💡 Run the `Preferences: Open Keyboard Shortcuts` vscode command and search for "neovim" to see all keybindings.
```
theol0403, line 617: That will only tell you that it's passed down to neovim, the meaning of the actual binding lives in the vim manual.

justinmk: can add another note to mention that.

but in general: why is someone using a vim extension if they don't know how to use vim?

theol0403: I think I am an example of that - I was using VS code, and then wanted to learn how to use vim. However, I missed a lot of VScode's features, like out of the box LSP and the UI. So actually, I learned how to use vscode-neovim before I learned how to use terminal vim, so the reference actually acted as a convenient cheat sheet for me.

theol0403 (approved):

README.md
justinmk: will add a note that mentions `:help` and also links to relevant parts of https://neovim.io/doc/user/index.html

xiyaowong:
Take the opportunity to format contributing.md.
image

xiyaowong (approved):

theol0403:
@justinmk I'll merge and release 1.0.2 once you push your last tweaks

justinmk:
WIll do

xiyaowong:
Suggest adding this to show how to remove a `keybinding`. `keybindings.json` needs to be maintained, and users are unwilling to learn!
key

PR #1622, opened 2023-11-16
feat: Add command "vscode-neovim.lua"

xiyaowong:
improve: Make vscode-neovim.stop can be executed multiple times

1. Customizing the when context to make keybindings more flexible.
2. Executing multiple commands.
3. ...

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -4,7 +4,7 @@ import { readFile } from "fs/promises";
 import path from "path";
 
 import { attach, NeovimClient } from "neovim";
-import vscode, { Range } from "vscode";
+import vscode, { Range, window } from "vscode";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { transports as loggerTransports, createLogger as winstonCreateLogger } from "winston";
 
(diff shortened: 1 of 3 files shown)


theol0403 (approved):
Can you think of a better place to put the lua handler? Maybe in commands manager or typing manager? It would be nice to have in the same place as nvim.input.

xiyaowong:
I personally think it would be more appropriate to put it in main_controller. commands_manager is not suitable and will be removed in the future.

src/main_controller.ts
```diff
@@ -157,6 +157,17 @@ export class MainController implements vscode.Disposable {
...
             vscode.commands.registerCommand("_getNeovimClient", () => this.client),
+            vscode.commands.registerCommand("vscode-neovim.lua", async (lua) => {
+                if (!lua) {
+                    window.showWarningMessage("No lua code provided");
+                    return;
+                }
+                try {
+                    await this.client.lua(lua);
+                } catch (e) {
+                    logger.error(e instanceof Error ? e.message : e);
```
justinmk, line 168: We should add a `vscode.commands.registerCommand` wrapper that all commands use, which could handle errors and show/log them.

xiyaowong: It's unnecessary. Handling errors or not is the responsibility of the handler.

justinmk: Yes, but at the "top level" in most cases the handling is the same: log it and show a message. That avoids lots of redundant try/catch deeper in the codepath.

When code actually has something useful to do with an error, it of course can use a try/catch.

PR #1633, opened 2023-11-18
fix(highlight): Do not clear custom highlight groups directly

### LINKED ISSUE #1631: bug: yank highlighting broke in v1.1.1, but works in v1.1.0

### Neovim version (nvim -v)

v0.9.4

### Operating system/version

MacOS 14.0

### Describe the bug

Thanks for the awesome integration, i'm new to vim so having an actual nvim instance integrated with the vscode is really helpful!

Yank highlighting broke after updating to v1.1.1, at first i thought that the issue was in the config files as i'm new to neovim and could brake something myself, tried different configs to get highlighting to work:
Vim script:
```vim
augroup highlight_yank
    autocmd!
    autocmd TextYankPost * silent! lua vim.highlight.on_yank { higroup="IncSearch", timeout=1000 }
augroup END
```
And also tried Lua:
```lua
local function augroup(name)
  return vim.api.nvim_create_augroup("lazyvim_" .. name, { clear = true })
end

vim.api.nvim_create_autocmd("TextYankPost", {
  group = augroup("highlight_yank"),
  callback = function()
    vim.highlight.on_yank()
  end,
})
```

Both configs worked just fine inside the nvim terminal, but not with `vscode-extension`, although i had it working in vscode too just today earlier. The configuration file does not contain anything else, I also tried on a fresh installation of VSCode Insiders without any other extensions - still no success. Downgrading to v1.1.0 fixed the problem.

### Steps To Reproduce

Install the v1.1.1 version.
Enable yank highlighting in the config file.

### Expected Behavior

Yank should be highlighted, just as it was in v1.1.0

xiyaowong:
Fixes #1631

The most ideal solution is to set all highlights in nvim, but it is difficult to do without the vscode API, unless we give up using the VSCode's ThemeColor. Please vote for https://github.com/microsoft/vscode/issues/32813

### DIFF
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -1,7 +1,9 @@
 ---@diagnostic disable: inject-field
 
 -- Copy global highlights and overrides highlights to the custom namespace, only external buffers use global namespace
+
 local api = vim.api
+local vscode = require("vscode-neovim.api")
 
 local NS = api.nvim_create_namespace("vscode-neovim-highlight")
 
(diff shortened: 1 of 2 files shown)


justinmk:
I haven't been paying much attention to these highlight-related PRs, but I as usual I hope we aren't adding complexity to support every possible config and plugin.

Is there a constraint we can declare that would vastly simplify the code? For example, could we say "we don't support non-standard highlights", or "we don't support plugin-defined highlights", or some other constraint?

If we always showed Nvim highlights as a single (configurable) color, how much code and complexity would that eliminate?

xiyaowong:
> to support every possible config and plugin.

The current requirement is generic and not specific to any particular configuration or plugin.

> Is there a constraint we can declare that would vastly simplify the code? For example, if we always showed Nvim highlights as a single (configurable) color, how much code and complexity would that eliminate?

The main issues with highlighting currently are: 
1. We need highlighting, but not all of it. 
2. We need to use vscode's theme colors for custom highlighting, but without affecting other highlights that are derived from the combination of custom highlighting.

theol0403 (approved):

justinmk:
> > to support every possible config and plugin.
> 
> The current requirement is generic and not specific to any particular configuration or plugin.

It's specific to Nvim plugins that use highlighting, which is only a subset, and not an important one.

Can't we use a hardcoded (minimally configurable) highlight for any highlights that come from Nvim? Would that simplify things?

PR #1643, opened 2023-11-20
fix(highlight): remove duplicate visual selection highlight

theol0403:
With https://github.com/vscode-neovim/vscode-neovim/pull/1597, applying the `Visual` highlight is no longer needed, we can entirely rely on vscode's selection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -339,9 +339,6 @@
                         "Search": {
                             "backgroundColor": "theme.editor.findMatchHighlightBackground",
                             "borderColor": "theme.editor.findMatchHighlightBorder"
-                        },
-                        "Visual": {
-                            "backgroundColor": "theme.editor.selectionBackground"
                         }
                     }
                 },
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -52,6 +52,7 @@ local function setup_globals()
     Normal       = {},
     NormalNC     = {},
     NormalFloat  = {},
+    Visual       = {},
     VisualNC     = {},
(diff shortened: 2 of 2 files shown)


xiyaowong:
There are still some problems when multi-byte characters exist, especially when emoji exist. Might be better to keep Visual as feedback

theol0403:
What kind of problems? Wouldn't it also cause cursor position problems? Not sure when the highlight will be more accurate than cursor position.

xiyaowong:
It’s not just a matter of cursor position, it’s also related to highlighting. Showing highlights from nvim gives a more accurate representation of the selections

image

xiyaowong:
However, most people do not use multi-byte characters, so deleting Visual is not a big deal.

In fact, Visual also has a potential function, that is, when the highlighting (or extmarks) is wrong, you can clear the wrong highlighting by selecting the area :joy:

xiyaowong:
Weird, I can't reproduce the previous bug again. Anyway, the main issue now is that it occurs when emojis are present. It can be fixed later when someone provides feedback.

xiyaowong (approved):

xiyaowong:
btw, there is a part of the logic here specifically for handling Visual. It is recommended to keep it.

https://github.com/vscode-neovim/vscode-neovim/blob/6b58cd374eb4f08a0f0ce6bc247a320396ca19b2/src/highlight_provider.ts#L153-L155

theol0403:
Does clearing the visual highlight in this PR affect that?

xiyaowong:
Some users may set up Visual by themselves.

Issue #1651, opened 2023-11-23
bug: extension attempts to create a new buffer for the editor when the buffer for the editor already exists


archilkarchava:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.9.4

### Operating system/version

MacOS 14.1.1

### Describe the bug

"nvim_buf_set_name: Failed to rename buffer" error is shown when selecting an item from the reference search list.
It happens because VS Code creates an editor twice.
First, a new editor is created when the file is opened but the reference hasn't been selected yet. The extension will create a neovim buffer by calling `nvim_buf_set_name` with the following arguments `[bufId, BUFFER_NAME_PREFIX + document.uri.toString()]`
After a reference is selected VS Code will dispose of the existing editor for the file and create a new one. The extension will first create a new buffer and will call the `nvim_buf_set_name` with the same arguments which will result in the "nvim_buf_set_name: Failed to rename buffer" error and only then the buffer for the disposed editor will be removed.

### Steps To Reproduce

1. Open a multi-file code base that has references.
2. Select a function/class that is referenced in multiple files and press "Shift+f12".
3. Focus a reference in a list without pressing enter in a file that was not opened since the extension was initialized to make sure that the new buffer will be created for the editor.
4. Press "Enter" to select a reference.

https://github.com/vscode-neovim/vscode-neovim/assets/16840190/f8188db1-2e18-4eab-9422-21c6a02a7fba

### Expected Behavior

The reference search is closed and a new neovim buffer is created for the editor with the name set to `BUFFER_NAME_PREFIX + document.uri.toString()`

stimw:
get this problem again...

uduse:
I still bump into this problem very often. @stimw did you try something to fix it?

stimw:
> I still bump into this problem very often. @stimw did you try something to fix it?

I found that I changed the name of the nvim configuration file, but did not synchronize the change in vscode.

PR #1652, opened 2023-11-23
fix(buffer): bufname conflict when creating a new buffer

archilkarchava:
Remove buffers for disposed editors before creating new buffers to avoid naming conflicts.

Fixes https://github.com/vscode-neovim/vscode-neovim/issues/1651

### Before the fix
https://github.com/vscode-neovim/vscode-neovim/assets/16840190/f8188db1-2e18-4eab-9422-21c6a02a7fba

### After the fix
https://github.com/vscode-neovim/vscode-neovim/assets/16840190/3bef4d45-a0fa-4b5e-a9bd-bfce1b606c9f

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -463,6 +463,29 @@ export class BufferManager implements Disposable {
         // store in copy, just in case
         const currentVisibleEditors = [...window.visibleTextEditors];
 
+        logger.debug(`Clean up windows and buffers`);
+        const unusedWindows: number[] = [];
+        const unusedBuffers: number[] = [];
+        // close windows
+        [...this.textEditorToWinId.entries()].forEach(([editor, winId]) => {
+            if (!currentVisibleEditors.includes(editor)) {
+                logger.debug(`Editor viewColumn: ${editor.viewColumn}, winId: ${winId}, closing`);
+                this.textEditorToWinId.delete(editor);
+                this.winIdToEditor.delete(winId);
+                unusedWindows.push(winId);
+            }
+        });
+        // delete buffers
+        [...this.textDocumentToBufferId.entries()].forEach(([document, bufId]) => {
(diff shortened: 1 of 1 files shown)


xiyaowong (approved):

xiyaowong:
nice, I also encountered this error recently

xiyaowong:
Tested, it works.

fiddler:
This seems to break <c-^> - so jumping between the two last open buffers stopped working after installing 1.1.4. Installing 1.1.3 makes <c-^> work again normally. However, I'm not sure if it's related to my setup. I've configured it with `nnoremap <Leader><Leader> <c-^>` in my init.vim. Does this happen for you guys as well?

xiyaowong:
@fiddler You mean <C-^> was normal before?

fiddler:
Yup! It works for me on 1.1.3, but after updating to 1.1.4, it no longer works. It looks like it jumps inside the buffer instead of jumping to the other buffer - see the attached video.

https://github.com/vscode-neovim/vscode-neovim/assets/477113/5e591155-8984-4aac-929b-f34947a29210

PS. I noticed that directly using <c-^> in my VS Code does not work at all, but using it via the `nnoremap <Leader><Leader> <c-^>` binding works fine on 1.1.3 as shown in the video.

xiyaowong:
Not sure. From the internal implementation of vscode-neovim(The management and switching of window and buffer should be performed by vscode), C-^ should not work properly. It was a coincidence that it worked normally before.

---

Possibly relevant code:
https://github.com/vscode-neovim/vscode-neovim/blob/50e3692ec40fe9edb42f6bef02f5de23dfe607f0/vim/vscode-neovim.vim#L101
https://github.com/vscode-neovim/vscode-neovim/blob/50e3692ec40fe9edb42f6bef02f5de23dfe607f0/src/buffer_manager.ts#L306-L310

PR #1670, opened 2023-12-03
feat(api): support passing a map-like table as args

xiyaowong:
Most commands either require multiple arguments or only require one object parameter. So when args is a map-like table, it is automatically treated as the first argument, which is more convenient and easier to understand.


xiyaowong:
Only manual testing

PR #1678, opened 2023-12-05
refactor: check required version in nvim

xiyaowong:


theol0403:
Is there an old enough <5 version that this won't work for?

xiyaowong:
not sure what you mean

theol0403:
Are there any nvm APIs used to get to the point where the message is shown (lua loading, API communication, vim.notify, etc) that doesn't exist in an older version? 

The old approach worked for any version of nvim. Now we gotta assume that nvim communication works.

xiyaowong:
Oh, I understand what you mean now.

runtime/modules/check_version.lua
```diff
@@ -0,0 +1,16 @@
+local MIN_VERSION = "0.9.0"
+
+vim.api.nvim_create_autocmd({ "UIEnter" }, {
+  group = vim.api.nvim_create_augroup("VSCodeCheckVersion", {}),
+  callback = function()
+    local cur = vim.version()
+    local min = vim.version.parse(MIN_VERSION)
```
justinmk, line 7: To support this on older Nvim versions, just check if `vim.version.parse` exists. And if it doesn't, use the old vimscript api to call `vscode.window.showErrorMessage` (or send a request manually via `rpcnotify()`).

justinmk:
This might still be worth doing, see above.

theol0403:
I do like the code reduction if it's possible

vim/vscode-neovim.vim
```diff
@@ -8,6 +8,24 @@ let &runtimepath = &runtimepath . ',' . s:currDir . '/vim-altercmd'
...
+if vim.version then
+  local cur = vim.version()
+  local min = vim.version.parse(MIN_VERSION)
+  outdated = vim.version.lt(cur, min)
+end
+if outdated then
+  local msg = "Please upgrade your nvim! This extension requires nvim version "
+    .. MIN_VERSION
+    .. " or higher, and it is recommended to use the "
+    .. "[latest stable version](https://github.com/neovim/neovim/releases/tag/stable)."
```
justinmk, line 24:
```suggestion
  local msg = "vscode-neovim requires nvim version "
    .. MIN_VERSION
    .. " or higher. Install the [latest stable version](https://github.com/neovim/neovim/releases/tag/stable)."
```

justinmk (approved):

vim/vscode-neovim.vim
```diff
@@ -8,6 +8,23 @@ let &runtimepath = &runtimepath . ',' . s:currDir . '/vim-altercmd'
 let s:runtimePath = fnamemodify(s:currDir, ':h') . '/runtime'
 let &runtimepath = &runtimepath . ',' . s:runtimePath
 
+lua << EOF
+local MIN_VERSION = "0.9.0"
+
+local outdated = not vim.version
```
justinmk, line 14: `vim.version` existed on older versions which didn't have `vim.version.parse`
```suggestion
local outdated = not (vim.version and vim.version.parse)
```

PR #1679, opened 2023-12-05
fix: fix sourcing our runtime files

xiyaowong:
Problem: The source command can only process one file at a time

Note: Although we have added the "runtimepath", some package managers may clean up the "runtimepath", so files in "plugin/" will not be automatically sourced. Therefore, they need to be sourced manually. However, if the user does not use a package manager, these files will be sourced twice.


justinmk:
> some package managers may clean up the "runtimepath", so files in "plugin/" will not be automatically sourced. 

Er... we shouldn't be adding workarounds for that kind of thing, especially if it means everyone else now suffers a "source everything twice" cost. If 'runtimepath' is broken, can we detect that and show an error?

xiyaowong:
> If 'runtimepath' is broken, can we detect that and show an error?

This is not an error, some plugin managers will take over the runtimepath, such as lazy.nvim

PR #1685, opened 2023-12-08
feat(keybindings): add keybindings for output panel

xiyaowong:


scripts/keybindings/6_vscode_integration.cjs
```diff
@@ -1,98 +1,68 @@
...
+}
+
+// - Why do we need to manually send keys?
+//    The output panel is not a "real" text editor, it cannot receive regular keyboard inputs.
+// - Why do we send these keys?
+//    We send certain basic keys for actions like moving, selecting, and copying.
+//    This allows us to easily view the output content and copy it, such as error messages.
+for (const item of [
+    // motion
+    "h",
```
theol0403, line 40: may as well send ALL `type` events (A-Z, 0-9, symbols, bspc/del/left/right/up/down/tab)

xiyaowong: The output is read-only

theol0403: Something like `yip` might still want to be supported though. Is there a disadvantage to supporting all keys?

theol0403: Or `v]ip"+y`.

xiyaowong: Need to avoid any key that may modify buffer content and enter insert mode. Because the output is read-only in vscode, changes made in nvim cannot be applied to vscode.

theol0403: I thought the output buffer was readonly in vim as well.

xiyaowong: This use case is not very common. Basic movement and copying are already sufficient

theol0403: Fair enough. I just want to reduce edge cases/inconsistency when possible.

theol0403: And things like support custom textobjects for copying. (Maybe someone has a url textobj they frequently use with `u`).

xiyaowong: This use case itself is an "edge case". I also want to be able to quickly copy using the textobject, but considering if more keys are added, it may lead to more "non-value" problems due to user-defined mappings in nvim.

theol0403: Again, can't we just set the output to readonly in nvim?

xiyaowong: Cannot. Because we need to apply vscode changes to nvim

xiyaowong: But I have thought of a solution, which is to temporarily modify the readonly and modifiable options of the buffer when applying vscode changes.

theol0403: I was about to suggest that!

theol0403:
Looks great. For completeness, should we also include 0-9 and also every non-printing key like bspc and del? I know, for example, that lightspeed.nvim uses the backspace key for "search repeat".

xiyaowong:
Definitely enough for this edge case :eyes:

theol0403 (approved):

PR #1688, opened 2023-12-09
feat: Add clipboard provider `g:vscode_clipboard`

### LINKED ISSUE #103: hook vscode clipboard api in neovim
will allow clipboard interop in wsl mode
`:help g:clipboard`

xiyaowong:
Use the VSCode clipboard by default in WSL

Close #103

### DIFF
--- a/runtime/modules/clipboard.lua
+++ b/runtime/modules/clipboard.lua
@@ -33,6 +33,7 @@ vim.g.vscode_clipboard = {
     ["+"] = paste,
     ["*"] = paste,
   },
+  cache_enabled = 0,
 }
 
 -- The user also can override g:clipboard in the init config
--- a/runtime/modules/clipboard.lua
+++ b/runtime/modules/clipboard.lua
@@ -1,0 +1,41 @@
+local code = require("vscode-neovim")
+
+local last_item = nil
+
+local function paste()
+  local curr_text = code.call("clipboard_read"):gsub("\r\n", "\n")
+  local curr_item = { vim.split(curr_text, "\n"), "v" }
+
+  if not last_item then
+    return curr_item
+  end
+
+  if not vim.deep_equal(last_item[1], curr_item[1]) then
+    return curr_item
+  end
+
+  return last_item
+end
+
+local function copy(lines, regtype)
+  last_item = { lines, regtype }
+  local text = table.concat(lines, "\n")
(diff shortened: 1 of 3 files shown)


theol0403:
Please add to docs next to vim.notify.

xiyaowong:
Maybe when vscode-neovim runs on the remote extension host, it also uses the vscode clipboard? I think most people would like that.

theol0403:
I like that!

xiyaowong:
I just tried to use it remotely, but unfortunately, it doesn't work. The API used for extension is related to the extension host, so when vscode-neovim runs on a remote extension host, it uses the clipboard of the system where the vscode server is located. :(

theol0403 (approved):

runtime/modules/clipboard.lua
```diff
@@ -0,0 +1,42 @@
...
+  return last_item
+end
+
+local function copy(lines, regtype)
+  last_item = { lines, regtype }
+  local text = table.concat(lines, "\n")
+  code.call("clipboard_write", { args = { text } })
+end
+
+vim.g.vscode_clipboard = {
```
justinmk, line 26: Don't we have a "vscode" Lua moduel? Why define a new `g:` variable? Instead we can document something like:

    vim.g.clipboard = require('vscode').clipboard

justinmk: Are we not deprecating the vimscript interface? It is confusing to have both.

> it is not convenient to set it in Vimscript if it is exported as a Lua field.

Setting it in vimscript is: 

    let g:clipboard = v:lua.require('vscode-neovim').clipboard

, which we can document. Does it need to be more convenient than that?

`require('vscode-neovim')` is awkward to type, but that is a general issue unrelated to this specific case. (I don't remember why `'vscode-neovim'` was chosen as the module name instead of 'vscode')

theol0403: Someone complained that it collided with their existing config and/or "vscode" theme. In hindsight, we should have told them tough luck.

xiyaowong: > was chosen as the module name instead of `'vscode'`)

This is what I've always wanted and was discussed while working on 1.0

theol0403: We could potentially make an alias to avoid annoyance.

PR #1689, opened 2023-12-09
fix: set correct cwd for WSL

### LINKED ISSUE #520: Working directory not set correctly with WSL installation
Right now, `:pwd` gives me `/mnt/c/<user>/AppData/Local/Programs/Microsoft VSCode` in every document and workspace. I think the working directory should actually be the current workspace, to match the behavior of VSCode's integrated terminal.

I'm using a Neovim installation on WSL. The working directory is the same regardless of whether I open a file in WSL or Windows.

xiyaowong:
Close #520

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import { ChildProcess, execSync, spawn } from "child_process";
 import { readFile } from "fs/promises";
 import path from "path";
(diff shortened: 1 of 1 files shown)


theol0403 (approved):

src/main_controller.ts
```diff
@@ -85,9 +87,9 @@ export class MainController implements vscode.Disposable {
         ];
 
         const workspaceFolder = vscode.workspace.workspaceFolders;
-        const cwd = workspaceFolder && workspaceFolder.length ? workspaceFolder[0].uri.fsPath : undefined;
-        if (cwd && !config.useWsl && !vscode.env.remoteName) {
-            args.push("-c", `cd ${cwd}`);
+        const cwd = workspaceFolder?.length ? workspaceFolder[0].uri.fsPath : undefined;
+        if (cwd && !vscode.env.remoteName) {
+            args.push("-c", `cd ${config.useWsl ? wslpath(cwd) : cwd}`);
```
justinmk, line 92: `cd` is somewhat clumsy, should we pass `cwd` to `spawn()`?  https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

xiyaowong: Haha, I realized this after the merger. Too lazy to make changes.

PR #1695, opened 2023-12-09
feat: Add option `afterInitConfig`

xiyaowong:
The main idea: Save the config to a temporary file, and then use nvim's startup option -c and command source to load the file.

This configuration is suitable for simple operations such as configuring mappings in a temporary environment, but it can also execute any logic. The reason for using "after" is to avoid conflicting with existing configurations, and at the same time, it can be used to determine whether local configurations have been loaded in this configuration, thus executing different logic.

In addition, this configuration will not be automatically ignored based on neovimClean. Users can use "finish" on the first line to end the logic.

Close #1350

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -1,4 +1,5 @@
 import { ChildProcess, execSync, spawn } from "child_process";
+import { existsSync, mkdirSync, writeFileSync } from "fs";
 import path from "path";
 
 import { attach, NeovimClient } from "neovim";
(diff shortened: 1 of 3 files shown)


theol0403 (approved):

src/main_controller.ts
```diff
@@ -119,6 +120,13 @@ export class MainController implements vscode.Disposable {
         if (!config.clean && config.neovimInitPath) {
             args.push("-u", config.neovimInitPath);
         }
+        if (config.afterInitConfig.length) {
+            const storagePath = this.extContext.globalStorageUri.fsPath;
+            if (!existsSync(storagePath)) mkdirSync(storagePath, { recursive: true });
+            const vim = path.join(storagePath, "afterInitConfig.vim");
+            writeFileSync(vim, config.afterInitConfig, { encoding: "utf8" });
```
justinmk, line 127: Writing a file should not be needed. Nvim has `nvim_exec2`, or you can `:source` a temporary buffer.

xiyaowong: Yes, I seem to have made it complicated (misled by the previous requirement of startup parameters in the #1350).

PR #1699, opened 2023-12-11
fix(buffer): prefer real path as buffer name

### LINKED ISSUE #798: Current file name in % register is invalid
The percent register `%` has an invalid file name which makes it unusable.

For example, I wanted to quickly run my current file through node, so I ran the following command:

```
:! node %
```

which got expanded as:

```
VSCode-Neovim:

:! node __vscode_neovim__-file:///Users/path/to/my/file.js
```

That prefix obviously made the command fail. Is that a hard requirement to have to make this extension work?

xiyaowong:
- [x] WSL
- [x] Test behavior in remote development

Resolve #798

### DIFF
--- a/runtime/lua/vscode-neovim/internal.lua
+++ b/runtime/lua/vscode-neovim/internal.lua
@@ -268,4 +268,16 @@ function M.start_visual(buf, anchor, active)
   api.nvim_feedkeys((v == "V" or v == "\x16") and "gvv" or "gv", "n", false)
 end
 
+---Translate from a Windows path to a WSL path
+---@param path string
+---@return string
+function M.wslpath(path)
+  local ok, ret = pcall(vim.fn.system, { "wslpath", path })
+  if not ok then
+    vim.notify(ret, vim.log.levels.ERROR)
+    return path
+  end
+  return vim.trim(ret)
+end
+
 return M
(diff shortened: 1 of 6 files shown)


runtime/lua/vscode-neovim/internal.lua
```diff
@@ -268,4 +268,15 @@ function M.start_visual(buf, anchor, active)
   api.nvim_feedkeys((v == "V" or v == "\x16") and "gvv" or "gv", "n", false)
 end
 
+---Translate from a Windows path to a WSL path
+---@param path string
+---@return string?
+function M.wslpath(path)
+  local f = io.popen(("wslpath '%s'"):format(path))
+  if f then
```
justinmk, line 276: wouldn't you want an error instead of silently ignoring a failure?

runtime/lua/vscode-neovim/internal.lua
```diff
@@ -268,4 +268,15 @@ function M.start_visual(buf, anchor, active)
   api.nvim_feedkeys((v == "V" or v == "\x16") and "gvv" or "gv", "n", false)
 end
 
+---Translate from a Windows path to a WSL path
+---@param path string
+---@return string?
+function M.wslpath(path)
```
theol0403, line 274: Why do we need this, can't we use the ts wsl function?

xiyaowong: Executing wsl.exe creates a new connection to the instance each time, which is unnecessary. Since nvim is already running in the instance, it is more efficient and reasonable to use it as a bridge to execute Linux commands.

theol0403 (approved):

PR #1705, opened 2023-12-13
refactor: source "afterInitConfig" using nvim_exec2

xiyaowong:


package.json
```diff
@@ -90,11 +90,8 @@
                     "markdownDescription": "Full path to custom neovim init.vim file on Linux or default WSL distro, equals to startup option `-u`. If the `useWSL` setting is checked, this path will be used to find `init.vim` in Windows Subsystem for Linux. You can also use `exists('g:vscode')` in your init.vim to check if neovim is being run in vscode and share init file between neovim/vscode"
                 },
                 "vscode-neovim.afterInitConfig": {
-                    "type": [
-                        "string",
-                        "array"
-                    ],
-                    "default": "",
+                    "type": "array",
```
justinmk, line 93: 👍 it's almost always better to require the more structured form, if possible.

justinmk (approved):

theol0403 (approved):

xiyaowong:
I think "neovimConfig" is not a good name, as users may be confused about its difference with "init.lua". Let's use "after" for now. And I intentionally chose "after" because I want this option to only serve as a supplementary option for simple setup, such as for temporary development environments like codespaces.

PR #1709, opened 2023-12-15
feat: override the default vim.ui interfaces

giancarlo-misasi:
Override the default vim.ui interfaces using the QuickPick and InputBox APIs. Uses two new actions, ui_select and ui_input to wire up the ui functions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -457,18 +457,6 @@
             {
                 "command": "vscode-neovim.ctrl-y",
                 "title": "Neovim: ctrl-y"
-            },
-            {
-                "command": "vscode-neovim.shift-m",
-                "title": "Neovim: shift-m"
-            },
-            {
-                "command": "vscode-neovim.shift-l",
-                "title": "Neovim: shift-l"
-            },
-            {
-                "command": "vscode-neovim:shift-h",
-                "title": "Neovim: shift-h"
             }
         ],
         "keybindings": [


xiyaowong:
1. Directly override the functions in the vim.ui module.
2. Use ActionManager to add new actions, see action.ts

giancarlo-misasi:
Good suggestions - I've made the recommended changes to support vim.ui.select for now. We could also support vim.ui.input in a follow-up pr this way as well.

xiyaowong:
I think the "input" should also be implemented in this pull request, there is no need to implement it in a new PR.

runtime/modules/ui.lua
```diff
@@ -0,0 +1,81 @@
...
+--                      • prompt (string|nil) Text of the prompt. Defaults to
+--                        `Select one of:`
+--                      • format_item (function item -> text) Function to format
+--                        an individual item from `items`. Defaults to
+--                        `tostring`.
+--                      • kind (string|nil) Arbitrary hint string indicating the
+--                        item shape. Plugins reimplementing `vim.ui.select` may
+--                        wish to use this to infer the structure or semantics of
+--                        `items`, or the context in which select() was called.
+--                      • match_on_detail (boolean|nil) optional flag to use the detail when filtering the items. items must have a detail field.
```
xiyaowong, line 33: It is unnecessary to fully match the options of VSCode's QuickPick, as this is not very useful and users will not even know about this option unless it is explicitly mentioned in the documentation. I suggest keeping the signature of the original vim.ui function consistent.

runtime/modules/ui.lua
```diff
@@ -0,0 +1,81 @@
...
+
+local function vscode_ui_select(items, opts, on_choice)
+  -- fix-up defaults
+  items = items or { 'a', 'b', 'c' }
+  opts = opts or {}
+  opts.prompt = vim.F.if_nil(opts.prompt, "Select one of")
+  opts.format_item = vim.F.if_nil(opts.format_item, tostring)
+  opts.kind = vim.F.if_nil(opts.prompt, "string")
+  opts.match_on_detail = vim.F.if_nil(opts.match_on_detail, false)
+  on_choice = vim.schedule_wrap(vim.F.if_nil(on_choice, function(_, _) end))
```
xiyaowong, line 46: https://github.com/neovim/neovim/blob/8f08b1efbd096850c04c2e8e2890d993bd4d9f95/runtime/lua/vim/ui.lua#L40-L44

runtime/modules/ui.lua
```diff
@@ -0,0 +1,81 @@
...
+  vscode.action("ui_select", {
+    args = {
+      items = vscode_items,
+      opts = vscode_opts,
+    },
+    callback = function(err, res)
+      if not err then
+        on_choice(res, 1)
+      else
+        on_choice(nil, -1)
```
xiyaowong, line 74: I believe the logic is incorrect.

https://github.com/neovim/neovim/blob/8f08b1efbd096850c04c2e8e2890d993bd4d9f95/runtime/lua/vim/ui.lua#L37-L38

src/test/suite/vscode-integration-specific.test.ts
```diff
@@ -531,4 +531,25 @@ describe("VSCode integration specific stuff", () => {
...
         await assertContent({ cursor: [2, 2] }, client);
     });
+
+    it("vim.ui.select works", async () => {
+        await openTextDocument({ content: "" });
+
+        const p1 = client.lua(`
+            vim.ui.select({ 'red', 'green', 'blue' }, {}, function(item, idx)
+                local buffer = vim.api.nvim_get_current_buf()
+                vim.api.nvim_buf_set_lines(buffer, 0, -1, false, {item.label})
```
xiyaowong, line 541: For this type of test, you can store the result in a temporary variable, and then assert it.

nvim: vim.g._ret = ...
vscode: client.getVar ...

xiyaowong (changes requested):

runtime/modules/ui.lua
giancarlo-misasi: I chose to include the match_on_detail as it is a useful vscode feature available. I would prefer to document it and make it available rather than limiting it to what the vim.ui.select spec currently supports. In my opinion there is already precedent, as the vim.ui.select function takes a kind and format function so that you can pass your own custom items. Passing vscode specific options to improve the vscode experience seems reasonable to me.

xiyaowong: It is a burden for both users and maintainers.

giancarlo-misasi: Take a look at the update and let me know your thoughts - as a middle ground, I removed the extra options, but I allow passing through 'detail' when users provide tables instead of just strings. This keeps the API signature of vim.ui.select unchanged, but, still enables users to take advantage of the two-line label/detail display with the filtering available in the vscode QuickPick.

xiyaowong:
I'm so sorry! I recently tried using the Github pull request plugin and it seems like I accidentally caused all of your submissions to be lost. Would you mind resubmitting the PR? Sorry about that.

giancarlo-misasi:
No problem, I've re-opened it here: https://github.com/vscode-neovim/vscode-neovim/pull/1710 - I had to pull the changes out of the reflog and my remote fork was also clobbered as well.

PR #1710, opened 2023-12-17
feat: override the default vim.ui interfaces

giancarlo-misasi:
Override the default vim.ui interfaces using the QuickPick and InputBox APIs. Uses two new actions, ui_select and ui_input to wire up the ui functions.
* Previous pull request: https://github.com/vscode-neovim/vscode-neovim/pull/1709

### DIFF
--- a/runtime/modules/ui.lua
+++ b/runtime/modules/ui.lua
@@ -1,0 +1,141 @@
+local vscode = require("vscode-neovim")
+
+--- Prompts the user to pick from a list of items, allowing arbitrary (potentially asynchronous)
+--- work until `on_choice`.
+---
+--- Example:
+---
+--- ```lua
+--- vim.ui.select({ 'tabs', 'spaces' }, {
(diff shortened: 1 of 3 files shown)


theol0403:
Fwiw, `vim.ui.input` already worked, using the cmdline -> quickpick as a backend. Does this provide a better approach? What about when `vim.ui.input` is called from within the cmdline, and vscode is told to render 2 quickpicks?

theol0403:
I also still think with these changes we can make `vim.notify = vscode.notify` the default.

xiyaowong:
> I also still think with these changes we can make `vim.notify = vscode.notify` the default.

VS Code notifications can be annoying sometimes. I rarely use "vscode.notify".

The input is indeed not much different from the native one, but I personally prefer to customize it to avoid mode switching and triggering events.

xiyaowong (approved):

theol0403 (approved):

Issue #1700, opened 2023-12-12
bug: nvim-treesitter-textobjects is not working


rohitjagan:
### Did you check docs and existing issues?

- [X] I have read all the vscode-neovim docs
- [X] I have searched the existing issues of vscode-neovim
- [X] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.9.4

### Operating system/version

MacOS 14.0

### Describe the bug

After an object in a buffer has been modified, attempting to select the object results in the following error:

```
E5108: Error executing lua ...r-textobjects/lua/nvim-treesitter/textobjects/select.lua:40: attempt to get length of a nil value
stack traceback:
	...r-textobjects/lua/nvim-treesitter/textobjects/select.lua:40: in function 'next_position'
	...r-textobjects/lua/nvim-treesitter/textobjects/select.lua:77: in function 'include_surrounding_whitespace'
	...r-textobjects/lua/nvim-treesitter/textobjects/select.lua:112: in function 'select_textobject'
	[string ":lua"]:1: in main chunk
```

https://github.com/vscode-neovim/vscode-neovim/assets/17104878/3fa6f825-e651-4c9f-a207-8e902c1f40ff

### Steps To Reproduce

1. Install `nvim-treesitter` and `nvim-treesitter-textobjects` via `lazy.nvim` using the default configs
2. Create a file with a function
3. Use `vif` to select inside the function (this works)
4. Delete the function and undo (`u`) to restore it
5. Try `vif`

### Expected Behavior

You should be able to select the objects even after the buffer has been changed.

xiyaowong:
```lua
vim.api.nvim_create_autocmd({ 'TextChanged', 'InsertLeave' }, {
  group = vim.api.nvim_create_augroup('AutoTresitterParse', {}),
  callback = (function()
    local timer
    return function()
      if timer and timer:is_active() then timer:close() end
      timer = vim.defer_fn(function()
        if vim.b._last_parse_changedtick ~= vim.b.changedtick then
          vim.b._last_parse_changedtick = vim.b.changedtick
          pcall(function() vim.treesitter.get_parser():parse() end)
        end
      end, 500)
    end
  end)(),
})
```

PR #1713, opened 2023-12-18
fix(treesitter): manually trigger parsing

xiyaowong:
issue #1700

This is a common use case, and "treesitter-objects" is useful. (Note: I haven't explored why treesitter doesn't automatically parse after text changes.)

### DIFF
--- a/runtime/modules/treesitter.lua
+++ b/runtime/modules/treesitter.lua
@@ -1,0 +1,20 @@
+-- TODO: Figure out why Treesitter doesn't parse.
+vim.api.nvim_create_autocmd({ "TextChanged", "InsertLeave" }, {
+  group = vim.api.nvim_create_augroup("VSCodeAutoTresitterParse", {}),
+  callback = (function()
+    local timer
+    return function()
+      if timer and timer:is_active() then
+        timer:close()
+      end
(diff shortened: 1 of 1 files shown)


theol0403:
This bug has been bothering me for a long time but it seemed too strange to make an issue for.

Ideally we figure out why it's not working... but it makes no sense to me. Operations like undo all happen on the nvim side.

xiyaowong:
> but it makes no sense to me.

I added a todo. It could be a simple bug for someone familiar with treesitter. Let’s wait for someone interested to contribute.

theol0403 (approved):

theol0403:
Actually I just tested this, and I can still reproduce if I undo and execute rapidly. Is there a way of removing the timer and making it triggered automatically? If not, we should maybe speed up the timer.

xiyaowong:
> rapidly

In most cases, it wouldn't be so fast... You can adjust the time down without any impact. I copied it directly from my config. I try to reduce unnecessary operations as much as possible based on my own needs

xiyaowong:
Actually one line is enough:
```vim
au TextChanged * silent! lua vim.treesitter.get_parser():parse()
```

PR #1717, opened 2023-12-18
chore: reduce treesitter parsing debounce time

xiyaowong:


theol0403:
Shouldn't we use the one-liner?

xiyaowong:
Unnecessary operations, treesitter sometimes doesn't parse quickly.

theol0403 (approved):

PR #1719, opened 2023-12-23
Add neovim.fullMode context key for mode(1)

### LINKED ISSUE #1718: feature: keybinding context for full mode (i.e. `mode(1)`)
### Did you check the docs?

There might be other ways to solve this or workarounds, but I have a use-case where I'd like to enable a keybinding only if `vscode-neovim` does not have an operator-pending. For example, something like this:
```jsonc
    {
        "key": "shift+5",
        "command": "rust-analyzer.matchingBrace",
        "when": "editorTextFocus && neovim.mode != 'insert' && neovim.mode != 'operator-pending' && editorLangId == 'rust'"
    }
```

Unfortunately, `neovim.mode` doesn't have the concept of operator-pending mode, and I think changing its behavior might break a lot of existing keybinds (since `operator-pending` aka `no[vV^V]` is really a subset of `n`).

I think there are multiple options here, but maybe the simplest one would be a new context key that feeds the result of `mode(1)` into a new context key, instead of trying to cover every case like [the current mapping](https://github.com/vscode-neovim/vscode-neovim/blob/master/src/mode_manager.ts#L18).

My proposal would be a new context key like `neovim.fullMode`, which just returns the same string that `mode(1)` does (e.g. `no`, `noV`, `Vs`, etc.). Since VSCode supports [matching](https://code.visualstudio.com/api/references/when-clause-contexts#match-operator) on these strings, this should be flexible enough to allow lots of different combinations. My example above might look something like this (probably would need to try it out):

```jsonc
    {
        "key": "shift+5",
        "command": "rust-analyzer.matchingBrace",
        "when": "editorTextFocus && !(neovim.fullMode =~ /^(i|no)/) && editorLangId == 'rust'"
    }
```

It might be feasible to write a custom function that calls back to VSCode, like `require('vscode-neovim').call("rust-analyzer.matchingBrace")`, but would require a remapping and might be tricky to cover all the edge cases. Also, it seems a bit silly to call into neovim just to call back to VSCode, instead of binding keys to a VSCode command directly.

Another option would be to attempt to add a simpler flag like `neovim.operatorPending`, but I could foresee people wanting to add other flags for various "sub-modes", like `niV`, `vs`, etc. (see `:help mode()` for a full list). Passing the full string through seems more flexible and allows for advanced use-cases, while `neovim.mode` can probably be used for most basic keybindings.

### Additional context

If this feature seems like a reasonable request, I'm happy to try implementing it myself. It seems like it would be fairly straightforward to add to the existing mode manager without breaking anything. Thanks!

ian-h-chamberlain:
Closes #1718 

Similar to `neovim.mode` but simpler, this is just a raw passthrough of `mode(1)` to the VScode keybinding contexts.

I'm happy to adjust the implementation if an alternative seems preferable, but this was easy enough to write so just figured I'd put up a PR based on my thoughts there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,6 +521,19 @@ To pass custom bindings to Neovim, for example <kbd>C-h</kbd> in normal mode, ad
 }
 ```
 
+If you need more precise information about the mode (i.e. `mode(1)` in neovim),
+you can use the `neovim.fullMode` context key for matching:
+
+```jsonc
+{
+    "command": "vscode-neovim.send",
+    "key": "ctrl+h",
+    // don't activate during insert or operator-pending mode
+    "when": "editorTextFocus && !(neovim.fullMode =~ /^(i|no)/)",
+    "args": "<C-h>"
+}
+```
+
 ### Disable keybindings
 
 There are three configurations for toggling keybindings:
--- a/src/eventBus.ts
+++ b/src/eventBus.ts
@@ -115,7 +115,7 @@ type EventsMapping = {
     ["open-file"]: [string, 1 | 0 | "all"];
     ["external-buffer"]: [BufferInfo, 1 | 0, number];
     ["window-changed"]: [number];
-    ["mode-changed"]: [string];
+    ["mode-changed"]: [string, string];
     ["notify-recording"]: undefined;
     reveal: ["center" | "top" | "bottom", boolean];
     ["move-cursor"]: ["top" | "middle" | "bottom"];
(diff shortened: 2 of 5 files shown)


xiyaowong:
I think this is not a very common case. You can use API to set the context.

```vim
au VimEnter,ModeChanged * call VSCodeCall('setContext', 'neovim.fullMode', mode(1))
```

PR #1742, opened 2024-01-12
fix: C-w in cmdline

### LINKED ISSUE #1741: bug: <C-w><C-w> mapping blocks <C-w> in cmdline

### Neovim version (nvim -v)

0.9.5

### Operating system/version

Windows 10

### Describe the bug

`ctrl+w ctrl+w` keybinding's incorrect when clause does not support `ctrl+w` usage in `cmdline` mode.

Typing `ctrl+w` in `cmdline` triggers `ctrl+w ctrl+w` keybinding with higher priority.

### Steps To Reproduce

1. `normal` mode: type `:` to enter `cmdline` mode
2. `cmdline` mode: type some random text `abc def ghi`
3. `cmdline` mode: press `ctrl+w` shortcut
4. Notice that the status line shows message `(Ctrl+W) was pressed. Waiting for second key of chord...`
5. if we press `ctrl+w` again, the cursor will focus on the active editor group as the result of triggering `ctrl+w ctrl+w` keybinding: `workbench.action.focusNextGroup`

### Expected Behavior

delete the word before the cursor in the `cmdline` box

xudyang1:
Fix #1741

### DIFF
--- a/package.json
+++ b/package.json
@@ -1627,7 +1627,7 @@
             {
                 "key": "ctrl+w q",
                 "command": "workbench.action.closeActiveEditor",
-                "when": "!editorTextFocus && !terminalFocus && !filesExplorerFocus && !searchViewletFocus"
+                "when": "!editorTextFocus && neovim.mode != 'cmdline' && !terminalFocus && !filesExplorerFocus && !searchViewletFocus"
             },
             {
                 "key": "ctrl+Escape",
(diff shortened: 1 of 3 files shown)


xiyaowong:
1. Do other keybindings not have this issue?
2. Modify the files in scripts/ to define keybindings.

xudyang1:
It seems that the test code in `command-line.test.ts` behaves differently than manually type `ctrl` keybindings.

https://github.com/vscode-neovim/vscode-neovim/blob/7e5222bb0c02e3f371be6f4c97179d1c71a6bb49/src/test/suite/command-line.test.ts#L66C1-L71C85

The when clause in vscode keybinding does not block `<C-w>` if it is sent by the script. As a result, the test would never fail even if those keybindings are assigned with incorrect when clause.

xiyaowong:
In theory, other shortcut keys will also have this problem, but now only C-w is used

theol0403 (approved):

Issue #1056, opened 2022-10-15
c6d227b breaks visual-multi plugin


Sduby22:
## expected behavior(9fa7f31)

after hitting `Ctrl-N` 3 times I created 3 cursors, and after hit `a` and `bbb` I entered insert mode and appended `bbb` 

https://user-images.githubusercontent.com/29040696/195978316-ad5f9e12-4479-4c28-848d-1ed4c0c49c83.mp4

## result behavior(c6d227b)
after hitting `bbb`, the letters immediately disappeared 

https://user-images.githubusercontent.com/29040696/195978220-a0a10e23-03c1-42cd-b2d1-8d95029fd856.mp4

## bad commit
c6d227b: feat(document_change_manager): replay insert-mode edits in realtime using `nvim_buf_set_text` (#992) …

## vscode version

```
Version: 1.72.1
Commit: 129500ee4c8ab7263461ffe327268ba56b9f210d
Date: 2022-10-10T17:23:16.506Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Darwin arm64 21.4.0
Sandboxed: No
```

theol0403:
Thanks for the report!

I think this has something to do with the fact that edits are now sent using nvim_buf_set_text, and maybe vim-visual-multi is incompatible with this.

It's possible there is a less significant bug going on here as well. 

In the case of the first potential cause, I think the solution would be to redirect incoming edits being sent by `nvim_buf_set_text` and convert them to `nvim_input` in certain cases. This is already mentioned in https://github.com/vscode-neovim/vscode-neovim/pull/992#issuecomment-1228734888 as an option to remove dot-repeat hack. Essentially, an algorithm must be made that converts a sequence of `nvim_buf_set_text` into a sequence of `nvim_input(<Left>/<Right> <BS>/<Del> text)`. 

This is something I was hoping to try, but unfortunately I won't have time to work on this in the short term. I have many midterms approaching and many things to do. I'll try to fix this when I get the chance though.

For now, stick with the older plugin version.

Sduby22:
Thank you!

To whom it may concern, sticking to `v0.90.0` should work now.

ynotzort:
Can confirm that version `v0.0.90` works but all later versions show this faulty behavior.

[...]

Issue #409, opened 2020-10-16
Better multi-cursor support (similar to VSCodeVim) to allow per-cursor motions


macintacos:
It'd be wonderful is this extension could treat each cursor that is generated as its own "context", and allow normal-mode to work with multiple cursors. I have no idea how complicated this would be to execute given the way that it is implemented, but let me explain.

Say I want to use VSCode's built-in functionality to add a cursor below the current line while in `normal` mode. This feels natural to me, and I use things like `cmd+d` to add cursors to the next occurrence of a given word all the time (for example). When I have added all the cursors I want, my first thought is "okay time to do a motion", so I execute something like `ciw`. In VSCodeVim, this behaves as I expect, but in vscode-neovim, it does not.

Here are some gifs that explain things visually a lot better than I can via text. Here's what happens in neovim:

![Screen Recording 2020-10-16 at 19 25 07](https://user-images.githubusercontent.com/8128941/96322570-a3820000-0fe7-11eb-9af2-8d95b9fe1a3f.gif)

And here's what happens in VSCodeVim:

![Screen Recording 2020-10-16 at 19 29 58](https://user-images.githubusercontent.com/8128941/96322521-66b60900-0fe7-11eb-9a4f-311c8b64eeb4.gif)

In order to get to what I think would be "good multi-cursor support", I think that the following would probably need to happen:

- Figure out a way to make it so that if you have multiple cursors, the first instance of `ESC` is interpreted as "go into normal mode".
- When in normal mode + multiple cursors, treat each cursor independently, and send motions/commands to those cursors specifically. If I hit `ciw` (like I did in the gifs about) each cursor should interpret its "context" to execute the motion on the text it has

Lastly, I'd just like to note that this issue is pretty much the only thing keeping me on VSCodeVim at this point. Aside from that, I really like this extension's implementation, and it has a ton of upsides in comparison to VSCodeVim's "emulation-focused" approach (which requires re-inventing the wheel, which feels like an uphill battle). If this extension got first-class multi-cursor support, there'd be nothing holding me back from singing this extension's praises and using it every single day.

Please let me know if you have any questions about the above details!

muhajirdev:
Looking for similar things.
Found this https://medium.com/@schtoeffel/you-don-t-need-more-than-one-cursor-in-vim-2c44117d51db

Might be useful for you

Shatur95:
[...]

PR #1755, opened 2024-01-19
make visual-multi works again! #1056 #409 #97 #992

rozbo:
# Background
As #1056 said, #992 breaks my favorite plugin [https://github.com/mg979/vim-visual-multi](https://github.com/mg979/vim-visual-multi),  I simply can't work without it. from neovim to ideavim, it’s always with me.
The vscode-neovim is a perfect tool except that it lacks this plugin. It's such a shame!
So, I decided to fix it.
I start from https://github.com/vscode-neovim/vscode-neovim/issues/1056#issuecomment-1280190144 , and study a lot by https://github.com/vscode-neovim/vscode-neovim/pull/992 , then study many Neovim api..., and then, I follow @theol0403 to convert the `nvim_buf_set_text` param `{buffer}, {start_row}, {start_col}, {end_row}, {end_col}, {replacement}` to nvim_input.
The 1st step is check it's on multi edit or not. this is easy by read the `visual-multi` source code, we can check it by 
```lua
local success, selection = pcall(vim.api.nvim_buf_get_var, bufnr, 'VM_Selection')
if success and selection ~= '' then
  print("In selection mode")
else
  print("Not in selection mode")
end
```
we can add this code to https://github.com/vscode-neovim/vscode-neovim/blob/d93300bc193c2fee18465f29acadff08c19bac6a/runtime/lua/vscode-neovim/internal.lua#L48-L52, since we can o avoid RPC round trips.

 At first, it ran very nice until I hit the backspace key, it broke me! I spent two hours hard work code to compare the old and new lines, and then figure out where the backspace key was needed. now I met Left and Right! Obviously, this cannot be calculated from these parameters. 😢 

It is also possible to roll back to the previous code, because according to the issue, it can be used before 0.9. It can be clearly seen from the pull request record that the `nvim_buf_set_lines` api was used before. Fortunately, there is not much difference between the two APIs. The original lines can be obtained through `nvim_buf_get_lines`, merged to produce the final lines, then updated with `nvim_buf_set_lines`. However, this does not work at all and does not restore visual-multi to its original state.

Finally, it was 2 o'clock in the morning and I still couldn't get it done. I thought it was time to give up. It seemed like an impossible task.
 At this moment, I suddenly thought that what I just did was to restore the user input to the keys queue and then pass it to `nvim_input`. If we capture the user's keys input and then send it directly to neovim without vscode processing, then Isn't that what we want?

https://github.com/vscode-neovim/vscode-neovim/blob/d93300bc193c2fee18465f29acadff08c19bac6a/src/typing_manager.ts#L128-L134

it check the isInsertMode  and not in Recording mode then bind the keys to neovim, we just update Recording to `isRecordingInInsertMode` when we met visual-multi, and when we final the edit, it will back to normal mode, now `mode_manager` will set `isRecordingInInsertMode` back!
https://github.com/vscode-neovim/vscode-neovim/blob/d93300bc193c2fee18465f29acadff08c19bac6a/src/mode_manager.ts#L105-L110

So, the final code is very simple.

By the way, it also fixes the previous problem of displaying content synchronously after editing is completed. Now, it works very well, just like in real neovim.

### DIFF
--- a/vim/vscode-neovim.vim
+++ b/vim/vscode-neovim.vim
@@ -80,6 +80,9 @@ function! s:onInsertEnter()
     if !empty(reg)
         call VSCodeExtensionNotify('notify-recording', reg)
     endif
+    if exists("b:VM_Selection") && !empty(b:VM_Selection)
+        call VSCodeExtensionNotify('notify-recording', reg)
+    endif
 endfunction
 
 


rozbo:
动画

xiyaowong:
More like a hack

theol0403:
Yeah not a fan, but might be the best option at the moment. At least should have a comment.

theol0403:
@rozbo can you add a comment in the code explaining what it is for?

theol0403 (approved):

rozbo:
haha, looks like I am later... 😆 
great job!

dgokcin:
@rozbo hello and thanks for your effort in trying to replicate the vim-visual-multi behaviour in neovim. It is also a plugin that I can't live without.

I am a little bit confused about what exactly I need to do to have the same functionality I have in my neovim configuration. I have read in the issues that some vscode keybindings needs to be set but I am not exactly sure what to do. Right now `<C-n> is just skipping to the next line in vscode.

Thanks!

mbhutton:
@dgokcin I took the liberty of looking at your LazyVim setup. Try removing the `cond = not vim.g.vscode` line in your `vim-visual-multi.lua` plugin spec file, as that's likely overriding the `vscode = "true"` line above it.

That should at least get the plugin to be active in VSCode.

mbhutton:
As a side note: the current behaviour makes the vim-visual-multi plugin interact badly with `InsertLeave` autocommands like the following:
```
-- When leaving insert mode, restore cursor to last edit position to avoid jumping backwards
vim.api.nvim_create_autocmd({ "InsertLeave" }, {
  command = "normal `^",
})
```

When the above autocmd is enabled, when leaving insert mode following a vim-visual-multi edit, an extra column of characters is inserted between the new and old text. E.g. inserting `new ` before 3 rows of `old`, then pressing `<Esc>`, yields the following:
```
new oold
new oold
new oold
```

I expect this is far too niche a case to support (hence being a comment and not an issue), so I plan to use vscode-neovim's existing great multi-cursor support instead. Justing commenting here in case someone else runs into the same issue.

dgokcin:
Thank you and especially for taking the time and looking at my config to see what is wrong. I do not have insertleave auto commands so this will work perfectly for me!

PR #1786, opened 2024-02-08
feat: expose logs in OutputChannel, log info on startup

justinmk:
- Introduce `Logger.info()` and "info" log-level.
- Cleanup settings descriptions.
- Create outputChannel as a `{ log: true }` channel for enhanced rendering and features.
- Allow the logger to log to the vscode Output channel. 
    - But disable it for noisy (and looping) sources like `CursorManager` and `DocumentChangeManager`.

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,6 +35,14 @@ npx vsce package -o vscode-neovim.vsix
 1. Open the repo in VSCode.
 2. Go to debug view and click `Run Extension` (F5).
 
+### Logging
+
+You can observe the extension logs via the `vscode-neovim` Output channel or from the dev tools
+console (run the `Developer: Toggle Developer Tools` vscode command to see the console).
+
+Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by
+the [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
+
 ### Run Tests
 
 1. Open the repo in VSCode.
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -3,20 +3,35 @@ import fs from "fs";
 import { inspect } from "util";
 
 import { Disposable, window } from "vscode";
+import * as vscode from "vscode";
 
 import { disposeAll } from "./utils";
 
 export enum LogLevel {
+    /** Disables all logging. */
     none = 0,
     error = 1,
     warn = 2,
-    debug = 3,
+    info = 3,
+    debug = 4,
 }
 
 export interface ILogger {
     debug(...args: any[]): void;
+    info(...args: any[]): void;
     warn(...args: any[]): void;
     error(...args: any[]): void;
+    /**
+     * Logs a message only if `uri` is not a log file/channel (to avoid infinite loop).
+     *
+     * If the user is viewing the extension OutputChannel or logfile, we must
+     * not log events triggered by that channel or file itself.
+     *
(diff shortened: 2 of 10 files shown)


src/logger.ts
```diff
@@ -3,20 +3,35 @@ import fs from "fs";
...
+    info = 3,
+    debug = 4,
 }
 
 export interface ILogger {
     debug(...args: any[]): void;
+    info(...args: any[]): void;
     warn(...args: any[]): void;
     error(...args: any[]): void;
+    /**
```
theol0403, line 24: How does this work with the old `debug`/`warn` methods? Should we stick to those (and submit the uri to them), or remove them in favour of `log`? Which one should you use?

theol0403 (commented):
Very awesome!

> Disable it for noisy (and looping) sources like CursorManager and DocumentChangeManager.

Won't this cause confusion when debugging `CursorManager` and seeing no logs come through the output?

justinmk:
> Won't this cause confusion when debugging `CursorManager` and seeing no logs come through the output?

It's only "disabled" for the output channel (to avoid infinite loop). If one needs to debug cursor movement they can use the dev tools console or log-file.

theol0403:
> It's only disabled for the output channel (to avoid infinite loop). If one needs to debug cursor movement they can use the dev tools console or log-file.

Just make sure this is documented because I see people trying to use the output channel for all debugging without knowing the dev-tools exists.

src/logger.ts
justinmk: Adding `uri` to the existing debug/warn/error methods would be very high-churn for a somewhat limited use-case. (Possibly could be done with typescript "overload" signature, but also not sure that's worth the effort).

justinmk: Is the docstring here not sufficient?

CONTRIBUTING.md
```diff
@@ -35,6 +35,14 @@ npx vsce package -o vscode-neovim.vsix
 1. Open the repo in VSCode.
 2. Go to debug view and click `Run Extension` (F5).
 
+### Logging
+
+You can observe the extension logs via the `vscode-neovim` Output channel or from the dev tools
+console (run the `Developer: Toggle Developer Tools` vscode command to see the console).
+
+Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by
```
justinmk, line 43: added doc

src/main_controller.ts
```diff
@@ -57,14 +57,14 @@ export class MainController implements vscode.Disposable {
...
+    public messagesManager!: MessagesManager;
     public viewportManager!: ViewportManager;
 
     public constructor(private extContext: ExtensionContext) {}
 
-    public async init(): Promise<void> {
+    public async init(outputChannel: vscode.LogOutputChannel): Promise<void> {
         const [cmd, args] = this.buildSpawnArgs();
-        logger.debug(`Spawning nvim: ${cmd} ${args.join(" ")}`);
+        logger.info(`starting nvim: ${cmd} ${args.join(" ")}`);
```
justinmk, line 67: Log Nvim start-command at startup. Will add more info logs in the future (e.g. the result of `stdpath('config')`)

src/logger.ts
theol0403: If I were a new contributor adding a new feature, I would be unsure whether to use `logger.warn` or `logger.log(uri, LogLevel.warn)`

justinmk: I understand. But in this case I can't think of a general guidance. New contributors sometimes need to figure things out. If they choose the "wrong" one, the cost is negligible, in this cas.

theol0403 (approved):

PR #1849, opened 2024-03-13
fix(startup): $VIMRUNTIME check fails on WSL

justinmk:
Problem:
Since c9ddc3ea90444, on startup, on WSL and flatpak, vscode-neovim shows a $VIMRUNTIME read error.

    Cannot read $VIMRUNTIME directory "/opt/nvim-linux64/share/nvim/runtime".
    Ensure that VSCode has access to that directory. Also try :checkhealth.

Solution:
Use vscode's VFS (virtual filesystem) instead of the node API. TODO: we should use vscode's VFS _everywhere_, if possible.

Fix #1848

PR #1850, opened 2024-03-13
fix: check the read permission of $VIMRUNTIME in nvim

### LINKED ISSUE #1848: Checking $VIMRUNTIME read permissions will always fail on wsl
`Cannot read $VIMRUNTIME directory "/opt/nvim-linux64/share/nvim/runtime". Ensure that VSCode has access to that directory. Also try :checkhealth.`
I'm getting that error but I do not fully understand what am I suppose to do, running on WSL, nvim installed from pre-built release

_Originally posted by @Yusuf007R in https://github.com/vscode-neovim/vscode-neovim/issues/1816#issuecomment-1992304302_

xiyaowong:
nvim and vscode may not be in the same file system.

Fix #1848


justinmk:
Can we see if https://github.com/vscode-neovim/vscode-neovim/pull/1849 works? We should generally be using the vscode VFS everywhere, anyway.

xiyaowong:
This problem is mainly caused by the nvim executable file in wsl used by the user, not the remote development function of vscode

src/main_controller.ts
```diff
@@ -369,20 +368,21 @@ export class MainController implements vscode.Disposable {
     private async validateNvimRuntime() {
         // WEIRD BUT TRUE: $VIMRUNTIME may be inaccessible even though Nvim itself is runnable! #1815
         const luaCode = `
-            return vim.env.VIMRUNTIME
+          local ok = pcall(function()
+            local dir = vim.loop.fs_opendir(vim.env.VIMRUNTIME)
```
justinmk, line 372: why not use `vim.fs` ?

justinmk:
> This problem is mainly caused by the nvim executable file in wsl used by the user, not the remote development function of vscode

The remote path is a WSL path. Though it's not in the "workspace", so that could cause problems.

justinmk (approved):

src/main_controller.ts
xiyaowong: I tried this. This function will not report an error even if there is no permission. I don't know how to judge whether the operation is successful.

xiyaowong: Not familiar with these APIs

justinmk: Just a suggestion. But if `vim.fs` returns zero files, then we can treat that as an error because $VIMRUNTIME must never be empty.

Note that `vim.fs.dir` returns an iterator.

xiyaowong:
It can be understood that wsl-nvim is called, not nvim itself.

justinmk (approved):
Nicely done

PR #1853, opened 2024-03-17
Pre-Commit Hooks

mbway:
Run linting and formatting before each commit. This is better for new contributors as they can get immediate feedback instead of waiting for the CI to reject their commits later.


package.json
```diff
@@ -2072,25 +2072,33 @@
...
+        "@commitlint/config-conventional": "^19.1.0",
+        "@johnnymorganz/stylua-bin": "^0.20.0",
         "@types/lodash-es": "^4.17.12",
         "@types/mocha": "^10.0.6",
         "@types/node": "^20.11.25",
         "@types/vscode": "^1.83.1",
         "@typescript-eslint/eslint-plugin": "^5.62.0",
         "@typescript-eslint/parser": "^5.62.0",
         "@vscode/test-electron": "^2.3.9",
+        "@vscode/vsce": "^2.24.0",
```
mbway, line 2101: I got a warning that `vsce` is deprecated so I updated to "@vscode/vsce"

package.json
```diff
@@ -2072,25 +2072,33 @@
         ]
     },
     "scripts": {
-        "lint": "eslint --ext .ts src",
-        "format": "prettier --ignore-path .gitignore --ignore-path .prettierignore --write .",
+        "lint:ts": "eslint --ext .ts,.cjs scripts src test_fixtures",
+        "lint:lua": "",
```
mbway, line 2076: lua linters like `luacheck` are available but are not provided through npm unfortunately. Requiring contributors to install an external tool would be annoying but not out of the question.

package.json
```diff
@@ -2072,25 +2072,33 @@
...
+        "format": "npm run format:ts && npm run format:lua",
         "keybind": "node ./scripts/keybindings/index.cjs && prettier --write ./package.json",
         "test": "node ./out/test/runTest.js",
         "pretest": "npm run test-compile",
         "vscode:prepublish": "webpack --mode production",
         "webpack": "webpack --mode development",
         "webpack-dev": "webpack --mode development --watch",
         "test-compile": "tsc -p ./",
-        "prepare": "husky install"
+        "prepare": "husky"
```
mbway, line 2088: `husky install` is deprecated and replaced with just `husky`

.husky/pre-commit
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
```
mbway, line 2: I don't think this header is required any more because `husky init` does not add it automatically

mbway:
I'm not a fan of how `lint-staged` automatically `git add`s the automatic changes but using `lint-staged` is a bit better than having to stash any changes not being committed manually.

theol0403:
Sweet! Can probably remove https://github.com/vscode-neovim/vscode-neovim/blob/master/.github/workflows/conventional_commits.yml if husky does it.

mbway:
still probably good to have the CI double check, as contributors can still bypass the hooks if they choose to (eg with `git commit --no-verify` or never running `npm run prepare`

CONTRIBUTING.md
```diff
@@ -265,10 +279,3 @@ For more information about possible improvements, see
 ### BufferManager
 
 BufferManager is responsible for the syncing of buffers and windows between vscode and nvim.
-
-## Maintenance
```
justinmk, line 269: do not remove this section. we need a place to keep notes about doing releases, etc. "style and tools" is not the right place for that.

justinmk (changes requested):
this is not worth 5000 lines of package-lock.json stuff. is there a way to do this without adding 5k lines to package-lock.json? every dependency are a maintenance burden.

it's really not that hard to run lint without husky.

justinmk:
> Can probably remove https://github.com/vscode-neovim/vscode-neovim/blob/master/.github/workflows/conventional_commits.yml if husky does it.

no, because there is zero guarantee that husky is setup correctly. CI is much more important than this local convenience.

theol0403:
I believe husky was added when we added the `npm run keybinds`. I would back any change that promotes simplicity.

mbway:
I had a go at re-creating this PR along the lines that have been requested:

- do not use husky
    - do not use any tooling like `pre-commit`?
- do not introduce dependencies on linting tools

But I'm not sure about how far along that route you want to go. The project already breaks these guidelines by depending on 7 `eslint` and `prettier` related packages contributing ~3400 lines to the `package-lock.json` file. I'm not sure where you want the line drawn. For `commitlint` and `stylua` should I install these with pinned versions in a script such as `.hooks/install.sh`:

```shell
npm install \
    @commitlint/cli@19.2.0 \
    @commitlint/config-conventional@19.1.0 \
    @johnnymorganz/stylua-bin@0.20.0
```
Should I do the same for `eslint` and `prettier`, removing them from `package.json`?

I think this approach has a greater possibility for people to forget to update if the pinned versions in the `.hooks/install` script ever change. It also makes it harder to provide `npm run lint` as it currently exists because the user may not have the dependencies installed, or is that not a concern?

theol0403:
I agree that dev dependencies on lint tools don't really add a maintenance burden. I'd rather things just be streamlined and concise. I don't think having an external install script promotes simplicity. I would be happy to merge this PR as-is, but if you think you can reduce the reliance on tooling/complexity, I'd be happy for it, even if it removes husky. I don't want any solution that adds extra steps or requires an explanation to use.

mbway:
Husky itself isn't needed but there will always need to be an installation step as the user has to opt in to run the hooks. Currently that is `npm run prepare`. If reliance on husky was removed that step would still be there, but would call out to a custom script that creates symlinks in `.git/hooks`.

I took a look and all the required linters are supported by pre-commit so I will open an alternative PR with that approach so you can compare and decide which approach is better for this project

justinmk:
> The project already breaks these guidelines by depending on 7 `eslint` and `prettier` related packages contributing ~3400 lines to the `package-lock.json` file

There is no objection at all to depending on linters. I originally thought this was only adding `lint-staged` and `conventional-commits`, but I see now that this also adds a dep on `stylua`, which I missed (it helps to make incremental PRs that do one thing at a time).

Can we skip `conventional-commits` for now? That is a trivial thing to "lint", it shouldn't require a local linter.

mbway:
I will amend this PR by removing husky, lintstaged and conventional-commits. I will create a `pre-commit` script that runs `npm run` commands one after the other.

This approach does not have the nice benefits of using a tool like [pre-commit the tool](https://pre-commit.com/), namely:

- it is slower, as linters have to be run on every file instead of only those staged
- output is less readable, as pre-commit summarizes which linters failed instead of printing the whole log
- if a linter fails on an unstaged file, the commit will still abort
    - pre-commit deals with this by 1) only running on staged files and 2) stashing and unstashing unstaged files

Here is a summary of what using pre-commit would look like. A single yaml file at the root of the repo would have to be added with the following content but there are no lock files or other dependencies.

```yaml
repos:
    - repo: local
      hooks:
          - id: keybind
            name: keybind
            entry: npm
            language: system
            args: ["run", "keybind"]
    - repo: https://github.com/pre-commit/pre-commit-hooks
      rev: v2.3.0
      hooks:
          - id: end-of-file-fixer
          - id: trailing-whitespace
    - repo: https://github.com/pre-commit/mirrors-eslint
      rev: "v8.56.0"
      hooks:
          - id: eslint
            additional_dependencies:
                - "eslint@8.57.0"
                - "@typescript-eslint/eslint-plugin@5.62.0"
                - "@typescript-eslint/parser@5.62.0"
                - "eslint-config-prettier@9.1.0"
                - "eslint-plugin-import@2.29.1"
                - "eslint-plugin-prettier@5.1.3"
    - repo: https://github.com/pre-commit/mirrors-prettier
      rev: "v3.1.0"
      hooks:
          - id: prettier
            args: ["--ignore-path", ".gitignore", "--ignore-path", ".prettierignore", "--write"]
    - repo: https://github.com/JohnnyMorganz/StyLua
      rev: v0.20.0
      hooks:
          - id: stylua-github
```

theol0403:
Sounds to me that pre-commit is a good balance between ease of use and minimalism. I would avoid re-inventing the wheel with a custom script. I like the yaml you shared.

mbway:
there are two ways that pre-commit can be used with an npm project. The first is how I have it now, where the linter versions are defined in the `.pre-commit-config.yaml` file and are installed into a separate environment by pre-commit. The alternative would be to have the linters as dependencies in `package.json` and have `local` hooks that run commands installed in `<project>/node_modules`. I'm not sure which is preferable.

.prettierrc.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
     semi: true,
     trailingComma: "all",
-    singleQuote: true,
```
mbway, line 4: duplicate key, I assume the last is taken

theol0403 (approved):
I like this new approach.

theol0403:
Make sure to fix CI: https://github.com/vscode-neovim/vscode-neovim/actions/runs/8402012282/job/23010961355?pr=1853#step:4:120

mbway:
I think the issue is that `test-compile` needs the project installed so I'll try installing it

theol0403:
Ready to merge after @justinmk 's review.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on: [pull_request, push]
+
+jobs:
+    lint:
+        name: Lint
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-python@v3
```
justinmk, line 11: Can we add a comment here mentioning why python is needed?

justinmk:
This turned out really nice, thank you!

.github/workflows/lint.yml
mbway: It originates from the suggested usage here: https://github.com/pre-commit/action?tab=readme-ov-file#using-this-action
I can open a PR adding this comment if you want

mbway:
This was tricky to get right but I'm glad we got there. Thank you both for your patience 🙂

.github/workflows/lint.yml
justinmk: Thanks! I added something in https://github.com/vscode-neovim/vscode-neovim/pull/1865

Issue #1760, opened 2024-01-21
feature: Inspect vscode state by expanding the lua API


mbway:
### Did you check the docs?

- [X] I have read all the vscode-neovim docs
- [X] I understand the fundamental limitations of the plugin and believe this request to be reasonable

### Is your feature request related to a problem? Please describe.

There are several situations where having more access to the vscode state in neovim/lua would be beneficial to configure
the editor in the ways that I want. (I hope I'm not missing something but I don't think this is possible yet)

For example, I would like to create a keybinding to toggle something but vscode only provides a pair of actions to do and undo something instead of also providing an action to toggle. I would like to be able to say:
```lua
-- toggle pinning the current tab
vim.keymap.set('n', '<leader>1', function()
    if tab_is_pinned() then
        vscode.action('workbench.action.unpinEditor')
    else
        vscode.action('workbench.action.pinEditor')
    end
end)

-- toggle between terminal and editor (hypothetical example.
-- I'm not sure if vscode neovim can handle keys when inside the terminal)
vim.keymap.set('n', '<C-t>', function()
    if terminal_is_active() then
        vscode.action('workbench.action.focusActiveEditorGroup')
    else
        vscode.action('workbench.action.terminal.focus')
    end
end)

```
where `terminal_is_active()` and `tab_is_pinned()` is what I am having difficulty with.

This is achieved in native vscode using the `editorTextFocus/terminalFocus` and `activeEditorIsPinned` context values respectively in the `when` part of a keybinding in `keybindings.json`.

It seems like the decision was made where extensions aren't able to access this context information: https://github.com/microsoft/vscode/issues/117980 but it looks like it would be possible to get some of the same information from the [vscode-api](https://code.visualstudio.com/api/references/vscode-api) eg `vscode.window.tabGroups.activeTabGroup.activeTab.isPinned` for `activeEditorIsPinned` and `vscode.window.activeTerminal != null` for `terminalFocus`

Another example is that I wanted to get the path of the document to be able to have a keybinding for running an autoformatting program. I wrote the following to parse the path from the neovim side:
```lua
function decode_percent_encoding(input)
    local converted = string.gsub(input, "%%(%x%x)", function(hex) return string.char(tonumber(hex, 16)) end)
    return converted
end

function current_path()
    local expected_prefix = '__vscode_neovim__-file://'
    local scratch_buffer_path = vim.fn.expand('%')
    local n = string.len(expected_prefix)
[...]

PR #1854, opened 2024-03-17
feat: `vscode.eval()`

mbway:
This PR introduces `vscode.eval()` which can be used to evaluate javascript passed as a string from lua. This can then be used to interact with the vscode API from neovim.

~~This PR introduces `vscode.get()` which can be used to obtain javascript values from lua. This is a simpler implementation than I originally imagined, but providing full API access including the ability to call arbitrary functions would be very complex and potentially error prone. `vscode.get()` is sufficient for the purposes of my original issue.~~

Fixes: https://github.com/vscode-neovim/vscode-neovim/issues/1760

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,18 +248,18 @@ local vscode = require('vscode-neovim')
 
 1. `vscode.action()`: asynchronously executes a vscode command.
 2. `vscode.call()`: synchronously executes a vscode command.
-3. `vscode.get()`: obtain a vscode API value.
-4. `vscode.on()`: defines a handler for some Nvim UI events.
-5. `vscode.has_config()`: checks if a vscode setting exists.
-6. `vscode.get_config()`: gets a vscode setting value.
-7. `vscode.update_config()`: sets a vscode setting.
-8. `vscode.notify()`: shows a vscode message (see also Nvim's `vim.notify`).
-9. `vscode.to_op()`: A helper for `map-operator`. See [code_actions.lua](./runtime/plugin/code_actions.lua) for the
+3. `vscode.on()`: defines a handler for some Nvim UI events.
+4. `vscode.has_config()`: checks if a vscode setting exists.
+5. `vscode.get_config()`: gets a vscode setting value.
+6. `vscode.update_config()`: sets a vscode setting.
+7. `vscode.notify()`: shows a vscode message (see also Nvim's `vim.notify`).
+8. `vscode.to_op()`: A helper for `map-operator`. See [code_actions.lua](./runtime/plugin/code_actions.lua) for the
    usage
-10. `vscode.get_status_item`: Gets a vscode statusbar item. Properties can be assigned, which magically updates the
-    statusbar item.
-11. `g:vscode_clipboard`: Clipboard provider using VSCode's clipboard API. Used by default when in WSL. See
+9. `vscode.get_status_item`: Gets a vscode statusbar item. Properties can be assigned, which magically updates the
+   statusbar item.
+10. `g:vscode_clipboard`: Clipboard provider using VSCode's clipboard API. Used by default when in WSL. See
     `:h g:clipboard` for more details. Usage: `let g:clipboard = g:vscode_clipboard`
+11. `vscode.eval()`: evaluate javascript in vscode and return the result
 
 ### vscode.action(name, opts)
 
@@ -301,7 +301,7 @@ Parameters:
 
 Returns: the result of the action
 
-#### Examples
+### Examples
 
 -   Format selection (default binding):
     ```vim
@@ -371,29 +371,6 @@ print(vscode.call("_wait", { args = { 1000 } })) -- outputs: ok
 -- Wait for 2 seconds with a timeout of 1 second
 print(vscode.call("_wait", { args = { 2000 } }), 1000)
 -- error: Call '_wait' timed out
-```
-
-### vscode.get(name)
-
-Obtain a [VSCode API](https://code.visualstudio.com/api/references/vscode-api) value.
-
-Parameters:
-
--   `name (string)`: The name of the variable to obtain
-    -   use `foo.bar` to access nested properties. e.g. `window.activeTextEditor.document.fileName`
-    -   use `.123` instead of `[123]` for array access e.g. `extensions.all.0`
-
-Returns:
-
--   return the value of the requested variable. Converted to a string if the value is not representable in lua (i.e.
-    "object" or "function")
-
-Examples:
-
-```lua
-local current_file = vscode.get("window.activeTextEditor.document.fileName")
-local current_tab_is_pinned = vscode.get("window.tabGroups.activeTabGroup.activeTab.isPinned")
-local workspace_uri = vscode.get("workspace.workspaceFolders.0.uri")
 ```
 
 ### vscode.on(event, callback)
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -248,16 +248,17 @@ local vscode = require('vscode-neovim')
 
 1. `vscode.action()`: asynchronously executes a vscode command.
 2. `vscode.call()`: synchronously executes a vscode command.
-3. `vscode.on()`: defines a handler for some Nvim UI events.
-4. `vscode.has_config()`: checks if a vscode setting exists.
-5. `vscode.get_config()`: gets a vscode setting value.
-6. `vscode.update_config()`: sets a vscode setting.
-7. `vscode.notify()`: shows a vscode message (see also Nvim's `vim.notify`).
-8. `vscode.to_op()`: A helper for `map-operator`. See [code_actions.lua](./runtime/plugin/code_actions.lua) for the
+3. `vscode.get()`: obtain a vscode API value.
```
xiyaowong, line 251: 11.

README.md
```diff
@@ -300,7 +301,7 @@ Parameters:
 
 Returns: the result of the action
 
-### Examples
+#### Examples
```
xiyaowong, line 304: We can place the 'example' under the third-level heading in the final section.

xiyaowong (commented):
My idea is to implement a method similar to `executeLua` in Neovim, which allows to execute a piece of code logic and get data. This can be achieved using "eval", "vm/sandbox" or other methods.

I suggest temporarily keeping 'get' and using a more specific name, such as 'get_vscode_value'. 'get' is too generic, and the current functionality doesn't quite match its name.

mbway:
I did attempt to see `eval` initially. I tried doing `eval("window")` and the result was null so I assumed it was sandboxed and moved on, but I may have been mistaken. It's worth another look. I should have time later in the week.

justinmk:
> I suggest temporarily keeping 'get' and using a more specific name, such as 'get_vscode_value'.

`evalVscode` or `eval<something>`



> which allows to execute a piece of code logic and get data. This can be achieved using "eval", "vm/sandbox" or other methods.

This could be a support burden. We may want to call out that use of this feature is not supported.

README.md
justinmk: yes, please don't add these micro-sections. it makes the doc too nested.

src/test/suite/index.ts
```diff
@@ -5,14 +5,21 @@ import Mocha from "mocha";
 import "source-map-support/register";
 
 export async function run(): Promise<void> {
+    let test_regex = process.env.VSCODE_NEOVIM_TEST_REGEX;
```
justinmk, line 8: `git grep -P 'process.env' shows that existing env vars (including test-only env vars) are named like `NEOVIM_xx`.

```suggestion
    let test_regex = process.env.NEOVIM_TEST_REGEX;
```

mbway:
I tried `eval` again and the issue I ran into originally seems specific to modules. In the global scope there is `import vscode, { ... } from "vscode"` and from `eval` the code cannot see the name `vscode`. However, variables and constants in the local and global scopes are visible.
So as a workaround I created a new `.ts` file specifically for controlling what names are in scope when running `eval` and did:
```typescript
import * as _vscode from "vscode";
const vscode = _vscode;
...
```
and that works! Javascript can be sent from neovim and run with `eval` and it can have access to the entire vscode api. This is much more useful than my original design so I will update the PR to use this approach instead.

justinmk:
> I suppose being able to run `eval` is enough to completely replace the existing API. For example, the functions like `vscode.has_config()` could be implemented in lua using the `eval` RPC internally.

Nice. That's definitely something to consider, though in some _very selective_ cases such as setting vscode options, it could make sense to have an "explicit" interface. But it's great that we can stop adding all of these shim APIs. Very nice!

README.md
mbway: I indented this section because it looks like it had the wrong indentation to be an example of using `vscode.call`. I don't know where the 'final section' is that you want me to move this to so I'm just going to undo the change for now, but I can move it if you want.

theol0403 (approved):
Amazing work! This looks great. I am happy to merge. However, if you want to take this opportunity to internally simplify some other commands (like `get_config`), that would be welcomed.

mbway:
ok. For now I'll just tidy up some trivial actions:
- `notify`
- `clipboard_read`
- `clipboard_write`

mbway:
I suppose, replacing a pre-defined action like `notify` with a call to `eval` on the lua side is not backwards compatible if the user is calling `vscode.action('notify')`. Maybe direct access to `vscode.action` should be deprecated before making those changes?

src/actions_eval.ts
xiyaowong, line 1: Why not just put it directly in the actions.ts ?

xiyaowong:
> I suppose, replacing a pre-defined action like `notify` with a call to `eval` on the lua side (so `require 'vscode-neovim'.notify()` uses `eval()` internally) is not backwards compatible if the user is calling `vscode.action('notify')` directly.

Maybe just replace simple data retrieval methods like clipboard_read and clipboard_write. I personally don't find eval elegant. Also, please note that there's a separate call to the notify method here.
https://github.com/vscode-neovim/vscode-neovim/blob/cf8636410648aa8784d4829e8386a18e903543b0/vim/vscode-neovim.vim#L24

> If I understand correctly, `vscode.action()` is performing two roles

`vscode.action` is the asynchronous version of `vscode.call`. So, the current implementation can also be used like this: `vscode.action('eval', ...)`.

xiyaowong (changes requested):

src/actions_eval.ts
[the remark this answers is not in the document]
mbway: To control which variables are in scope when the code is evaluated

mbway:
would it make sense to create a helper: `vscode.eval_async()` that uses `vscode.action` instead of `vscode.call`?

xiyaowong:
> would it make sense to create a helper: `vscode.eval_async()` that uses `vscode.action` instead of `vscode.call`? Or it could be an option like `vscode.eval("...", {..., async = true})`

Neither of these is necessary. In most use cases of eval, it's just about retrieving some simple data, and even if there's a need for execution, it's often fast. eval_async seems redundant.

theol0403 (approved):
After reading the discussion, my suggestion is exactly what has been implemented 👍

theol0403:
I see use cases for eval_async, especially since vim plugins now have the ability to directly interface with the vscode UI.

mbway:
Btw I just spotted an issue with the eval signature (timeout isn't listed). I can post a fix quickly if you want it fixed by the time of the release

src/actions_eval.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * - the variable `vscode` can be used to access the VSCode API
+ * - the variable `args` can be used to access the arguments passed from lua
+ *
+ * @param code the code to evaluate
+ * @param args arguments passed from lua
+ *
+ * @returns the result of evaluating the code, serialized to send back to lua
+ */
+export async function eval_for_client(code: string, args: any): Promise<any> {
```
justinmk, line 27: `args` is not used, is it still needed?

src/actions.ts
```diff
@@ -85,6 +76,7 @@ class ActionManager implements Disposable {
             await new Promise((resolve) => setTimeout(resolve, ms));
             return "ok";
         });
+        this.add("eval", (code: string, args: any) => eval_for_client(code, args));
```
justinmk, line 79: (off topic for this PR) `ActionManager` is just a wrapper around `registerCommand`.  Assuming it is still needed, it should be renamed to `Commands`. Introducing new jargon just makes things confusing.

src/actions_eval.ts
mbway: It is 'used' in the sense that it is accessible from the code being evaluated. This is used to pass data from lua. There are examples of this in the unit tests.

Issue #1864, opened 2024-03-26
bug: clipboard throws error ("vscode" not defined during eval())


brendonscript:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2698+g00e71d3da

### Operating system/version

WSL Arch Linux

### Describe the bug

Using the VSCode Neovim extension v1.8.0 with the latest LazyVim version any command that writes or reads from clipboard will return an error. Using any version prior does not return an error. Using the option `vscode-neovim.neovimClean = true` this issue does not occur. I figured this could be a LazyVim issue but was directed here when I posted the bug at LazyVim/LazyVim#2832

Using `yy`:
```
t read properties of null (reading 'env')
2024-03-26 12:11:57.404 [info] 
^I...liakov.vscode-neovim-1.8.0/runtime/modules/clipboard.lua:23: in function <...liakov.vscode-neovim-1.8.0/runtime/modules/clipboard.lua:20>
2024-03-26 12:15:11.527 [info] 
table: 0x7f222808cb78
2024-03-26 12:17:27.559 [info] 
nil
2024-03-26 12:20:06.372 [info] 
Already at oldest change
2024-03-26 12:20:07.082 [info] 
Already at oldest change
2024-03-26 12:20:12.083 [error] MainController: Request error:  Cannot read properties of null (reading 'env')
2024-03-26 12:20:12.085 [info] 
^I...liakov.vscode-neovim-1.8.0/runtime/modules/clipboard.lua:23: in function <...liakov.vscode-neovim-1.8.0/runtime/modules/clipboard.lua:20>
2024-03-26 12:20:13.029 [error] MainController: Request error:  Cannot read properties of null (reading 'env')
2024-03-26 12:20:13.030 [info] 
^I...liakov.vscode-neovim-1.8.0/runtime/modules/clipboard.lua:23: in function <...liakov.vscode-neovim-1.8.0/runtime/modules/clipboard.lua:20>
2024-03-26 12:20:32.746 [info] 
E518: Unknown option: g:clipboard
2024-03-26 12:20:55.939 [error] MainController: Request error:  Cannot read properties of null (reading 'env')
2024-03-26 12:20:55.940 [info] 
[...]

PR #1866, opened 2024-03-27
fix: vscode null in eval

mbway:
Prevent `vscode` from becoming `null` during build. Fixes: https://github.com/vscode-neovim/vscode-neovim/issues/1864

Observations:
- with `import vscode from "vscode"`, the name `vscode` is completely removed.
- with `import _vscode from "vscode"; const vscode = _vscode;`, the name `vscode` is set to null at build time.
- with `const vscode = require("vscode")`, everything works but `vscode` is identified as unused and linters prefer `import` over `require`.

Instead of relying on require to not have the global removed, I think the most robust solution is to make sure the static analysis tools don't think any of these variables are unused in the first place. The most correct approach in the presence of `eval()` is for the tools to assume nothing, but that isn't what happens in practice. This PR introduces a hack so that (basic) static analysis does not think that the variables we want to keep around are unused.

I have tested this approach including compiling to `.vsix` and `vscode` seems to now be set correctly.

### DIFF
--- a/src/actions_eval.ts
+++ b/src/actions_eval.ts
@@ -1,17 +1,11 @@
-/* eslint-disable @typescript-eslint/ban-ts-comment */
-/* eslint-disable @typescript-eslint/no-unused-vars */
-
-import * as _vscode from "vscode";
+import _vscode from "vscode";
 
 import { createLogger } from "./logger";
 
-// @ts-ignore
+const vscode = _vscode;
 const logger = createLogger("eval");
 
-// for some reason, the name used in the import statement is not always visible to the code being run with eval()
-// but global variables/constants are always visible.
-// @ts-ignore
-const vscode = _vscode;
+_use_variables([vscode, logger]);
 
 /**
  * Execute javascript code passed from lua in an async function context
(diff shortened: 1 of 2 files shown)


theol0403 (approved):
I'm happy with this.

src/actions_eval.ts
```diff
@@ -1,17 +1,11 @@
-/* eslint-disable @typescript-eslint/ban-ts-comment */
-/* eslint-disable @typescript-eslint/no-unused-vars */
-
-import * as _vscode from "vscode";
+import _vscode from "vscode";
 
 import { createLogger } from "./logger";
 
-// @ts-ignore
+const vscode = _vscode;
```
justinmk, line 5: `(void)foo` is enough to tell tools (or at least typescript compiler) that `foo` is used.  Maybe worth a try?
```suggestion
const vscode = _vscode;
(void)vscode;
```

mbway: Perfect. I wasn't sure of the best way to mark a variable as used in JavaScript and this is much better

mbway:
using `void` statements, everything seems to work. The linters do not consider the names unused and the names are not removed in the `.vsix` file. However the void statements themselves are removed by the build step so in the minified javascript the names are considered unused. But that's probably fine.

PR #1894, opened 2024-04-20
feat: Auto find a suitable neovim executable

### LINKED ISSUE #195: Feature request: New install experience
I would like to improve on the first-install experience.

1. Have vscode-neovim.neovimPath be 'nvim' and search $PATH for it by default.
2. If nvim is found but have the wrong version. ERROR OUT and show an error on start.

This plugin requires me to always open up the config to use the tool when I install vscode with the neovim ext. on a new computer. Which I do.. often! And equally often I will screw up on the configuration and wonder why all keyboard input is being dropped. After several restarts and WTFs I realize that "oh its called nvim not neovim" or "oh neovim was old on this system I need to install it out of tree" or something else.

xiyaowong:
Close #195

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -1,7 +1,7 @@
 import { ChildProcess, spawn } from "child_process";
 import path from "path";
 
-import { attach, NeovimClient } from "neovim";
+import { attach, findNvim, NeovimClient } from "neovim";
 import vscode, { Disposable, Range, window, type ExtensionContext } from "vscode";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { transports as loggerTransports, createLogger as winstonCreateLogger } from "winston";
@@ -11,6 +11,7 @@ import { BufferManager } from "./buffer_manager";
 import { CommandLineManager } from "./command_line_manager";
 import { CommandsController } from "./commands_controller";
 import { config } from "./config";
+import { NVIM_MIN_VERSION } from "./constants";
 import { CursorManager } from "./cursor_manager";
 import { DocumentChangeManager } from "./document_change_manager";
 import { eventBus } from "./eventBus";
@@ -104,6 +105,7 @@ export class MainController implements vscode.Disposable {
         this.client.on("request", this.onNeovimRequest);
         this.setClientInfo();
         await this.client.setVar("vscode_channel", await this.client.channelId);
+        await this.client.setVar("vscode_nvim_min_version", NVIM_MIN_VERSION);
 
         // This is an exception. Should avoid doing this.
         Object.defineProperty(actions, "client", { get: () => this.client, configurable: true });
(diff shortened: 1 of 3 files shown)


src/main_controller.ts
```diff
@@ -184,8 +186,21 @@ export class MainController implements vscode.Disposable {
             }
         }
 
+        let neovimPath = config.neovimPath;
+        // Only try to find nvim if the path is the default one
+        // And if we are not using WSL
```
justinmk, line 191: could findNvim() be enhanced to work with WSL? PR welcome at https://github.com/neovim/node-client/

src/main_controller.ts
```diff
@@ -184,8 +186,21 @@ export class MainController implements vscode.Disposable {
...
+        if (neovimPath === "nvim" && !config.useWsl) {
+            const nvimResult = findNvim({ minVersion: NVIM_MIN_VERSION });
+            logger.debug("Find nvim result: ", nvimResult);
```
justinmk, line 194: should be info()

src/main_controller.ts
```diff
@@ -104,6 +105,7 @@ export class MainController implements vscode.Disposable {
         this.client.on("request", this.onNeovimRequest);
         this.setClientInfo();
         await this.client.setVar("vscode_channel", await this.client.channelId);
+        await this.client.setVar("vscode_nvim_min_version", NVIM_MIN_VERSION);
```
justinmk, line 108: instead of adding more global vars can we change `g:vscode` to be a dict, and store things in its keys

justinmk (approved):
I have a local branch that does this, but this is fine for now.

`findNvim()` might change to async, that's one reason why I didn't do this yet

src/main_controller.ts
xiyaowong: I personally think it's not very meaningful. The use case for nvim in WSL is limited, and WSL essentially functions like a virtual machine.

src/main_controller.ts
xiyaowong: But it’s not convenient to update some data, right? Sometimes variables are not defined at the same time

xiyaowong:
The time taken by findNvim isn't too long, which is acceptable. However, I hope findNvim can provide an option to stop searching when it finds a valid nvim. 🤔

src/main_controller.ts
justinmk: untested, but just calling `vim.g.vscode = vim.g.vscode or {}` first is a good enough pattern (eventually Nvim core should make this easier):

```typescript
await this.client.lua("vim.g.vscode = vim.g.vscode or {}");
await this.client.lua("vim.g.vscode.channel = ...", await this.client.channelId);
await this.client.lua("vim.g.vscode.nvim_min_version", = ...", NVIM_MIN_VERSION);
```

xiyaowong: In this case, there are quite a few places that need to be modified. If refactoring is required, a separate PR would be more appropriate.

xiyaowong: > Untested:

I just tested it and it doesn't work properly

justinmk: > I just tested it and it doesn't work as expected.

what exactly does that mean?

xiyaowong: image

justinmk: Oh that's because the container is marshalled. So it will need to be vimscript intsead of lua.

PR #1895, opened 2024-04-20
feat: handle nvim statusline

### LINKED ISSUE #1809: feature: support for statusline plugins
### Did you check the docs?

Currently, there does not seem to be support for statusline plugins such as [lualine](https://github.com/nvim-lualine/lualine.nvim).

It would be awesome to either have these available, or to have greater integration with the existing VSCode status bar. For example, [the plugin](https://github.com/kensleDev/vscode-nvim-ui) mentioned in this issue https://github.com/vscode-neovim/vscode-neovim/issues/1224 appears to no longer be supported. This appears to be something that is offered in VsCodeVim.

Support for statusline plugins, or greater integration with status bar.

Use existing status bar.

### Additional context

_No response_

xiyaowong:
Resolve #1809

"get_status_item" has been removed.

This is a relatively simple implementation at the moment, with a few points to discuss:
1. ~~Setting a different background color.~~ Be simple
2. ~~Making the refresh interval configurable.~~ Be simple
3. ~~In nvim, detecting changes in the statusline and then notifying vscode for updates can reduce RPC calls and make refreshes more timely. However, this approach leads to less concise code.~~ RPC calls are cheap.


justinmk:
Why is this needed? Can users use `vscode.get_status_item` to set the statusline? Can we document some Lua config they can add to achieve that?

> Making the refresh interval configurable.

Please, no.

> detecting changes in the statusline and then notifying vscode for updates can reduce RPC calls

Premature optimization. RPC calls are cheap.

xiyaowong:
> Why is this needed? Can users use `vscode.get_status_item` to set the statusline to the result of `nvim_eval_statusline` ? Can we document some Lua config they can add to achieve that?

There are two differences:
1. This works out of the box.
2. I put the statusline on the far left, the status_item on the far right, and the priority between multiple status_items is uncertain.

My personal understanding is that `get_status_item` provides a simple API for manipulating the start bar of VSCode, while the content added in this PR deals with the statusline of Neovim.

I am neutral about this PR

justinmk:
Alternatively, after this PR, can we remove `get_status_item` ?

xiyaowong:
I suggest not. There are already plugins using this API. 😶

justinmk:
> There are already plugins using this API. 😶

That really doesn't matter. The worst case scenario is simply that people get the default statusline. That doesn't break anyone.

xiyaowong:
Let's keep this for now, because it's indeed more convenient compared to statusline. Here are the current use cases. It would be difficult to achieve the same effect without letting users configure it if we were to use statusline.
 
- https://github.com/vscode-neovim/vscode-multi-cursor.nvim/blob/67e101404db09a26acdfccb3a0fdbcc49768b834/lua/vscode-multi-cursor/state.lua#L10
- https://github.com/folke/flash.nvim/pull/269

justinmk:
> because it's indeed more convenient compared to statusline

How is it more convenient? Nvim 'statusline' allows users to configure their statusline. If users want to set the vscode statusline, they could just configure the Nvim 'statusline'.

> It would be difficult to achieve the same effect without letting users configure it if we were to use statusline.

? https://github.com/folke/flash.nvim/pull/269 is just setting the `.text` field.  Why do we need a separate API for that, which only avoids users configure Nvim 'statusline'?

xiyaowong:
After careful consideration, you are correct. I was thinking from the perspective of a vscode-neovim user and being familiar with the `get_status_item` API, which was unreasonable. For regular plugin authors, this adds maintenance overhead, and for users, it is often unnecessary.

src/status_line_manager.ts
```diff
@@ -33,28 +31,38 @@ export class StatusLineManager implements Disposable {
...
-                    this._cmdText = status;
-                    break;
-                case StatusType.Msg:
-                    this._msgText = status;
-                    break;
+        const refreshNvimStatusLineTimer = setInterval(async () => {
+            let sl = (await this.client.lua(`
+            if vim.o.laststatus == 0 then return "" end
+            return vim.api.nvim_eval_statusline(vim.o.statusline, {}).str
+            `)) as any as string;
```
justinmk, line 39: Just FYI: after the next node-client release, the `as any` can be dropped. https://github.com/neovim/node-client/pull/353

justinmk (approved):
Beautiful!

runtime/lua/vscode-neovim/api.lua
```diff
@@ -418,64 +418,4 @@ do
...
-  ---```lua
-  ---local test = vscode.get_status_item('test')
-  ---test.text = 'hello' -- Show the text
-  ---test.text = '' -- Hide the item
-  ---test.text = nil -- Close the item
-  ---test.text = '' -- error: The status item "test" has been closed
-  ---```
-  ---
-  ---@param id string The identifier of the item
-  function M.get_status_item(id)
```
justinmk, line 452: to help users migrate, we could leave this as an empty function for a couple releases, and show an info message.

```lua
function M.get_status_item()
    M.notify("Nvim statusline is now shown in vscode automatically. get_status_item was removed.")
end
```

src/status_line_manager.ts
```diff
@@ -33,28 +31,38 @@ export class StatusLineManager implements Disposable {
...
-                case StatusType.Cmd:
-                    this._cmdText = status;
-                    break;
-                case StatusType.Msg:
-                    this._msgText = status;
-                    break;
+        const refreshNvimStatusLineTimer = setInterval(async () => {
+            let sl = (await this.client.lua(`
+            if vim.o.laststatus == 0 then return "" end
+            return vim.api.nvim_eval_statusline(vim.o.statusline, {}).str
```
justinmk, line 38: 'statusline' is blank by default, which then uses Nvim's internal statusline logic. That's a historical quirk of Vim, not really up to vscode-neovim to work around it. Tracking issue: https://github.com/neovim/neovim/issues/28444

xiyaowong: > When 'statusline' option is blank, Nvim still shows a statusline. In that case, the vscode statusline will be blank even though Nvim's _rendered_ statusline is not blank.

I thought about this, I don't think anyone needs this default statusline

> OTOH, it's probably good that users can disable this vscode-neovim behavior by setting 'statusline' to empty

Yes. So I removed the settings for the laststatus and statusline options instead of setting a default value :)

PR #1897, opened 2024-04-20
fix: tabstops now considered when calculating line width for highlights

### LINKED ISSUE #1766: bug: Scramble when highlighted

### Neovim version (nvim -v)

0.9.2

### Operating system/version

win 11

### Describe the bug

In the C language environment,When I search for a word using *,It will have garbled characters


bug

### Steps To Reproduce

nomal mode

### Expected Behavior

fix this bug

ollien:
Fixes #1766 

It might be easiest to review this commit by commit, rather than as a full diff

There was a small bug in the `HighlightProvider` which incorrectly calculated with width of tab chars (it assumed all tabs were equal to the tab width, but this is not always the case, as tabs move to the next tab stop). This is what lead to the "scrambling" we see in #1766, as highlights were offset by the "excess" tab width.

Part of my test-adding process involved breaking up the `utils` file. Unfortunately, I couldn't import "utils" in unit tests, due to the use of the `vscode` module, and leaving a stray function in its own file didn't feel right. Sorry for blowing up the PR length of what is ultimately a simple fix :sweat_smile:. I want to go back and add some more tests around other parts of `HighlightProvider`, but I already made this PR big enough :)

https://github.com/vscode-neovim/vscode-neovim/assets/977151/8c53fd60-680c-44e0-a0ee-5940b0a1cfd5


src/utils/nvim.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ILogger, LogLevel } from "../logger";
+
+/**
+ * Wrap nvim callAtomic and check for any errors in result
+ * @param client
+ * @param requests
+ * @param logger
+ * @param prefix
+ */
+export async function callAtomic(
```
justinmk, line 12: (off topic) we could eliminate this and `executeLua` instead

justinmk (approved):

src/utils/nvim.ts
ollien: Interesting - I haven't been deep into neovim internals before, so this is a bit new to me. Is there any chance of only some operations completing if were to convert a block like this to use many statements in `executeLua`? I only ask because the docs for the neovim client say that `callAtomic` has some guarantees around atomicity, and I don't see that for `executeLua`

https://github.com/vscode-neovim/vscode-neovim/blob/master/src/buffer_manager.ts#L605-L620

justinmk: The "atomicity" of https://neovim.io/doc/user/api.html#nvim_call_atomic() is mostly just about performing multiple things in a single RPC request. Since `nvim_exec_lua` was introduced it is no longer needed.

theol0403 (approved):

PR #1901, opened 2024-04-21
refactor: extract highlight provider grid logic into its own data structure

ollien:
There's a bunch of things I want to do before undrafting this, but my goal is to move all "grid calculations" into an isolated data structure -- `HighlightGrid` -- rather than the `HighlightProvider`. By separating the logic of providing the highlights to VSCode from the logic to keep track of the highlights, we can much more easily unit test the highlight logic. You'll notice that most of the code in `HighlightGrid` is just a copy/paste from the previous `HighlightProvider`; this is deliberate, and I'm thinking of keeping it mostly the same for now, just not to change too much at once.

Current TODO
- [ ] Add a more functional way to get highlight ranges out of the `HighlightGrid`
- [ ] Add unit tests around some basic highlighting cases
- [ ] Use this new highlight range extraction to build highlight decorations
- [ ] (Maybe another PR?) simplify the old highlight building logic from the `HighlightProvider`

### DIFF
--- a/src/highlight_provider.ts
+++ b/src/highlight_provider.ts
@@ -1,6 +1,4 @@
-import GraphemeSplitter from "grapheme-splitter";
 import { cloneDeep } from "lodash-es";
-import wcswidth from "ts-wcwidth";
 import {
     DecorationOptions,
     DecorationRangeBehavior,
@@ -13,7 +11,7 @@ import {
     window,
 } from "vscode";
 
-import { expandTabs } from "./utils";
+import { getWidth } from "./utils";
 import { config } from "./config";
 import { Highlight, HighlightGrid, ValidCell } from "./highlight_grid";
 
@@ -99,16 +97,6 @@ function normalizeDecorationConfig(config: ThemableDecorationRenderOptions): The
     newConfig.overviewRulerColor = normalizeThemeColor(newConfig.overviewRulerColor);
     return newConfig;
 }
-
-// 你 length:1 width:2
-// 🚀 length:2 width:2
-// 🕵️ length:3 width:2
-// ❤️ length:2 width:1
-const isDouble = (c?: string) => wcswidth(c) === 2 || (c ?? "").length > 1;
-const segment: (str: string) => string[] = (() => {
-    const splitter = new GraphemeSplitter();
-    return (str) => splitter.splitGraphemes(str);
-})();
 
 export class HighlightProvider implements Disposable {
     /**
@@ -209,16 +197,10 @@ export class HighlightProvider implements Disposable {
         }
         const gridHl = this.highlights.get(grid)!;
 
-        // TODO: Remove this from here
-        const getWidth = (text?: string) => {
-            const t = expandTabs(text ?? "", tabSize);
-            return segment(t).reduce((p, c) => p + (isDouble(c) ? 2 : 1), 0);
-        };
-
         // TODO: Break this out somehow
         const validCells: ValidCell[] = [];
         {
-            const idealMaxCells = Math.max(0, getWidth(lineText) - vimCol);
+            const idealMaxCells = Math.max(0, getWidth(lineText, tabSize) - vimCol);
             const currMaxCol = gridHl.maxColInRow(row);
             const maxValidCells = Math.max(idealMaxCells, currMaxCol);
             const eolCells: ValidCell[] = [];
--- a/src/highlight_provider.ts
+++ b/src/highlight_provider.ts
@@ -13,8 +13,9 @@ import {
     window,
 } from "vscode";
 
-import { calculateEditorColFromVimScreenCol, expandTabs } from "./utils";
+import { expandTabs } from "./utils";
 import { config } from "./config";
+import { Highlight, HighlightGrid, ValidCell } from "./highlight_grid";
 
 export interface VimHighlightUIAttributes {
     foreground?: number;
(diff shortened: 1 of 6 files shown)


src/highlight_provider.ts
```diff
@@ -452,7 +267,7 @@ export class HighlightProvider implements Disposable {
 
         const gridHl = this.highlights.get(grid);
         if (gridHl) {
-            gridHl.forEach((rowHighlights, row) => {
```
ollien, line 455: Github won't let me put this comment on the line I want, but one question I have for those more familiar with this code: is the highlight created by `createColVirtTextOptions` strictly for virtual text? I only ask because on line 467, the check tests for either `colHighlights.length > 1` or `colHighlights[0].virtText`, which implies to me there may be some other case other than virtual text where this gets triggered. If there is, what is it?

ollien: @xiyaowong Sorry for the ping - if you have a moment and wouldn't mind weighing in here, that would be appreciated. The blame says this came from your refactor, so I figure if anyone would know, it's you :)

xiyaowong: The virtText and the original text of cell is the same.
- `colHighlights.length > 1` indicates virtual text on multi-width characters, which include 2 cells.
- `colHighlights[0].virtText` indicates virtual text on single-width characters, and this value is used to distinguish normal highlights from virtual text.
The above two conditions essentially represent "isVirtText".

xiyaowong: > is the highlight created by `createColVirtTextOptions` strictly for virtual text?

Yes.

> the check tests for either `colHighlights.length > 1` or `colHighlights[0].virtText`

The virtText and the original text of cell is the same.

- `colHighlights.length > 1` indicates virtual text on multi-width characters, which include 2 cells.
- `colHighlights[0].virtText` indicates virtual text on single-width characters, and this value is used to distinguish normal highlights from virtual text.

The above two conditions essentially represent "isVirtText".

xiyaowong:
highlight_grid I think is good.

However, a reminder, the granularity of the highlight module is moderate, and the functions, data structures, etc. inside will basically not be reused by other modules, so please consider splitting them all into utils as appropriate.

ollien:
@xiyaowong Sorry, what do you mean by granularity?

That said, yeah, I kinda waffled a bit on where to put those. I was thinking of making a separate `highlight` directory, rather than shoving them in `utils`, but I'm certainly flexible

src/highlight_provider.ts
theol0403: Off the top of my head, this might also support decorators that don't need to be virtual text but that workaround vscode limitations, like text/fake cursor after EOL.

ollien: Do you know of a situation where text after EOL would be used? Just trying to take in all of this

xiyaowong: Just consider that any text that is inconsistent with the original text is treated as virtual text.

theol0403: > Do you know of a situation where text after EOL would be used? Just trying to take in all of this

Onemore visual cursor, where the fake visual cursor is placed after eol. 

Leap.nvim, if the label is on the last character.

ollien:
Ok, I'm going to undraft this. There's _a lot_ of code around the tests (mostly white space, tbh), so I figure I'm going to leave it at that for now.

Overall, there's pretty decent code coverage (c8 tells me about 89% for `highlight_grid.ts`). The only thing that's really uncovered is this span of lines, because honestly I couldn't figure out what it was for exactly, and didn't want to shoe-horn cases for something I didn't understand. If you have an idea for a case here, I'm more than happy to apply it!

https://github.com/ollien/vscode-neovim/blob/eccbdd2310ce5129fdc4ee72f0dc2c6c5e449f95/src/highlights/highlight_grid.ts#L83-L95

src/utils/text.ts
```diff
@@ -17,6 +18,11 @@ export function expandTabs(line: string, tabWidth: number): string {
     return expanded;
 }
 
+export function splitGraphemes(str: string): string[] {
+    const splitter = new GraphemeSplitter();
```
xiyaowong, line 22: I remember the singleton I used before.

src/highlights/highlight_provider.ts
```diff
@@ -532,7 +333,7 @@ export class HighlightProvider implements Disposable {
         colHighlights.forEach(({ virtText, hlId, text }) => {
             // In certain edge cases, the right-side highlight may be appended later,
             // resulting in the column being converted to virt text type.
-            // So, the left-side highlight may not include virtText.
+            // So, the left-side highlight may not include vrtText.
```
xiyaowong, line 336: ? virtText

xiyaowong:
>  (especially for `leftHls`)
@ollien

src/utils/text_cells.ts
xiyaowong, line 1: I still think isDouble, getWidth, splitGraphemes should be put into highlight_grid, either as instance methods or static methods.

ollien: Yeah sure, I can do that. I just figured they'd be free functions but I don't feel strongly one way or the other

src/highlights/highlight_provider.ts
ollien: Heh, mistaken backspace

theol0403 (approved):
Oooh love this test-writing style.

src/utils/text.ts
ollien: Oh huh, I totally missed that subtlety with it being an IIFE. Nice!

ollien:
@xiyaowong Thank you for your gif above - it was very helpful :)

PR #1904, opened 2024-04-24
refactor: remove use of nvim_call_atomic

justinmk:
nvim_call_atomic is deprecated since nvim_exec_lua can be used to batch calls and handle errors.

btw, i get this when i run `npm run lint` locally, now

```
> vscode-neovim@1.10.0 format
> pre-commit run --all prettier && pre-commit run --all stylua-github
sh: pre-commit: command not found
```

even after running `npm install -g pre-commit`  (which, really, I shouldn't need to do)


justinmk:
CI not triggering...

src/utils/nvim.ts
```diff
@@ -1,38 +0,0 @@
...
-    }
-    const returned_errors = res[1];
-    if (returned_errors) {
-        const [failing_call_idx, err_type, err_msg] = returned_errors;
-        const call = requests[failing_call_idx];
-        const requestName = call[0];
-        const errMsg = `${requestName}: ${err_msg} (Error type: ${err_type})`;
-        // TODO: Determine cause for errors for both of these requests
-        if (requestName === "nvim_input" || requestName === "nvim_win_close") {
-            logger.log(undefined, LogLevel.warn, errMsg);
```
theol0403, line 33: Don't we lose our logs here?

src/buffer_manager.ts
```diff
@@ -602,22 +602,34 @@ export class BufferManager implements Disposable {
...
+            local bufId, lines, vscode_editor_options, docUri, docUriJson, bufname, isExternalDoc = ...
+            vim.api.nvim_buf_set_lines(bufId, 0, -1, false, lines)
+            -- set vscode controlled flag so we can check it neovim
+            vim.api.nvim_buf_set_var(bufId, "vscode_controlled", true)
+            -- In vscode same document can have different insertSpaces/tabSize settings per editor
+            -- however in neovim it's per buffer. We make assumption here that these settings are same for all editors
+            vim.api.nvim_buf_set_var(bufId, "vscode_editor_options", vscode_editor_options)
+            vim.api.nvim_buf_set_var(bufId, "vscode_uri", docUri)
+            vim.api.nvim_buf_set_var(bufId, "vscode_uri_data", docUriJson)
+            vim.api.nvim_buf_set_name(bufId, bufname)
```
justinmk, line 616: I confirmed that these variables are set as expected.

PR #1915, opened 2024-04-27
refactor: simplify highlight range building logic

ollien:
Just a small refactor :) It also cleans up one of the tests that I felt was a bit weird, but AFAICT has no impact on functionality

### DIFF
--- a/src/highlights/highlight_grid.ts
+++ b/src/highlights/highlight_grid.ts
@@ -271,55 +271,48 @@ export class HighlightGrid {
     }
 
     private rangesForRow(rowHighlights: Highlight[][], row: number, topLine: number): HighlightRange[] {
-        const res: HighlightRange[] = [];
-
         const line = row + topLine;
-        let currHighlight: Highlight | null = null;
-        let currStartCol = 0;
-        let currEndCol = 0;
+
+        const normalHighlights: Map<number, NormalTextHighlightRange> = new Map();
+        const virtualHighlights: VirtualTextHighlightRange[] = [];
         rowHighlights.forEach((colHighlights, col) => {
+            if (colHighlights.length === 0) {
+                // Should never happen, but defensive
+                return;
+            }
+
             if (colHighlights.length > 1 || colHighlights[0].virtText) {
-                res.push({
+                virtualHighlights.push({
                     textType: "virtual",
                     highlights: colHighlights,
                     line,
                     col,
                 });
-            } else {
-                // Extend range highlights
-                const highlight = colHighlights[0];
-                if (currHighlight?.hlId === highlight.hlId && currEndCol === col - 1) {
-                    currEndCol = col;
-                } else {
-                    if (currHighlight) {
-                        res.push({
-                            textType: "normal",
-                            hlId: currHighlight.hlId,
-                            line,
-                            startCol: currStartCol,
-                            endCol: currEndCol + 1,
-                        });
-                    }
+                return;
+            }
 
-                    currHighlight = highlight;
-                    currStartCol = col;
-                    currEndCol = col;
-                }
+            const colHighlight = colHighlights[0];
+            const existingHighlight = normalHighlights.get(colHighlight.hlId);
+            if (existingHighlight !== undefined) {
+                // Extend our existing highlight if we already have it
+                existingHighlight.endCol = col + 1;
+            } else {
+                const highlight = {
+                    textType: "normal" as const,
+                    hlId: colHighlight.hlId,
+                    line,
+                    startCol: col,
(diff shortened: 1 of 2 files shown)


ollien:
Huh. I don't have a Mac so I can't try and reproduce that test failure. Is that test flaky?

xiyaowong:
Something broken
image

ollien:
@xiyaowong I'll take a look. Thanks

xiyaowong:
> Huh. I don't have a Mac so I can't try and reproduce that test failure. Is that test flaky?

It's OK, macOS often fails inexplicably, just re-run it :)

xiyaowong:
> Something broken 

@ollien It is necessary to add tests for this. Currently, highlighting is mainly done manually, which is quite troublesome.

ollien:
It's like you read my mind :)

ollien:
Ah. It looks like I made a bad assumption that a single highlight id would be contiguous. In a test on my machine, your example groups the two "hello"s on the second line under the same highlight group. That's certainly unexpected to me, but I guess my assumption is never stated in the neovim docs. Will address

src/highlights/highlight_grid.ts
```diff
@@ -313,6 +269,61 @@ export class HighlightGrid {
...
+                    highlights: colHighlights,
+                    line,
+                    col,
+                });
+                return;
+            }
+
+            const colHighlight = colHighlights[0];
+            const existingHighlights = normalHighlights.get(colHighlight.hlId) ?? [];
+            const matchingHighlight = rfind(
```
xiyaowong, line 296: findLast?

src/highlights/highlight_grid.ts
```diff
@@ -292,10 +292,18 @@ export class HighlightGrid {
...
             const colHighlight = colHighlights[0];
-            const existingHighlight = normalHighlights.get(colHighlight.hlId);
-            if (existingHighlight !== undefined) {
+            const existingHighlights = normalHighlights.get(colHighlight.hlId) ?? [];
+            const matchingHighlight = rfind(
+                existingHighlights,
+                (hl) =>
+                    hl.endCol === col ||
+                    // double characters' start columns are offset by one (i.e. col is one higher)
+                    (HighlightGrid.isDouble(colHighlight.text) && hl.endCol === col - 1),
```
xiyaowong, line 301: I don't get it. Wasn't the highlight already processed?

xiyaowong (changes requested):

src/highlights/highlight_grid.ts
ollien: Sure. I just stole the name from Python, haha

src/highlights/highlight_grid.ts
ollien: This is actually how I fixed this test

https://github.com/vscode-neovim/vscode-neovim/pull/1915/files#diff-170234a32b0ecf69f0c4d9fa8465ac6542ecc30db6ada74a9dd1754e16a3b96dL200-L208

In that test, once we get to the ` `, `col` is 13. Once we hit the rocket emoji, `col` jumps to 15, due to the fact the char is double-wide, and `hl.endCol` at this point is 14, because we just processed column 13. If we simply checked `hl.endCol === col`, this would produce two separate highlight ranges, so we make the adjustment for this double-wide char.

Perhaps there's some argument to be made that the previously calculated highlight is wrong? Stepping through `processHighlightCellsEvent`, the additional width is coming from this check, which I suppose is right by coincidence, as ` "🚀".length === 2`

https://github.com/vscode-neovim/vscode-neovim/blob/3c82cc01e69e0817eb92f7093844136a3038bcff/src/highlights/highlight_grid.ts#L151-L152

ollien: On second thought, I think you're right... I think this is just compensating for a bug in calculating the highlight ranges. I'll go ahead and revert this. I'm actually tracking down another bug (predating this PR) that might stem from the same problem, so I'll fix this there if I can.

src/highlights/highlight_grid.ts
xiyaowong: I mean `existingHighlights.findLast`

ollien: Oh, huh.  I couldn't find that before when I went looking for it

ollien: Ah, ok. `findLast` is only available in ES2023, and we target ES2017. We could change that I guess? Not sure the implications of doing so to be honest

xiyaowong (approved):

xiyaowong (approved):

PR #1917, opened 2024-04-28
feat!: better composite key support

### LINKED ISSUE #330: Composite escape handler doesn't check if another non escape character was typed before

### LINKED ISSUE #460: Configurable composite escape timeout

### LINKED ISSUE #493: Is there a way to map jj to Escape and Save
I'd like to replicate this map `inoremap jj <ESC>:w<CR>`

### LINKED ISSUE #563: More generic composite key escape method (e.g. for jj/jk/kj etc)

### LINKED ISSUE #644: Composite Escape breaks macro recording

xiyaowong:
The main idea is to always register the `type` command, but when there is no need to take over user input (generally in insert mode), forward the received inputs to the `default:type` command.

The implementation is to judge the input before forwarding. If it matches, call the configured command, otherwise continue to forward to the `default:type`

At present, this implementation is normal, but it feels that forwarding input causes a slight delay in input (this may be a psychological effect). Need help testing this. **If there is indeed a delay, consider forwarding inputs only when the user needs the composite keys. If the user does not need to use composite keys, then just dispose the command when there is no need to take over user input**. ( Even if there is a delay, it is very slight.)

---

New configurations:
- compositeTimeout
- compositeKeys

Removed commands: `vscode-neovim.compositeEscape1` `vscode-neovim.compositeEscape2`

**Usage**

1. jj to escape
```json
{
    "vscode-neovim.compositeKeys": {
        "jj": {
            "command": "vscode-neovim.escape"
        }
    }
}
```
2. jk to escape and save
```json
{
    "vscode-neovim.compositeKeys": {
        "jk": {
            "command": "vscode-neovim.lua",
            "args": ["vim.api.nvim_input('<ESC>')\nrequire('vscode-neovim').action('workbench.action.files.save')"]
        }
    }
}
```

---

- close #563
- close #460
- close #330
- close #493
- close #644
- close #892

### DIFF
--- a/package.json
+++ b/package.json
@@ -400,6 +400,14 @@
                                 "command": "vscode-neovim.escape",
                                 "args": []
                             }
+                        },
+                        {
+                            "jk": {
+                                "command": "vscode-neovim.lua",
+                                "args": [
+                                    "vim.api.nvim_input('<ESC>') require('vscode-neovim').action('workbench.action.files.save')"
+                                ]
+                            }
                         }
                     ],
                     "patternProperties": {
@@ -417,7 +425,8 @@
                                 "command"
                             ]
                         }
-                    }
+                    },
+                    "description": "Composite key bindings"
                 }
             }
         },
--- a/package.json
+++ b/package.json
@@ -389,6 +389,8 @@
                 "vscode-neovim.compositeTimeout": {
                     "type": "number",
                     "default": 300,
+                    "minimum": 100,
+                    "maximum": 1000,
                     "description": "Timeout in milliseconds to wait for the next key in a composite key sequence"
                 },
                 "vscode-neovim.compositeKeys": {
--- a/package.json
+++ b/package.json
@@ -385,6 +385,39 @@
                     "type": "string",
                     "default": "|",
                     "description": "Separator used by statusline"
+                },
+                "vscode-neovim.compositeTimeout": {
+                    "type": "number",
+                    "default": 300,
+                    "description": "Timeout in milliseconds to wait for the next key in a composite key sequence"
+                },
+                "vscode-neovim.compositeKeys": {
+                    "type": "object",
+                    "default": {},
+                    "examples": [
+                        {
+                            "jj": {
+                                "command": "vscode-neovim.escape",
+                                "args": []
+                            }
+                        }
+                    ],
+                    "patternProperties": {
+                        "^[a-zA-Z]{2}$": {
+                            "type": "object",
+                            "properties": {
+                                "command": {
+                                    "type": "string"
+                                },
+                                "args": {
+                                    "type": "array"
+                                }
+                            },
+                            "required": [
+                                "command"
+                            ]
+                        }
+                    }
                 }
             }
         },
@@ -412,14 +445,6 @@
             {
                 "command": "vscode-neovim.escape",
                 "title": "Neovim: Send escape key to neovim, or other keys that will exit insert mode"
-            },
-            {
-                "command": "vscode-neovim.compositeEscape1",
-                "title": "Neovim: Composite escape key 1"
-            },
-            {
-                "command": "vscode-neovim.compositeEscape2",
-                "title": "Neovim: Composite escape key 2"
             },
             {
                 "command": "vscode-neovim.commit-cmdline",
--- a/src/test/integ/composite-escape.test.ts
+++ b/src/test/integ/composite-escape.test.ts
@@ -1,7 +1,8 @@
+import { strict as assert } from "assert";
+
 import { NeovimClient } from "neovim";
 import vscode from "vscode";
 
-import { strict as assert } from "assert";
 import {
     assertContent,
     attachTestNvimClient,
@@ -11,7 +12,7 @@ import {
     sendEscapeKey,
     sendInsertKey,
     sendVSCodeKeys,
-    wait
+    wait,
 } from "./integrationUtils";
 
 describe("Composite escape key", () => {
--- a/src/test/integ/composite-escape.test.ts
+++ b/src/test/integ/composite-escape.test.ts
@@ -23,7 +23,7 @@ describe("Composite escape key", () => {
         await closeAllActiveEditors();
     });
 
-    it("Works", async () => {
+    it.skip("Works", async () => {
         await openTextDocument({ content: "" });
 
         await sendInsertKey();
--- a/src/test/integ/composite-escape.test.ts
+++ b/src/test/integ/composite-escape.test.ts
@@ -1,83 +1,90 @@
-import vscode from "vscode";
 import { NeovimClient } from "neovim";
+import vscode from "vscode";
 
+import { strict as assert } from "assert";
 import {
-    attachTestNvimClient,
-    sendVSCodeCommand,
     assertContent,
-    wait,
+    attachTestNvimClient,
     closeAllActiveEditors,
     closeNvimClient,
-    sendEscapeKey,
     openTextDocument,
+    sendEscapeKey,
     sendInsertKey,
+    sendVSCodeKeys,
+    wait
 } from "./integrationUtils";
 
 describe("Composite escape key", () => {
     let client: NeovimClient;
     before(async () => {
+        await vscode.workspace.getConfiguration("vscode-neovim").update(
+            "compositeKeys",
+            {
+                jj: {
+                    command: "vscode-neovim.escape",
+                },
+                jk: {
+                    command: "vscode-neovim.lua",
+                    args: ["vim.g.__composite_escape_test = 'jk'"],
+                },
+            },
+            vscode.ConfigurationTarget.Global,
+        );
+        await vscode.commands.executeCommand("vscode-neovim.restart");
+        await wait(300);
         client = await attachTestNvimClient();
     });
     after(async () => {
         await closeNvimClient(client);
         await closeAllActiveEditors();
+        await vscode.workspace
+            .getConfiguration("vscode-neovim")
+            .update("compositeKeys", undefined, vscode.ConfigurationTarget.Global);
+        await vscode.commands.executeCommand("vscode-neovim.restart");
     });
 
-    it.skip("Works", async () => {
+    it("Works", async () => {
         await openTextDocument({ content: "" });
 
         await sendInsertKey();
-        await sendVSCodeCommand("vscode-neovim.compositeEscape1", "j");
-        await wait(500);
-        await sendVSCodeCommand("vscode-neovim.compositeEscape1", "j");
+        await sendVSCodeKeys("j", 500); // Default composite time is 300ms
         await assertContent(
             {
                 mode: "i",
-            },
-            client,
-        );
-        await sendEscapeKey();
-        await assertContent(
-            {
-                content: ["jj"],
-            },
-            client,
-        );
-        await sendInsertKey("A");
-        await sendVSCodeCommand("vscode-neovim.compositeEscape1", "j");
-        await wait(500);
-        await sendVSCodeCommand("vscode-neovim.compositeEscape2", "k");
-        await assertContent(
-            {
-                mode: "i",
-            },
-            client,
-        );
-        await sendEscapeKey();
-        await assertContent(
-            {
-                content: ["jjjk"],
-            },
-            client,
-        );
-        await sendInsertKey();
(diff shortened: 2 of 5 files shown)


src/test/integ/composite-escape.test.ts
```diff
@@ -1,83 +1,91 @@
...
+    sendEscapeKey,
     sendInsertKey,
+    sendVSCodeKeys,
+    wait,
 } from "./integrationUtils";
 
 describe("Composite escape key", () => {
     let client: NeovimClient;
     before(async () => {
+        await vscode.workspace.getConfiguration("vscode-neovim").update(
```
justinmk, line 21: we should use our `Config` abstraction for all interactions with vscode settings. And eventually we should add more safeguards in that abstraction, because vscode settings can be a source of bugs (no type-checking, data can be garbage if users have messed up settings.json).

justinmk: Oh, this is in an integ test so maybe my comment isn't relevant for this code.

src/typing_manager.ts
```diff
@@ -45,11 +34,46 @@ export class TypingManager implements Disposable {
...
+        return this.main.modeManager.isRecordingInInsertMode;
+    }
+
+    private vscodeDefaultType = (text: string) => commands.executeCommand("default:type", { text });
+
     public constructor(private main: MainController) {
+        // Prepare configs for composite keys
+        this.compositeKeys = config.compositeKeys;
+        Object.keys(this.compositeKeys).forEach((key) => {
+            const [first, second] = key.split("");
```
ollien, line 70: Do we otherwise validate that the composite keys are of length 2? It would be nice if we could give a more useful error than just an unpacking crash if someone enters something of length 1 (or a warning if something is greater than 2)

src/typing_manager.ts
```diff
@@ -135,57 +118,109 @@ export class TypingManager implements Disposable {
...
-        } else if (this.isInComposition) {
-            this.composingText += type.text;
-        } else if (this.main.modeManager.isInsertMode && !this.main.modeManager.isRecordingInInsertMode) {
-            if ((await this.client.mode).blocking) {
-                this.client.input(normalizeInputString(type.text, !this.main.modeManager.isRecordingInInsertMode));
+    async compositeInput(key: string) {
+        if (!this.compositeMatchedFirstKey) {
+            if (this.compositeFirstKeys.includes(key)) {
+                this.compositeMatchedFirstKey = key;
+                this.compositePromise?.resolve();
```
ollien, line 149: If we're already concerned about the possibility of `compositePromise` being non-null, should we call `clearTimer` if the composite timer is non-null as well?

xiyaowong: `await null` and `clearTimeout(null)` are all valid

ollien: Ah cool - then it would be a good idea to do `clearTimeout(this.compositeTimer)` here :)

xiyaowong: Here, timer cannot be non-null

src/typing_manager.ts
xiyaowong: I've thought about it too. I'm just too lazy to do it 👀 . Secondly, the configuration schema should guide users to configure the correct configuration.

src/typing_manager.ts
```diff
@@ -135,57 +118,109 @@ export class TypingManager implements Disposable {
...
+            return;
+        }
+
+        await this.compositePromise?.promise;
+        await this.vscodeDefaultType(key);
+    }
+
+    private onVSCodeType = async (_editor: TextEditor, _edit: TextEditorEdit, { text }: { text: string }) => {
+        if (!this.takeOverVSCodeInput) {
+            if (this.isInsertMode) this.compositeInput(text);
```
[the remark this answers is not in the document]
ollien: Just given this tries to process composite inputs whle we're in insert mode, I'm curious what the usecase is for that (unless `jj` is meant to navigate down in insert mode?)

src/typing_manager.ts
xiyaowong: I removed the promise now 😂

package.json
```diff
@@ -385,6 +385,48 @@
...
+                            "jk": {
+                                "command": "vscode-neovim.lua",
+                                "args": [
+                                    "vim.api.nvim_input('<ESC>') require('vscode-neovim').action('workbench.action.files.save')"
+                                ]
+                            }
+                        }
+                    ],
+                    "patternProperties": {
+                        "^[a-zA-Z]{2}$": {
```
xiyaowong, line 414: @ollien VSCode does not warn about configuration errors here, not sure what's wrong.

ollien: I'll take a look :)

xiyaowong: thx!

ollien: Without altering the schema, I don't think we can do this without making the composite key itself a value (e.g. we could have an array of composite key mapping objects). There's `patternErrorMessage` in the [schema](https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema), but I don't think that works for anything but string values (i.e. not keys).

xiyaowong: thx! I actually forgot the vscode docs...

ollien:
FWIW, I've been using this most of today and haven't noticed any(?) perceptible delays on input

ollien (approved):

MaheshNat:
I used to have jk mapped to composeEscape 1 and 2 using the following:

  {
    "command": "vscode-neovim.compositeEscape1",
    "key": "j",
    "when": "neovim.mode == insert && editorTextFocus",
    "args": "j"
  },
  {
    "command": "vscode-neovim.compositeEscape2",
    "key": "k",
    "when": "neovim.mode == insert && editorTextFocus",
    "args": "k"
  },

after switching to:

  "vscode-neovim.compositeKeys": {
    "jk": {
      "command": "vscode-neovim.lua",
      "args": [
        "vim.api.nvim_input('<ESC>')\nrequire('vscode-neovim').action('workbench.action.files.save')"
      ]
    }
  }

I am getting a bug where when I have my cursor position set in insert mode (say on line x), then use my mouse to click elsewhere (say on line y), then press jk, my cursor ends up going back to line x instead of staying on line y. This throws me off and did not occur with the old settings.

xiyaowong:
@MaheshNat  Change args to 
`require('vscode-neovim').action('vscode-neovim.escape') require('vscode-neovim').action('workbench.action.files.save')`

theol0403:
Composite escape should only be for insert mode commands.

uduse:
Hey thanks for the feature!

This actually breaks my use case. My escape is ",,", which worked with the previous version but not this one.

Another feedback is that it's not clear in the documentation that we are removing a key config and replacing it with something `settings.json`.

goyalyashpal:
hi! would it allow to use `alt-key` as `escape key` for all the keys used by neovim?

i.e. say, i press `alt-j` in insert mode, then it would become equivalent to pressing `escape` then `j` thus bringing me in normal mode and then navigate down.

this would help me immensly.

* https://github.com/helix-editor/helix/issues/3162

ollien:
@goyalyashpal please open a separate issue for this, as it's pretty unrelated to this PR

echaya:
The new setting will trigger a file save in "interactive python window" which I have no intention to save the results there. Can we use anything other than `'workbench.action.files.save` so it does not trigger the file saving please?

xiyaowong:
```json
{
    "vscode-neovim.compositeKeys": {
        "jj": {
            "command": "vscode-neovim.escape"
        }
    }
}
```
@echaya 
You can use any commands you want. `workbench.action.files.save` is just for example.

KenzKD:
> Hey thanks for the feature!
> 
> This actually breaks my use case. My escape is `,,` (two commas), which worked with the previous version but not this one.
> 
> Another feedback is that it's not clear in the documentation that we are removing a key config and replacing it with something `settings.json`.

I had made this error and was placing the code snippet in the keyboard shortcuts.json file. It would really help if I this specific change was mentioned on the composite Keys section of the Details panel. I had to find out about this by asking on Reddit.

echaya:
> > Hey thanks for the feature!
> > This actually breaks my use case. My escape is `,,` (two commas), which worked with the previous version but not this one.
> > Another feedback is that it's not clear in the documentation that we are removing a key config and replacing it with something `settings.json`.
> 
> I had made this error and was placing the code snippet in the keyboard shortcuts.json file. It would really help if I this specific change was mentioned on the composite Keys section of the Details panel. I had to find out about this by asking on Reddit.

I second this. I also had to come to this post to figure out it is `settings.json` that i need to add the new setting.

xiyaowong:
If there is a problem, please open an issue.

k-chainalysis:
> > > Hey thanks for the feature!
> > > This actually breaks my use case. My escape is `,,` (two commas), which worked with the previous version but not this one.
> > > Another feedback is that it's not clear in the documentation that we are removing a key config and replacing it with something `settings.json`.
> > 
> > 
> > I had made this error and was placing the code snippet in the keyboard shortcuts.json file. It would really help if I this specific change was mentioned on the composite Keys section of the Details panel. I had to find out about this by asking on Reddit.
> 
> I second this. I also had to come to this post to figure out it is `settings.json` that i need to add the new setting.

Same! But now that I figured out `settings.json` (not `keybindings.json`) good to go. Thanks.

PR #1918, opened 2024-04-28
fix: remove extraneous spaces when performing virtual hls

ollien:
I'm making this a draft because I'm not 100% that I'm not messing something up by doing this. Here's what I do know

1) the disparity between `lineChars` and `filledLineChars` can lead to confusing differences when calling `calcTabCells` (I basically only discovered this when trying to do a refactor, which added complexity I wasn't sure I wanted to deal with unless I could define a clear reason for it).
2) It already added some extra spaces in cases I wouldn't expect (see changes to the test `allows overlaying virtual text on an existing line with double-wide characters)
3) Stock neovim does not add these extra spaces 

For the document
```
👋 yo
✅ yo
你 yo
```

I added virtual text to the first column with `lua vim.api.nvim_buf_set_extmark(0, vim.api.nvim_create_namespace("test"), <lineNumber>, 0, {virt_text = {{"h", "DiffAdd"}}, virt_text_pos = "overlay"})` (substituting in the line number, obviously). 

As of today, this renders like this (notice in particular that the space after the hand gets highlighted)

image

After this change, though, you'll notice that this space is not covered, but the hand slightly peeks out (it is slightly wider than two cols)

image

Standard neovim handles this case entirely differently, but I think due to VSCode limitations we would not be able to achieve what it does

image

IMO, neither the first case nor the second one looks "exactly" right, but the second one has the benefit of simplifying some of the code.

 
Another case I'm seeing, which is represented in the tests is that overlaying "hey!" on the first line omits the y, but looks fine on the third. I don't think this is a serious drawback (we already make certain compromises due to emojis).

Before:

image

After
 
image

Stock nvim:
image

@xiyaowong If you get a moment, I'd love some input on this. Perhaps there's a case I'm not seeing that should be added to the tests


xiyaowong:
It's intentional.
image

Emojis have multiple Unicode ranges, making complete compatibility difficult. (Even Neovim and VSCode don't have full emoji support)

ollien:
Alrighty, will close this then. Thanks for taking a look :)

xiyaowong:
@ollien However, it's worth optimizing. But I hope to prioritize support for language characters.

ollien:
@xiyaowong makes sense. I'm playing a fragile game of trying to figure out what's intentional and what's a bug :)

PR #1920, opened 2024-04-29
docs: Add remote development section

### LINKED ISSUE #262: Allow installation on Github Codespaces

xiyaowong:
- close #262
- close #1000

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ mode and editor commands, making the best use of both editors.
     -   [Jumplist](#jumplist)
     -   [Wildmenu completion](#wildmenu-completion)
     -   [Multiple cursors](#multiple-cursors)
+    -   [Remote development](#remote-development)
 -   [⚡️ API](#%EF%B8%8F-api)
 -   [⌨️ Keybindings (shortcuts)](#%EF%B8%8F-keybindings-shortcuts)
     -   [Add keybindings](#add-keybindings)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -243,6 +244,28 @@ See gif in action:
 The built-in multi-cursor support may not meet your needs. Please refer to the plugin
 [vscode-multi-cursor.nvim](https://github.com/vscode-neovim/vscode-multi-cursor.nvim) for more multi-cursor features
 
+### Remote development
+
+We intend to use vscode-neovim as a UI extension, running locally instead of installing and configuring nvim in a remote
```
justinmk, line 249: can we instead declare `["ui", "workspace"]` and try to do the right thing (and possibly show a message) automatically?

README.md
```diff
@@ -243,6 +244,28 @@ See gif in action:
...
+environment. If you prefer to use nvim in a remote environment, you can set the following in your VSCode
```
ollien, line 250: It might be worth noting this isn't a requirement. I've never done it on any server I've connected to remotely

xiyaowong: Are you sure that the extension is running on remote extension host?

ollien: I'm not, no, but this paragraph makes it sound like these settings are a requirement for using the extension when remotely connecting

README.md
xiyaowong: There should be no way to determine where the user wants to install the extension.

README.md
xiyaowong: Adjusted a little

README.md
```diff
@@ -243,6 +244,30 @@ See gif in action:
 The built-in multi-cursor support may not meet your needs. Please refer to the plugin
 [vscode-multi-cursor.nvim](https://github.com/vscode-neovim/vscode-multi-cursor.nvim) for more multi-cursor features
 
+### Remote development
+
+We intend to use vscode-neovim as a UI extension, so when you're using remote development, vscode-neovim is enabled in
+the Local Extension Host, and it should work out of the box.
+
+If you prefer to use Neovim in the remote environment, vscode-neovim should be installed in the Remote Extension Host.
```
ollien, line 252:
```suggestion
If you prefer to use the remote environment's copy of Neovim, rather than the locally installed one, vscode-neovim should be installed in the Remote Extension Host.
```

Better, just a small tweak you can take or leave

ollien (approved):

Issue #1922, opened 2024-04-30
bug: Composite keys don't work with non-alphabetical keys


uduse:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

NVIM v0.9.5

### Operating system/version

macOS 14.2.1 (23C71)

### Describe the bug

The recent update on composite keys breaks my use case. My escape is `,,` (two commas), which worked with the previous version but not this one.

### Steps To Reproduce

```
  "vscode-neovim.compositeKeys": {
    ",,": {
      "command": "vscode-neovim.lua",
      "args": [
        "vim.api.nvim_input('<ESC>')\nrequire('vscode-neovim').action('workbench.action.files.save')"
      ]
    }
  }
```

### Expected Behavior

The regex here should be the problem, it should capture any arbitrary key code rather than alphabetical ones.

https://github.com/xiyaowong/vscode-neovim/blob/b4053715ac78e8181d9523b3c14d70165787cf25/src/typing_manager.ts#L108-L112

### Proposed Solution
https://github.com/vscode-neovim/vscode-neovim/pull/1923

PR #1923, opened 2024-04-30
fix: Match any ASCII characters in composite keys

uduse:
Ref Issue: #1922

Ref: https://stackoverflow.com/questions/3203190/regex-any-ascii-character

### DIFF
--- a/src/typing_manager.ts
+++ b/src/typing_manager.ts
@@ -105,7 +105,7 @@ export class TypingManager implements Disposable {
         this.compositeFirstKeys = [];
         this.compositeSecondKeysForFirstKey = new Map();
         Object.keys(this.compositeKeys).forEach((key) => {
-            if (!/^[a-zA-Z]{2}$/.test(key)) {
+            if (!/^[ -~]{2}$/.test(key)) {
                 window.showErrorMessage(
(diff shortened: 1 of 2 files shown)


src/typing_manager.ts
```diff
@@ -105,7 +105,7 @@ export class TypingManager implements Disposable {
         this.compositeFirstKeys = [];
         this.compositeSecondKeysForFirstKey = new Map();
         Object.keys(this.compositeKeys).forEach((key) => {
-            if (!/^[a-zA-Z]{2}$/.test(key)) {
+            if (!/^[ -~]{2}$/.test(key)) {
                 window.showErrorMessage(
                     `Invalid composite key: ${key}. Composite key must be exactly 2 characters long.`,
```
xiyaowong, line 110:
```suggestion
                    `Invalid composite key: ${key}. Composite key must be exactly 2 ASCII characters long.`,
```

xiyaowong (changes requested):

xiyaowong:
Should also change config schema in package.json?

uduse:
@xiyaowong I don't understand what else need to be changed. Can you help me here?

xiyaowong (approved):

PR #1930, opened 2024-05-02
refactor: update vscode-neovim.lua command to accept multiple lua codes

xiyaowong:
vscode-neovim.lua is typically used in vscode. It's cumbersome to use with only a single string parameter in JSON format.

Additionally, command may be invoked by other extensions with varying parameter formats, so a more flexible approach was adopted.

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -112,13 +112,14 @@ export class MainController implements vscode.Disposable {
 
         this.disposables.push(
             vscode.commands.registerCommand("_getNeovimClient", () => this.client),
-            vscode.commands.registerCommand("vscode-neovim.lua", async (lua) => {
-                if (!lua) {
+            vscode.commands.registerCommand("vscode-neovim.lua", async (...codes: (string | string[])[]) => {
+                const luaCode = codes.flat().join("\n");
+                if (!luaCode.length) {
                     window.showWarningMessage("No lua code provided");
                     return;
                 }
                 try {
-                    await this.client.lua(lua);
+                    await this.client.lua(luaCode);
                 } catch (e) {
                     logger.error(e instanceof Error ? e.message : e);
                 }


src/main_controller.ts
```diff
@@ -112,13 +112,14 @@ export class MainController implements vscode.Disposable {
 
         this.disposables.push(
             vscode.commands.registerCommand("_getNeovimClient", () => this.client),
-            vscode.commands.registerCommand("vscode-neovim.lua", async (lua) => {
-                if (!lua) {
+            vscode.commands.registerCommand("vscode-neovim.lua", async (...codes: (string | string[])[]) => {
+                const luaCode = codes.flat().join("\n");
```
justinmk, line 116: it's probably less surprising if each chunk is run separately. joining them could result in unintentionally shadowed variables, etc.

xiyaowong: Or use this one:
```typescript
            vscode.commands.registerCommand("vscode-neovim.lua", async (code: string | string[]) => {
                const luaCode = typeof code === "string" ? code : code.join("\n");
```

xiyaowong: The first one might be better because commands might have different parameter configurations in different setups. For example, with `arg = ['foo', 'bar']`, the command could be called as `cmd(arg)` or `cmd(...arg)`

justinmk: This is why variadic parameters (`...foo`) are often an anti-feature. They add useless ambiguity for very little gain. They should rarely be used, except when there is a clear purpose.

xiyaowong: I'm going to merge it. This command is for users to use in vscode. There may be some other extensions that need to call the command, such as multi-commands.

justinmk: > I'm going to merge it. This command is for users to use in vscode. There may be some other extensions that need to call the command, such as multi-commands.

why does that imply the need for a variadic parameter? how would the usage look like with and without a variadic parameter?

xiyaowong: Just make this command a little more flexible. We don't know how users will call it. Maybe changing it to `(...code: string[])` can be compatible with most use cases

justinmk: why is that worth the ambiguity that we identified above? 

> We don't know how users will call it.

we define how it should be called.

Issue #1921, opened 2024-04-30
[REGRESSION BUG] IME text twitches in v1.11.0


wenfangdu:
This is exactly the reason that I don't use https://github.com/VSCodeVim/Vim, now vscode-neovim also has it...........

OS: macOS 14
VSCode: v1.88.1
IME: https://shurufa.sogou.com/

v1.11.0

https://github.com/vscode-neovim/vscode-neovim/assets/28700378/a727e408-3027-40d6-9117-93ba58ba9f8a

v1.10.1

https://github.com/vscode-neovim/vscode-neovim/assets/28700378/2fab8ce7-1abd-437a-ae65-b517eee2b143

Might be caused by https://github.com/vscode-neovim/vscode-neovim/pull/1917.
@xiyaowong Bro, take a look.

xiyaowong:
能不能别老用这种命令的语气？这个按理说是vscode的问题，vscode提供default:type命令就是为了vim这种插件的

xiyaowong:
我明天抽空修复下

wenfangdu:
@xiyaowong 兄弟，有空看看就行，我锁版本了，抱歉语气不好，俺没在意………………

theol0403:
@xiyaowong I don't think we should have merged such a major change so soon. Registering the type event during insert mode is something the extension has explicitly avoided for years. 

Currently, the new approach will break Russian layouts/dead keys/IME for regular typing in insert mode, something we can't do.

https://github.com/vscode-neovim/vscode-neovim/issues/653
https://github.com/vscode-neovim/vscode-neovim/issues/1466

xiyaowong:
@theol0403 These should not be the same problem. First of all, this issue is actually a problem with vscode. Now vscode-neovim uses default:type correctly. I will optimize this problem when I am free today.

xiyaowong:
According to this issue phenomenon, it's not lag, it's just that the characters are jumping.

xiyaowong:
> Currently, the new approach will break Russian layouts/dead keys/IME for regular typing in insert mode, something we can't do.

In ime composition, we didn't do anything superfluous.

theol0403:
I might be wrong here actually. I forgot that we are now using type instead of forwarding using vscode key bindings.

vlwkaos:
Can confirm this with Korean as well, it not only visually twitches, but actually hinders typing.

xiyaowong:
[...]

PR #1931, opened 2024-05-02
refactor: register type commands only when needed

xiyaowong:
At present, the problem in #1921 seems to be unavoidable. The current approach is that when composite keys are not used, the logic is the same as before. See if users can accept it  @theol0403 @wenfangdu @vlwkaos

### DIFF
--- a/src/typing_manager.ts
+++ b/src/typing_manager.ts
@@ -38,7 +38,7 @@ export class TypingManager implements Disposable {
      * Flag indicating that we should take over vscode input
      * If false, we should forward all input received from "type" to "default:type"
      */
-    private takeOverVSCodeInput = false;
+    private _takeOverVSCodeInput = false;
 
     // configs
     private compositeKeys!: CompositeKeys;
@@ -47,6 +47,30 @@ export class TypingManager implements Disposable {
     // logic variables
     private compositeMatchedFirstKey?: string;
     private compositeTimer?: NodeJS.Timeout;
+
+    // Forwarding the arguments of replacePreviousChar to default:replacePreviousChar
+    // causes text jitter during ime composition, so dispose this command when
+    // the vscode input is not taken over
+    private replacePreviousCharHandler?: Disposable;
+
+    private get takeOverVSCodeInput() {
+        return this._takeOverVSCodeInput;
+    }
+
+    private set takeOverVSCodeInput(value: boolean) {
+        this._takeOverVSCodeInput = value;
+        if (!value) {
+            this.replacePreviousCharHandler?.dispose();
+            this.replacePreviousCharHandler = undefined;
+            return;
+        }
+        if (!this.replacePreviousCharHandler) {
+            this.replacePreviousCharHandler = commands.registerCommand(
+                "replacePreviousChar",
+                this.onReplacePreviousChar,
+            );
+        }
+    }
 
     private get client() {
         return this.main.client;
@@ -105,10 +129,7 @@ export class TypingManager implements Disposable {
         };
 
         this.takeOverVSCodeInput = true;
-        this.disposables.push(
-            commands.registerTextEditorCommand("type", this.onVSCodeType),
-            commands.registerCommand("replacePreviousChar", this.onReplacePreviousChar),
-        );
+        this.disposables.push(commands.registerTextEditorCommand("type", this.onVSCodeType));
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const registerCommand = (cmd: string, cb: (...args: any[]) => any) => {
             this.disposables.push(commands.registerCommand(cmd, cb, this));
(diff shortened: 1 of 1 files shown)


wenfangdu:
@xiyaowong Bro, leave a `.vsix` file, so that we can try.

xiyaowong:
@wenfangdu https://github.com/vscode-neovim/vscode-neovim/actions/runs/8927691229

wenfangdu:
@xiyaowong You mean get `.vsix` from that page?

xiyaowong:
@wenfangdu 
image

wenfangdu:
@xiyaowong Seems normal to me:

https://github.com/vscode-neovim/vscode-neovim/assets/28700378/0c1d78e2-e6f9-4cd7-9055-cdfc332f9a8e

theol0403:
Code looks fine. Merge when tested.

wenfangdu:
@xiyaowong Seems like https://github.com/vscode-neovim/vscode-neovim/pull/1923 already fixed https://github.com/vscode-neovim/vscode-neovim/issues/1921, not reproducible in v1.11.1 🤔.

src/typing_manager.ts
```diff
@@ -197,7 +237,7 @@ export class TypingManager implements Disposable {
             const matchedFirstKey = this.compositeMatchedFirstKey;
             this.compositeMatchedFirstKey = undefined;
             const { command, args } = this.compositeKeys[matchedFirstKey + key];
-            await commands.executeCommand(command, ...(args ? args : []));
+            commands.executeCommand(command, ...(args ? args : []));
```
justinmk, line 240: why remove `await`? same question for line 325

xiyaowong: As seen in the discussion of vscode related issues, make the execution as lightweight as possible. Whether or not await is added here has no effect on vscode-neovim.

src/typing_manager.ts
```diff
@@ -35,19 +35,62 @@ export class TypingManager implements Disposable {
...
+    }
+
+    private set takeOverVSCodeInput(takeOver: boolean) {
+        this._takeOverVSCodeInput = takeOver;
+
+        if (takeOver) {
+            if (!this.typeHandler) this.typeHandler = commands.registerTextEditorCommand("type", this.onVSCodeType);
+            if (!this.replacePreviousCharHandler)
+                this.replacePreviousCharHandler = commands.registerCommand(
+                    "replacePreviousChar",
```
justinmk, line 80: why *register* the command every time `takeOverVSCodeInput=true` is assigned? I would think that the command(s) can be registered once, and there implementation can check state to decide what to do.

xiyaowong: >     // Notes:
>     // 1. "type" and "replacePreviousChar" must be registered at the same time
>     // 2. Forwarding the arguments of replacePreviousChar to default:replacePreviousChar
>     //    causes text jitter during ime composition
>     //
>     // The compromise solution used here:
>     // 1. When composite keys are needed, always register type and
>     //    replacePreviousChar at the same time, which unavoidably causes text
>     //    flickering during ime composition.
>     // 2. When composite keys are not needed, only register type and
>     //    replacePreviousChar when it's necessary to take over vscode input.

justinmk: oh, I see that this is overwriting a vscode builtin: https://github.com/microsoft/vscode/blob/3982027f02dee5aa52622f7f2812c70f04ce337a/src/vs/editor/browser/coreCommands.ts#L2167

it sure would be worth mentioning that in a code comment somewhere.

justinmk: I did read that comment, but it doesn't mention that this is a vscode builtin, which is critical for understanding this.

xiyaowong: We can mention [vim-sample](https://github.com/microsoft/vscode-extension-samples/tree/main/vim-sample) in CONTRIBUTING.md

justinmk: sure, but that still leaves other developers with a puzzle to piece together. It costs nearly zero to mention in this context that "replacePreviousChar" is a vscode builtin.

xiyaowong: Oh, the relevant content has been explained here: https://github.com/vscode-neovim/vscode-neovim/blob/master/CONTRIBUTING.md#typingmanager

justinmk: Again, we are talking about a 1-line addition that is *near* this code.

xiyaowong: What's the relation between this and this PR? More than just these two vscode commands are overridden.

justinmk: This is something I noticed while reviewing this PR. It's a zero-cost improvement that can be made, regardless of other possible, future improvements. Is any other justification needed?

Issue #1888, opened 2024-04-14
bug: hlsearch is persistent in places text no longer exists, or never was


ollien:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

v0.9.5

### Operating system/version

Fedora 39

### Describe the bug

When I use `hlsearch`, sometimes the "highlights" stick around in places the text no longer exists, similar to https://github.com/vscode-neovim/vscode-neovim/issues/1555. I'm not sure when it originated, it's been a problem for as long as I've used the plugin, but I've never quite pinned it down until now (it felt "random")

I've attached a video of a repro, but it's definitely not consistent in how it fails. In this case, you'll notice that once I hit `J` to join the lines, the `I` from `Insert` is duplicated. You'll also notice that on line 69, an extra `Insert` is overlaid the `await`.

https://github.com/vscode-neovim/vscode-neovim/assets/977151/854d4b88-1cae-46ff-8c3a-c1eddfb39c7f

 

### Steps To Reproduce

1. Search for some text
2. Edit text around the vicinity of the highlight, such that the highlighted portion moves,
3. After some number of tries, the highlighted portion is placed in a strange location

### Expected Behavior

The highlight should move with the remaining text, if the text is still a match, just like in the commandline nvim

ollien:
FWIW, I am interested in working to solve this bug, so if you have any pointers as to where I could look, it would be appreciated! Otherwise, I'm going to do some digging of my own.

justinmk:
https://github.com/vscode-neovim/vscode-neovim/issues/1555 looks related. It was closed by https://github.com/vscode-neovim/vscode-neovim/pull/1575 , so looking at that code would be a place to tart.

ollien:
I've been staring at this for a little while, and I'm having trouble capturing the precise nature of what's going on, but I do think I have some kind of idea, and I wanted to share my findings.

In order to calculate offsets, the highlight manager requests details about the current viewport, which implies that these `grid_line` events are dependent on other viewport events.  https://github.com/vscode-neovim/vscode-neovim/blob/f7f89382571dedb7e84aaadf0ccf917371066ce2/src/highlight_manager.ts#L51

When a batch of redraws is received, the extension broadcasts all the redraw events to the various `Manager` listeners
[...]

Issue #372, opened 2020-09-20
Easymotion incorrect jump to search


theol0403:
When using https://github.com/asvetliakov/vim-easymotion, and doing `<leader>/` for search, when I type the search then choose the target, it jumps to the first non-whitespace col in the line below the target instead of directly to the target.

![Peek 2020-09-20 11-04](https://user-images.githubusercontent.com/16546293/93718452-2d49c500-fb31-11ea-9d0f-43ce18da9ac5.gif)

asvetliakov:
The problem here is the automatically accepting return prompt caused by easymotion. But for easymotion search it's "autoaccepted" already

MarWeUMR:
Hi,
is it possible, that this bug is still persisting? I have the exact same problem that is showcased above.

DominickVale:
I'm having the same issue as well.

PR #1940, opened 2024-05-05
fix: redraw events are processed in order

ollien:
**REVIEW NOTE:** This is almost certainly going to be easier to review if you ignore whitespace when viewing the diff.

This is a bit of an untested prototype. I'm not 100% what it fixes or breaks, but it comes back to a comment I left [on another issue](https://github.com/vscode-neovim/vscode-neovim/issues/1888#issuecomment-2054223810). The TL;DR is that neovim actually assumes we are processing `redraw` events in order, but we don't guarantee that, due to the way the `eventBus` is set up (see the linked comment for more details). My suspicion is that this will fix a couple of those "random" bugs we see here and there, but I can't prove that.

There's a couple of things that seem suspect to me immediately.

1. Cursor events will fire `processCursorMoved` after each item in the redraw batch, rather than after all the events have completed. This may or may not be a problem, but it may be prudent to build some way to flush those updates to the UI only when the batch finishes.
2. The statusline manager checks if mouse events exist within the batch, but I've commented that check out for now. It shouldn't be hard to build some kind of proxy for it, though.

Thoughts are welcome!


ollien:
Looking back, it seems that (2) in my above list was a hack to fix this issue https://github.com/vscode-neovim/vscode-neovim/issues/372. The fork of that extension doesn't work anymore, and I couldn't produce the case with the original extension (more accurately, I couldn't get the binding for `<leader>/` to actually produce jump-points), so I'm inclined to just fix this by removing that hack and avoid adding complexity to solve it

xiyaowong:
I tried this before, but I don't quite remember why I didn't continue. It could be due to performance issues, or sometimes certain modules would conflict.

justinmk (approved):
I don't see how this could cause any performance impact. And it's more correct. Using `map()` to process a redraw batch in arbitrary order is completely wrong.

xiyaowong:
> I don't see how this could cause any performance impact.

More "fire"

> Using `map()` to process a redraw batch in arbitrary order is completely wrong.

The issue here is not with the map itself; it merely encapsulates the data. It's entirely possible to iterate through the data after the mapping process.

justinmk:
> More "fire"

ok, then should the handler for `eventBus.fire("redraw", ...)` be (also) changed?

xiyaowong (approved):

ollien:
Thanks for the look guys. Nice to know I'm not completely off-base. I haven't gotten through what I need to in order to fix my cursor manager concerns (and the highlight manager too, probably, for similar reasons). My thought is I can emit some kind of event once all the "redraw" handlers finish (a "redraw-done" event of some kind), but I'm still exploring the best way to do that.

Just to circle back to some of the above conversation:

> More "fire"

I don't think "fire" per-se is a problem. Under the covers, "fire" is mostly just a for loop to broadcast to each listener (there are some asterisks on that), so we're effectively just "moving" the redraw handling.

> Some modules need data initialization for redraw handling. Currently, each redraw requires initialization every time. 

Can you think of any examples that concern you? A quick look doesn't show very much work going on per event. 

> Additionally, this approach can't ensure 100% sequential event execution. We lack a lock. Sequential execution becomes synchronous, potentially causing lag, especially with the highlight manager due to VSCode's processing speed.

It's true, we can't have 100% sequential event execution with this approach, since, as you point out, there are some modules that require asynchronous execution. This is definitely going to be a lot closer though, given JS is effectively single-threaded, and no other call will run until the prior returns.

xiyaowong:
I cannot provide specific examples either. When I previously attempted to distribute events one by one, some strange issues did arise. Perhaps with ongoing fixes, this potential factor may have disappeared. In any case, I just brought it up to indicate which aspects need attention.

ollien:
Highlights were a fun problem to solve, given the use of `async` code in there. I stole a concept from Go, [`WaitGroup]`(https://pkg.go.dev/sync#WaitGroup)s, to ensure that flush events wait for the redraws to complete :)

src/cursor_manager.ts
```diff
@@ -63,9 +64,9 @@ export class CursorManager implements Disposable {
      */
     public wantInsertCursorUpdate = true;
     /**
-     * Set of grid that needs to undergo cursor update
+     * Set of grids that needs to undergo cursor update
      */
-    private gridCursorUpdates: Set<number> = new Set();
+    private gridCursorUpdates: PendingUpdates<number> = new PendingUpdates();
```
justinmk, line 69: Nice re-use. This makes me notice that PendingUpdates has set-like behavior, does that mean the `addForceUpdate` method has a misleading name? Because it's not appending something, but rather *setting* a key, which may overwrite the previous entry.

src/utils/async.ts
```diff
@@ -30,6 +30,41 @@ export class ManualPromise {
     }
 }
 
+export class WaitGroup {
```
justinmk, line 33: needs a docstring that calls out the purpose and behavior. seems a bit weird that `add()` takes no arguments

src/cursor_manager.ts
ollien: Aha! So, yes and no. It has set like behavior in that for each key you add, it only stages that resource as a single entry (check out [this test](https://github.com/vscode-neovim/vscode-neovim/blob/master/src/test/unit/pending_updates.test.ts#L11-L17). If we were to be using conditional updates, this would mean that multiple calls to `addConditionalUpdate` would have one key, with multiple conditions. So no, nothing is overwritten.

As for the name, yeah, I felt weird writing `addForceUpdate` here, since there's no contrasting "conditional" usage. I welcome an alternate name though!

src/utils/async.ts
```diff
@@ -30,6 +30,57 @@ export class ManualPromise {
     }
 }
 
+/**
+ * WaitGroup tracks the progress of multiple asynchronous tasks, allowing a caller to wait for all tasks to complete.
```
justinmk, line 34: When should this be used instead of Promise.all() / Promise.allSettled() ? Docstring should probably answer that.

ollien: Take a look now :)

src/test/unit/utils/async.test.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+describe("WaitGroup", () => {
+    it("should resolve immediately on a new object", async () => {
+        const wg = new WaitGroup();
+
+        await withTimeout(wg.promise, 100);
+    });
+
+    it("should wait forever if there are an imbalanced number of add calls", async () => {
```
justinmk, line 22: Semaphores have the same problem (one of many). Seems like this is reinventing semaphores, but nowhere does it mention that.

src/utils/async.ts
```diff
@@ -30,6 +30,61 @@ export class ManualPromise {
...
+ *
+ * Before each task is spawned, a consumer will call `add` for each spawned task, to indicate that there is a new task
+ * to wait for. Each task independently marks itself as done with the `done` method. Once the number of `done` calls
+ * equals the number of `add` calls, the promise resolves, and a consumer will know the tasks are complete.
+ *
+ * This is useful in contrast to `Promise.all`/`Promise.allSettled`, where it may not be easily determinable up front
+ * how many tasks you are waiting on (e.g. if you wish all event handler functions for event A to fully complete
+ * before an event handler for event B runs).
```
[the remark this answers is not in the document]
ollien: It's a bit easier to understand in context. When the `HighlightManager` gets a redraw event, we add to the WaitGroup before deferring on an async task. Then, if while waiting for that async task, we get `redraw-flush`, we want to make sure that `redraw` event has completed (i.e. said async task has finished) before continuing. 

Your example doesn't fit this bill because if we add to `listA` once `Promise.all` has been called, adding items to listA will not prolong the resolution of `Promise.all` (at least, as far as I know). This structure allows for that

src/test/unit/utils/async.test.ts
justinmk: https://www.npmjs.com/package/@jpwilliams/waitgroup has no dependencies and has some more docs + references to Golang analogs, so +1

src/status_line_manager.ts
```diff
@@ -68,17 +68,36 @@ export class StatusLineManager implements Disposable {
...
 
         // if there is mouse_on event after return prompt, then we don't need automatically accept it
         // use case: easymotion search with jumping
-        let hasMouseOnAfterReturnPrompt = false;
-        data.forEach(({ name, args, lastArg, firstArg }, idx) => {
-            switch (name) {
-                case "msg_showcmd": {
-                    const [content] = firstArg;
-                    let str = "";
+        const hasMouseOnAfterReturnPrompt = false;
```
theol0403, line 76: You should be able to delete this line entirely no?

theol0403:
Great PR! Can you update CONTRIBUTING.md with an explanation of this PR? Alternatively, write a doc block in the code explaining how the processing logic/async timing works.

ollien:
@theol0403 Absolutely! I'll prefer to do it in line, if that's ok with you. Easy for lots of things to get missed in a stray markdown document.

justinmk (approved):

theol0403:
Bug report: the cursor style is incorrect on extension startup

ollien:
Oh interesting. I think this may have uncovered a subtle bug as a result of event ordering. It seems like `mode_info_set` is sent after `mode_change`, so when processing the `mode_change`, we don't have information about how to render the cursor. A quick fix is to simply call `updateCursorStyle` every time we get a flush event, but that's probably not the most robust. Will have to think on this a bit

ollien:
Nope, I just made a silly mistake. I shouldn't be going over the list of events backwards :sweat_smile:. The `splice` in `main_controller` tripped me up

src/main_controller.ts
```diff
@@ -313,22 +313,38 @@ export class MainController implements vscode.Disposable {
...
+                    batch.forEach((batchItem) => {
+                        const eventData = {
+                            name: batchItem[0],
+                            args: batchItem.slice(1),
+                            get firstArg() {
+                                return this.args[0];
+                            },
+                            get lastArg() {
+                                return this.args[this.args.length - 1];
+                            },
```
xiyaowong, line 339: Perhaps delete these two getters. With one less array level, these two seem a bit redundant.

xiyaowong (approved):

src/highlight_manager.ts
```diff
@@ -39,39 +52,46 @@ export class HighlightManager implements Disposable {
...
+    }
 
-        if (pendingUpdates.size() > 0) {
-            this.applyHLGridUpdates(pendingUpdates);
+    private async handleRedrawFlush() {
+        // Wait for any redraw events that have been received to finish
+        // their work, so that we can flush them only after their changes are staged.
+        await this.redrawWaitGroup.wait();
+
+        if (this.pendingGridUpdates.size() === 0) {
```
xiyaowong, line 79:
```suggestion
        if (this.pendingGridUpdates.empty()) {
```

src/cursor_manager.ts
```diff
@@ -177,7 +181,11 @@ export class CursorManager implements Disposable {
      * Called when cursor update received. Waits for document changes to complete and then updates cursor position in editor.
      */
     private processCursorMoved(): void {
-        for (const gridId of this.gridCursorUpdates) {
+        for (const [gridId, shouldUpdate] of this.gridCursorUpdates.entries()) {
+            if (!shouldUpdate) {
```
xiyaowong, line 185:
```suggestion
            if (!shouldUpdate()) {
```

xiyaowong (changes requested):

src/highlight_manager.ts
```diff
@@ -135,8 +155,8 @@ export class HighlightManager implements Disposable {
         }
     }
 
-    private applyHLGridUpdates(pendingUpdates: PendingUpdates<number>): void {
-        for (const [grid, update] of pendingUpdates.entries()) {
+    private applyHLGridUpdates(): void {
+        for (const [grid, update] of this.pendingGridUpdates.entries()) {
```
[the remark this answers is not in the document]
ollien: `update` is a bit clearer in this case, IMO, because there is an actual mutation performed (the `update` function calls `this.highlightProvider.processHLCellsEvent`. That said, if you feel strongly I'll change it

src/cursor_manager.ts
ollien: Maybe I'll open another PR to add some type-checked rules to eslint like https://typescript-eslint.io/rules/no-unnecessary-condition/

xiyaowong (approved):

xiyaowong:
Used for two days, working fine.

Issue #843, opened 2022-03-25
Why there is a ‘呐’ in vscode status bar shows as a vscode neovim item?


RekiDunois:
<img width="331" alt="image" src="https://user-images.githubusercontent.com/23435534/160082413-3f22ecb0-593a-4105-997d-2a105edfd085.png">
when I enter insert mode, and then back to normal mode, it just gone.

what does '呐' means?

theol0403:
Can't reproduce. Might be something in your init.vim

PR #1941, opened 2024-05-05
fix: set an explicit id for status bar item

### LINKED ISSUE #1572: bug: setting affinity clashes with clangd outputs in status bar

### Neovim version (nvim -v)

NVIM v0.9.4

### Operating system/version

Windows 11

### Describe the bug

i have two extensions installed only:
1. Your
2. `llvm-vs-code-extensions.vscode-clangd`

So if i set affinity of your extension as 1, then in status bar, both clangd and vscode neovim extension fight for showing current status in status bar. for example clangd tries to show this

image

and vscode neovim wants to show mode status (insert/visual etc)

they both show their status alternatively each getting only 10ms time (i suppose, but its very fast)
also vscode neovim option is not visible in status bar options when above setting is set.

image

if i disable clangd extension option, then vscode neovim mode status is also not shown

but when i set clangd extension affinity also to 1 alongside your extension's, everythings back to normal. i can see both clangd and vscode neovim options in status bar.

### Steps To Reproduce

1. install your extension and clangd extension.
2. set affinity of ur extension to 1 and restart vscode.
3. open a cpp file 
4. type anything in it for example `#inc`
5. in below status bar you will see both clangd and neovim wanting to show their status.

### Expected Behavior

i felt the problem is in your extension and not clangd is because clangd is still shown in status bar and yours is not. 

So please fix it.
Its literally the best extension on neovim in vscode i have used.

ManuelHentschel:
Fix #1572 (probably related: #1251 #843)

A similar issue can be observed in combination with other extensions

The issue https://github.com/microsoft/vscode/issues/201795 causes status bar items to clash if they are created without explicit ids and in different extension hosts (e.g. by setting extension affinity).

While the underlying issue is not fixed, it can be avoided by assigning an explicit id (and there are no downsides to doing this, as far as I am aware).

### DIFF
--- a/src/status_line_manager.ts
+++ b/src/status_line_manager.ts
@@ -28,7 +28,7 @@ export class StatusLineManager implements Disposable {
     }
 
     public constructor(private main: MainController) {
-        this.statusBar = window.createStatusBarItem(StatusBarAlignment.Left, -10);
+        this.statusBar = window.createStatusBarItem("vscode-neovim-status", StatusBarAlignment.Left, -10);
         this.statusBar.show();
         this.disposables.push(this.statusBar, eventBus.on("redraw", this.handleRedraw, this));
 


theol0403 (approved):

xiyaowong:
No wonder sometimes the status bar disappears for a while or moves to the right.

xiyaowong (approved):

Issue #288, opened 2020-06-24
bug: arrows don't work for cmdline history navigation 


kpeters-cbsi:
I use regexen extensively in vim / neovim. Often I will mess up a regex somehow and need to correct it, and I do that by editing it from my search history (i.e. in vim, hitting "/" then pressing up arrow until I reach the regex I want to edit). How do I do this in vscode-vim? I've been having to spawn vim in a terminal, do the regex stuff, and then exit.

chrisjaquet:
I have the same question and would extend it to command history too.

theol0403:
Yeah, for some reason the arrows don't work in command mode anymore. I tried installing some very old versions of this plugin and still could not get it to work, so I assume an update to VSCode broke it, not this plugin. Trying to manually map the arrow keys don't work either, however, putting this in your keybindings is a mediocre fix:
```
    {
        "command": "vscode-neovim.history-up-cmdline",
        "key": "ctrl-p",
        "when": "neovim.mode == cmdline_normal"
    },
    {
        "command": "vscode-neovim.history-down-cmdline",
        "key": "ctrl-n",
        "when": "neovim.mode == cmdline_normal"
    },
```

chasingmaxwell:
I remember this working for me and I don't think it was too long ago. It no longer does, but that workaround did the trick for now. Thanks @theol0403!

kpeters-cbsi:
That worked for me too. Thanks, @theol0403!

genso:
It works also for me!. Thanks to @theol0403

Shatur95:
@theol0403, thanks!
 @asvetliakov, maybe made this shortcuts be default? In Neovim `<C-p>` and `<C-n`> works for search too.

bisho:
If you could get arrows to work will be ideal. Muscle memory is hard to change. I have seen other vim integrations that even include a drop-down with the history, so moving up/down is easier.

asvetliakov:
VSCode started to disallow to remap `<up>`/`<down>` in quick pick. Blame vscode devs ! 👿  👿  👿

uladkasach:
@asvetliakov  Any thoughts on moving the command line to the VSCode status bar (the bottom bar in the ide)? 

The `VSCodeVim` extension exposes the vim commandline in that bar - it shows what you're typing and allows you to navigate through history with arrow keys due to it. Could resolve this problem

theol0403:
[...]

PR #1943, opened 2024-05-05
fix: support up/down for cmdline history

theol0403:
Fixes https://github.com/vscode-neovim/vscode-neovim/issues/288

As of https://github.com/microsoft/vscode/pull/211756, arrow keys are now bindable in quickpick.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1008,6 +1008,18 @@
             },
             {
                 "command": "vscode-neovim.send-cmdline",
+                "key": "up",
+                "when": "neovim.init && neovim.mode == cmdline",
+                "args": "<Up>"
+            },
+            {
+                "command": "vscode-neovim.send-cmdline",
+                "key": "down",
+                "when": "neovim.init && neovim.mode == cmdline",
+                "args": "<Down>"
+            },
+            {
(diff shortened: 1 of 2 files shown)


theol0403:
I think more thought should be put towards how we handle navigating the suggestions.

My proposal: arrows navigate history, C-n/C-p navigate suggestions. Perhaps we can be clever and say that C-n/C-p navigate history when there are no suggestions, but that might be more tricky.

ollien:
I think that's a good default, but the surrounding UX could stand to be explored. My from-the-hip suggestion would be that going "down" go back in history if there are no suggestions, or "enter the suggestion list" if there are. If you're in the suggestion list, up goes to the previous item, or the command box if you're already at the top.

Or maybe we can 80/20 this and just put in the title of the tooltip "C-n/C-p to navigate suggestions" or some such.

xiyaowong (approved):

ollien:
Any reason not to merge this at this time?

theol0403:
I haven't had time to look into this but the risk is that it currently removes the ability to select and accept suggestions using the keyboard.

theol0403:
I'm cooking up something better :eyes:

PR #1949, opened 2024-05-09
build: replace pre-commit with npm-managed lint tools

### LINKED ISSUE #1905: build: lint tasks should not depend on pre-commit
# Problem

our "lint" and "format" tasks depend on pre-commit: https://github.com/vscode-neovim/vscode-neovim/blob/5868ad1c950599a8ae121c4947f3c9b03a5cd0e3/package.json#L2089-L2090

this makes no sense. pre-commit was introduced for CI. devs shouldn't have to install pre-commit to lint the codebase.

ollien:
Fixes #1905

Note with this, we do lose some niceties that pre-commit provided to us (such as stashing unstaged changes so the lint tools won't run on them), but I have chosen to follow a "Keep it Simple" approach just to get this transition moving, and will revisit some of the nicer DX features afterward.

Sorry about blowing up the PR with `package-lock` changes. This was necessary to reinstall the lint tools :(


package.json
```diff
@@ -2133,19 +2139,28 @@
         "vscode:prepublish": "webpack --mode production",
         "webpack": "webpack --mode development",
         "webpack-dev": "webpack --mode development --watch",
-        "test-compile": "tsc -p ./"
+        "test-compile": "tsc -p ./",
+        "prepare": "husky"
     },
     "devDependencies": {
+        "@johnnymorganz/stylua-bin": "^0.20.0",
```
ollien, line 2146: I don't _love_ this dependency, because it ultimately just fetches a binary from Github, but I think it's at least an ok compromise to ensure the version is managed.

xiyaowong:
With just this little thing, we keep tweaking it back and forth. It's actually making things more complicated for this project :dizzy_face:

.github/workflows/lint.yml
```diff
@@ -8,13 +8,20 @@ jobs:
...
-            - uses: pre-commit/action@v3.0.1
+
+            - name: Check Lint
+              run: npm run lint:check
+
+            - name: Check Formatting
+              run: npm run format:check
+
+            - name: Check Lua Styling
+              run: npm run stylua:check
```
justinmk, line 27: formatting is just another kind of lint. "check" is redundant. 

actually, why do these need names? the script name is more useful than these other names.

```suggestion
            - run: npm run lint:check
            - run: npm run format:check
            - run: npm run stylua:check
```

ollien: The names are just easier to grok sometimes, but I don't have a strong preference. If it helps people realize what commands they actually are, it's A-OK to me

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+echo "Running pre-commit tasks..." >&2
+npm run lint:check || (echo 'Lint failed. Fix the above errors, and/or run `npm  run lint` to autofix where possible, then re-commit.' >&2 && exit 1)
+npm run format:check || (echo 'Bad code formatting. Run `npm run format`, then re-commit.' >&2 && exit 1)
+npm run test-compile || (echo 'Test compilation failed. Fix the above errors, then recommit.' >&2 && exit 1)
```
justinmk, line 4: verbalism

```suggestion
npm run lint:check
npm run format:check
npm run test-compile
```

CONTRIBUTING.md
```diff
@@ -20,8 +20,6 @@ git clone https://github.com/vscode-neovim/vscode-neovim
 
 ```sh
 npm install
-pipx install pre-commit  # https://pre-commit.com/#installation
-pre-commit install
 ```
```
justinmk, line 23: 👍  this is a signal that we are going in the right direction

.husky/pre-commit
ollien: I put these in because it's a little more friendly to new contributors. The output of `npm run lint` is going to suggest using `--fix`, which won't really be useful, since `--check` and `--fix` can't coexist AFAIK.

Ditto with Prettier. It isn't obvious what command you need to run to fix it.

package.json
justinmk: hmm. but the scripts are still using `npx`, is that intentional?

.husky/pre-commit
justinmk: adding hints is useful. but `'Test compilation failed. Fix the above errors, then recommit.'` isn't helping anyone, it's noise

justinmk (approved):
Thank you for doing The Work!

justinmk:
> some niceties that pre-commit provided to us (such as stashing unstaged changes so the lint tools won't run on them)

I don't understand why any developer would expect a build system to do this for them. I don't want a pre-commit hook fiddling with my workspace.

.husky/pre-commit
ollien: I dunno, I won't push too hard on this, and if you feel strongly I'll cut down on the wording, but I do think it's nice for those unfamiliar with the commands whizzing by to know what stage failed (pre-commit did this with fancy terminal art, but that was a bit much for this). 

Difference of opinion :)

package.json
ollien: Correct. It still runs via node, but it's just a thin wrapper to run the binary. I figure this extra layer of dependency actually helps us ensure everyone's running the same stylua version, which is probably a good thing. I wish it weren't just pulling it from Github, though.

https://github.com/JohnnyMorganz/StyLua/blob/049bb7c2c5ff0f764eec8ab3627e114ff23ddabf/stylua-npm-bin/binary.js#L123-L136

ollien:
> I don't understand why any developer would expect a build system to do this for them. I don't want a pre-commit hook fiddling with my workspace.

It can be useful if you're doing a partial commit, and don't want lint errors on other areas of your workspace to hold up your commit. But again, that's just preference.

.husky/pre-commit
justinmk: what contributors are you envisioning that will be committing code contributions but are unfamiliar with `npm run test-compile` ?

justinmk:
> But again, that's just preference.

Should preferences be built into the build system?

.husky/pre-commit
ollien: None, it's just a lot of commands whizzing by on commit, and can make their lives a bit easier. 

It seems you do feel strongly, so I'll tone it down a bit

ollien:
> Should preferences be built into the build system?

Depends on who's using it! But yeah, if you don't like it I won't put it in. I'm just stating why it _can_ be useful

package.json
justinmk: If we are using `npx`, then I don't think the `devDependencies` entry is (technically) needed, is it?

If we want this entry for "explicitness", that seems fine. But does `npx` actually look at this?

justinmk (approved):

package.json
ollien: Oh, I used `npx` beccause `stylua-bin` didn't run anything, but going back, I realize now that npm handles just `stylua` just fine. :facepalm: 

Will replace

PR #1951, opened 2024-05-10
fix: include time and scope in log file

xiyaowong:


justinmk:
the title is misleading, because this is only for "console" ? 

for `logToOutputChannel`, timestamp is redundant because vscode already shows a timestamp next to each message.

also, `chore` should be banned, it's a useless prefix. almost everything is either a "fix" or a "feature" (or a few other specific types). "chore" is over-used. this PR for example is a "fix".

xiyaowong:
> the title is misleading, because this is only for "console" ?
> 

??? 

> also, `chore` should be banned, it's a useless prefix. almost everything is either a "fix" or a "feature" (or a few other specific types). "chore" is over-used. this PR for example is a "fix".

Personally, I think it's just a small adjustment...

ollien (approved):

ollien:
> also, chore should be banned, it's a useless prefix. almost everything is either a "fix" or a "feature" (or a few other specific types). "chore" is over-used. this PR for example is a "fix".

In principle, I agree. The Angular ones are

- build
- chore
- ci
- docs: 
- style
- refactor
- perf
- test

One could argue this PR is a `feat`, but I don't know if it's something that needs to be explicitly release-noted, either. I usually like `chore` for things that really couldn't fit into anything else, like linter changes/runs (since those aren't really "build", and perhaps aren't always "style")

justinmk (approved):

PR #1955, opened 2024-05-12
build: migrate to eslint flat config format

ollien:
Modern versions of eslint have migrated to the ["flat" configuration style](https://eslint.org/docs/latest/use/configure/migration-guide), deprecating `.eslintrc.js`. I've gone ahead and migrated the project to the new format, which will allow us to safely upgrade to newer versions of eslint. I believe I've preserved all existing rules, but it's a bit hard to prove a negative :sweat_smile: 

The `globals` dependency is needed due to a change in how the new configuration format handles known globals. See: https://eslint.org/docs/latest/use/configure/migration-guide#configuring-language-options


xiyaowong:
Nice. I've been wanting to do this for a long time. 

Also need to turn on "eslint.experimental.useFlatConfig"

https://github.com/vscode-neovim/vscode-neovim/blob/1fc6752429e1326f98557bd7a5961030e935b5f7/.vscode/settings.json#L14

eslint.config.cjs
```diff
@@ -0,0 +1,80 @@
...
+            "@typescript-eslint/explicit-function-return-type": "off",
+            "@typescript-eslint/no-non-null-assertion": "off",
+            "@typescript-eslint/no-explicit-any": "off",
+        },
+    },
+    {
+        files: ["src/test/**/*.test.ts"],
+        languageOptions: {
+            globals: {
+                ...globals.jest,
```
xiyaowong, line 74: We are not using jest

ollien: Good callout. I took ths from the old config without realizing this. Looks like there's some `mocha` globals we can use.

xiyaowong: :joy: Okay, I haven’t looked at eslintrc carefully.

xiyaowong:
I don't understand why it's not "chore" but "build".

ollien:
I can change the title if you want. I know there were discussions of not using `chore` so I picked the next best thing. I have no strong opinions one way or the other

xiyaowong:
😵

justinmk:
"chore" is a useless prefix. Of course this is either "build" or "lint". Why wouldn't it be?

justinmk (approved):

xiyaowong:
Should delete `.eslintignore`

ollien:
> "chore" is a useless prefix. Of course this is either "build" or "lint". Why wouldn't it be?

FYI, `lint` is not a prefix we allow. From the CI:

`allowed-commit-types: feat,fix,docs,style,refactor,test,build,perf,ci,chore,revert,merge,wip`

I agree that would be the most fitting here

justinmk:
> `lint` is not a prefix we allow.

we can add it (and remove "chore")

theol0403 (approved):

xiyaowong:
> we can add it

👍🏻

 (and remove "chore")

👎🏻

Most front-end projects use something like chore(lint)

justinmk:
> Most front-end projects use something like chore(lint)

Why does that matter? 

The problem is that people over-use "chore", it ends up being 30% of the commits.

PR #1961, opened 2024-05-14
build: enable recommended typechecked rules

ollien:
I'm still in my linting rabbit hole for right now... I've gone ahead and enabled the `recommendedTypeChecked` rules in the linter. I want to eventually move towards `strictTypeChecked` (especially rules like [`no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition), which are very helpful for catching bugs), but one step at a time.

You can view the rules which are enabled, here https://typescript-eslint.io/rules/?=recommended-typeInformation

There are some code changes involved here, but I'm fairly confident they won't cause any issues. You can review the commits individually to better understand why each change was made

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -8,7 +8,7 @@ const globals = require("globals");
 
 module.exports = tseslint.config(
     eslint.configs.recommended,
-    ...tseslint.configs.recommended,
+    ...tseslint.configs.recommendedTypeChecked,
     {
         plugins: {
             import: importsPlugin,
@@ -19,6 +19,10 @@ module.exports = tseslint.config(
             globals: {
                 ...globals.node,
                 ...globals.es6,
+            },
+            parserOptions: {
+                project: true,
+                tsconfigRootDir: __dirname,
             },
         },
         rules: {
@@ -65,6 +69,19 @@ module.exports = tseslint.config(
             "@typescript-eslint/explicit-function-return-type": "off",
             "@typescript-eslint/no-non-null-assertion": "off",
             "@typescript-eslint/no-explicit-any": "off",
+
+            "@typescript-eslint/no-misused-promises": "off",
+            "@typescript-eslint/no-unsafe-argument": "off",
+            "@typescript-eslint/no-unsafe-assignment": "off",
+            "@typescript-eslint/no-unsafe-call": "off",
+            "@typescript-eslint/no-unsafe-declaration-merging": "off",
+            "@typescript-eslint/no-unsafe-enum-comparison": "off",
+            "@typescript-eslint/no-unsafe-member-access": "off",
+            "@typescript-eslint/no-unsafe-return": "off",
+            "@typescript-eslint/restrict-template-expressions": "off", // too jumpy
+            "@typescript-eslint/no-floating-promises": "off", // jumpy; would be nice to turn on, but we have a lot of these
+            "@typescript-eslint/unbound-method": "off", // jumpy, given how vscode's API binds this. Would be good to remove.
+            "@typescript-eslint/require-await": "off", // TODO: remove
         },
     },
     {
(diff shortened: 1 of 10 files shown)


eslint.config.cjs
```diff
@@ -65,6 +71,15 @@ module.exports = tseslint.config(
             "@typescript-eslint/explicit-function-return-type": "off",
             "@typescript-eslint/no-non-null-assertion": "off",
             "@typescript-eslint/no-explicit-any": "off",
+            "@typescript-eslint/restrict-template-expressions": "off", // too jumpy
+            "@typescript-eslint/no-misused-promises": "off", // too jumpy
+            "@typescript-eslint/no-floating-promises": "off", // jumpy; would be nice to turn on, but we have a lot of these
```
justinmk, line 76: This one is very useful. We should try to enable it soon.

eslint.config.cjs
```diff
@@ -65,6 +71,15 @@ module.exports = tseslint.config(
...
+
+            // This codebase interacts with `any` a lot, so for now, all no-unsafe-* we violate are currently disabled
```
justinmk, line 78: A good compromise is to allow code to explicitly use `foo: any` to opt-in to "any". I think that's a tsconfig option, IIRC.

justinmk (approved):
Nice!

xiyaowong (dismissed):
Tips: `npx eslint-flat-config-viewer`

src/eventBus.ts
```diff
@@ -172,23 +171,21 @@ class EventBus implements Disposable {
      *
      * @param name - The name of the event.
      * @param handler - The handler function for the event.
-     * @param thisArg - The `this` context used when invoking the handler function.
```
xiyaowong, line 175: Why remove this?

ollien: This was to satisfy `no-unbound-method`. ESLint isn't smart enough to dive into this implementation to realize we are binding `this`. In practice, when setting up a listener, we never bound anything other than the `this` that already existed in lexical scope, so I preferred to use arrow functions, which capture `this`, rather than this binding.

eslint.config.cjs
ollien: I think we do, but maybe not. I'll check.

This basically prevents `any` from escaping your current scope, which I'm not sure that casting will satisfy.

src/eventBus.ts
xiyaowong: In general, this topic is also highly relevant to personal habits. I don't engage in too much discussion. The `thisArg` parameter is designed similarly in many APIs within VS Code as well. Regarding this rule, we can consider it as a suggestion rather than a limitation.

ollien: You're right, it's a style thing. We could follow the way VSCode APIs are set up, but we wouldn't be able to reap the benefits of this rule, which eliminate an entire class of bug. 

Would you be amenable to replacing the arrow functions with `this.foo.bind(this)`? It would have the same effect, and satisfy the linter.

xiyaowong:
IMO, this PR elevates personal style to a restriction for everyone.

ollien:
@xiyaowong these rules are not style, but enforce correctness. 

Are there any you would like to see removed?

xiyaowong:
Please keep the "thisArg". 

The significance of these changes is minor.
image

ollien:
I agree with @justinmk here, but I would much rather have some more guard rails, rather than none of them because we can't agree on one of them. 

Given all this, I'll revert that commit and disable this rule, but I may introduce it in a follow-on PR (as I plan to do with some of the other rules when I get the brain cycles)

xiyaowong (approved):

ollien:
Force pushed to drop 5b49a1454e4258e68a42ce0378681af4d3a094fa and fix a minor conflict in `master`

ollien:
Failure is expected in the tests: https://github.com/vscode-neovim/vscode-neovim/pull/1973#issuecomment-2118024739

PR #1962, opened 2024-05-14
feat!: use VSCode log level instead of our own

ollien:
After my confusion [here](https://github.com/vscode-neovim/vscode-neovim/discussions/1952), I changed the extension logging to use the VSCode setting, rather than our own. This also removes the `logLevel` setting we previously had.

The biggest change here, by far, is the movement from our `LogLevel` to the VSCode `LogLevel` (which unfortunately, reverses the order of the levels! Argh!)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,11 +37,19 @@ npx vsce package -o vscode-neovim.vsix
 
 ### Logging
 
-You can observe the extension logs via the `vscode-neovim` Output channel or from the dev tools console (run the
-`Developer: Toggle Developer Tools` vscode command to see the console).
+You can view the extension logs in one of three locations
 
-Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
-[`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
+1. Via the `vscode-neovim` Output channel
+
+    - **NOTE**: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
+      [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
+
+2. From the dev tools console (run the `Developer: Toggle Developer Tools` vscode command to see the console) by
+   enabling the `vscode-neovim.logOutputToConsole` setting.
+3. From a log file of your choosing, by configuring the `vscode-neovim.logPath` setting.
+
+VSCode, by default will only show messages at the "Info" level or above, but than can be changed by running the command
+"Developer Set Log Level..." -> "vscode-neovim" and selecting the desired log level.
 
 ### Run Unit Tests
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,7 @@ You can view the extension logs in one of three locations
 
 1. Via the `vscode-neovim` Output channel
 
-    - **NOTE**: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
+    - Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
       [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
 
 2. From the dev tools console (run the `Developer: Toggle Developer Tools` vscode command to see the console) by
@@ -49,7 +49,8 @@ You can view the extension logs in one of three locations
 3. From a log file of your choosing, by configuring the `vscode-neovim.logPath` setting.
 
 VSCode, by default will only show messages at the "Info" level or above, but than can be changed by running the command
-"Developer Set Log Level..." -> "vscode-neovim" and selecting the desired log level.
+`Developer Set Log Level...` -> `vscode-neovim` and selecting the desired log level. You can also do this by clicking
+the gear icon in the output pane, with `vscode-neovim` selected.
 
 ### Run Unit Tests
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -31,7 +31,7 @@ export async function activate(context: vscode.ExtensionContext, isRestart = fal
     disposables.push(outputChannel);
 
     config.init();
-    rootLogger.init(vscode.env.logLevel, config.logPath, config.outputToConsole, outputChannel);
+    rootLogger.init(outputChannel.logLevel, config.logPath, config.outputToConsole, outputChannel);
     eventBus.init();
     actions.init();
     context.subscriptions.push(
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -4,7 +4,7 @@ import actions from "./actions";
 import { config } from "./config";
 import { EXT_ID, EXT_NAME } from "./constants";
 import { eventBus } from "./eventBus";
-import { LogLevel, createLogger, logger as rootLogger } from "./logger";
+import { createLogger, logger as rootLogger } from "./logger";
 import { MainController } from "./main_controller";
 import { VSCodeContext, disposeAll } from "./utils";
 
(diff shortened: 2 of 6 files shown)


theol0403 (approved):

src/logger.ts
```diff
@@ -90,39 +120,38 @@ export class Logger implements Disposable {
...
+                case vscode.LogLevel.Warning:
                     this.outputChannel.warn(fullMsg);
                     break;
-                case LogLevel.info:
-                case LogLevel.debug:
-                    // XXX: `vscode.LogOutputChannel` loglevel is currently readonly:
-                    //      https://github.com/microsoft/vscode/issues/170450
-                    //      https://github.com/PowerShell/vscode-powershell/issues/4441
-                    // So debug() drops messages unless the user has increased vscode's log-level.
-                    // Use info() until vscode adds a way to set the loglevel.
```
xiyaowong, line 116: Does this problem no longer exist? When vscode starts, the log level is set to 'info'. It takes a while before `onDidChangeLogLevel` is triggered.

xiyaowong:
I usually have vscode-neovim completely disabled. Currently, there is no good way to achieve this besides customizing the log level. Personally, I am rather averse to using outputChannel to output vscode-neovim logs. If everyone agrees that this modification is acceptable, I would like to add a configuration option to disable outputChannel, for example: 'vscode-neovim.useOutputChannel"

Not sure whether vscode will synchronize the log level of the output across different development (environments.)

src/logger.ts
justinmk: This problem was resolved in recent versions of vscode. User can set the log-level on the channel. Note that they must also choose `Set as default`

image

justinmk:
> I would like to add a configuration option to disable outputChannel, for example: 'vscode-neovim.useOutputChannel"

Can't the `Off` setting address that use-case? (Then choose `Set as default`).
image

CONTRIBUTING.md
```diff
@@ -37,11 +37,19 @@ npx vsce package -o vscode-neovim.vsix
...
-`Developer: Toggle Developer Tools` vscode command to see the console).
+You can view the extension logs in one of three locations
 
-Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
-[`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
+1. Via the `vscode-neovim` Output channel
+
+    - **NOTE**: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
+      [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
+
```
justinmk, line 46: unnecessary blank lines (which spread out the rendered markdown)
```suggestion
    - **NOTE**: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
      [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
```

CONTRIBUTING.md
```diff
@@ -37,11 +37,19 @@ npx vsce package -o vscode-neovim.vsix
...
+2. From the dev tools console (run the `Developer: Toggle Developer Tools` vscode command to see the console) by
+   enabling the `vscode-neovim.logOutputToConsole` setting.
+3. From a log file of your choosing, by configuring the `vscode-neovim.logPath` setting.
+
+VSCode, by default will only show messages at the "Info" level or above, but than can be changed by running the command
+"Developer Set Log Level..." -> "vscode-neovim" and selecting the desired log level.
```
justinmk, line 52: `Developer: Set Log Level...` is a drastic way to set the log level, because it increases the logs for all vscode components.

The log-level can also be set per-channel. We should definitely mention that, and also mention that `Set As Default` should be selected if you want it to persist.

image

```suggestion
`Developer: Set Log Level...` -> "vscode-neovim" and selecting the desired log level.
```

CONTRIBUTING.md
```diff
@@ -37,11 +37,19 @@ npx vsce package -o vscode-neovim.vsix
...
 
-You can observe the extension logs via the `vscode-neovim` Output channel or from the dev tools console (run the
-`Developer: Toggle Developer Tools` vscode command to see the console).
+You can view the extension logs in one of three locations
 
-Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
-[`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
+1. Via the `vscode-neovim` Output channel
+
+    - **NOTE**: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
```
justinmk, line 44: this note is not as important as it thinks it is
```suggestion
    - Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
```

src/extension.ts
```diff
@@ -31,7 +31,7 @@ export async function activate(context: vscode.ExtensionContext, isRestart = fal
     disposables.push(outputChannel);
 
     config.init();
-    rootLogger.init(LogLevel[config.logLevel], config.logPath, config.outputToConsole, outputChannel);
+    rootLogger.init(vscode.env.logLevel, config.logPath, config.outputToConsole, outputChannel);
```
justinmk, line 34: this isn't what we want. we want the channel-local log level (which is persisted by choosing `Set As Default`)

```suggestion
    rootLogger.init(outputChannel.logLevel, config.logPath, config.outputToConsole, outputChannel);
```

src/logger.ts
```diff
@@ -40,42 +32,80 @@ function getTimestamp(): string {
...
+        this.listenForLogLevelChanges();
+        this.setupLogFile();
     }
 
     public dispose(): void {
         disposeAll(this.disposables);
     }
 
-    private log(level: LogLevel, scope: string, logToOutputChannel: boolean, args: any[]): void {
+    private listenForLogLevelChanges() {
```
justinmk, line 70: don't we use the `on` prefix for listeners?
```suggestion
    private onLogLevelChanged() {
```

src/logger.ts
```diff
@@ -40,42 +32,80 @@ function getTimestamp(): string {
...
+                dispose: () => {
+                    if (!this.fd) {
+                        return;
+                    }
+
+                    fs.closeSync(this.fd);
+                },
+            });
+        } catch {
+            // ignore
```
justinmk, line 104: can we log this somehow? (console.log in the worst case)

ollien: Yeah for sure. Honestly I don't know why I didn't think to, I was just focused on moving this around. No reason to leave it that way

src/logger.ts
ollien: This isn't a listener, but rather sets one up. That said I could move the subscription into the constructor and then rename this; that's definitely more consistent with how this project does things.

src/extension.ts
ollien: Oh interesting. I'll have to play with this a bit later.

CONTRIBUTING.md
justinmk: > Given this, is there something different between what your screenshot does and what I have written here?

Oh my mistake. I see also that one can hover over the checkmark icon in that menu to `Set as default`:
image

CONTRIBUTING.md
ollien: Weird, it doesn't make a difference in the VSCode preview pane, but maybe that renderer handles it differently.

src/extension.ts
ollien: Huh, I actually think that given we don't modify the output channel log level, these two are actually equivalent. That said, I'll change it, as it is definitely clearer to do it that way.

From the [docs](https://code.visualstudio.com/api/references/vscode-api#LogLevel):

> The current log level of the channel. Defaults to [editor log level](https://code.visualstudio.com/api/references/vscode-api#env.logLevel).

justinmk (approved):
It's great to have this, thank you!

ollien:
@theol0403 Yeah totally. I guess I was focusing on in-instance logs 😅

PR #1964, opened 2024-05-14
fix: setup highlights timely

### LINKED ISSUE #1963: bug: "references" preview pane shows tree-sitter highlights after scrolling
### Check the following:

### Neovim version (nvim -v)

v0.9.5

### Operating system/version

Fedora 39

### Describe the bug

When opening the "references" pane, highlighting will (usually) show up correctly. However, once I click on an item to scroll to, the tree-sitter highlights return. A video is worth a thousand words:

https://github.com/vscode-neovim/vscode-neovim/assets/977151/63d35917-6dfb-420d-b374-eb1975e24d56

This seems related to #1739 (and perhaps #1940?). 

### Steps To Reproduce

1. Go to a file where you will have multiple references for a single variable
2. Run "go to references"
3. Click on one of the jump-points, and notice that the highlight colors change.

### Expected Behavior

The tree-sitter highlights should not bleed through.

ollien:
This looks like a missed case from #1944; it intended to add `WinNew`, `BufNew`, but seemingly accidentally removed `WinScrolled`!

Fixes #1963

### DIFF
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -164,12 +164,9 @@ local function setup()
         -- highlights of custom namespace
         setup_syntax_overrides()
       end
-      if ev.event == "Syntax" then
-        -- wait syntax things done
-        vim.defer_fn(setup_syntax_groups, 200)
-      else
-        setup_syntax_groups()
-      end
+      setup_syntax_groups()
+      -- wait syntax things done
+      vim.defer_fn(setup_syntax_groups, 200)
     end,
   })
   api.nvim_create_autocmd({ "VimEnter", "UIEnter" }, {
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -148,7 +148,7 @@ end
 
 local function setup()
   local group = api.nvim_create_augroup("VSCodeNeovimHighlight", { clear = true })
-  api.nvim_create_autocmd({ "WinNew", "WinEnter", "BufNew", "BufEnter", "BufWinEnter" }, {
+  api.nvim_create_autocmd({ "WinNew", "WinEnter", "WinScrolled", "BufNew", "BufEnter", "BufWinEnter" }, {
     group = group,
     callback = function()
       setup_win_hl_ns()
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -148,6 +148,8 @@ end
 
 local function setup()
   local group = api.nvim_create_augroup("VSCodeNeovimHighlight", { clear = true })
+  -- WinScrolled is perhaps not strictly necessary (or shouldn't be), but is useful in ensuring
+  -- we don't get syntax highlights in peek windows.
   api.nvim_create_autocmd({ "WinNew", "WinEnter", "WinScrolled", "BufNew", "BufEnter", "BufWinEnter" }, {
     group = group,
     callback = function()


xiyaowong:
😂In fact, I found this problem after that pr. I've been wondering if I should add it back

runtime/lua/vscode-neovim/highlight.lua
```diff
@@ -164,12 +164,9 @@ local function setup()
         -- highlights of custom namespace
         setup_syntax_overrides()
```
ollien, line 165: Uninformed question: is there a reason not to remove this `ColorScheme` conditional as well?

xiyaowong: Since these two methods have already been executed, there is no need to execute them again unless a theme is set

ollien: Gotcha. I figured there was no harm in re-executing it (just like with `setup_syntax_groups`), but that make sense. I'm guessing `setup_syntax_groups` is just needed more often than we already execute it?

xiyaowong:
In the past, it was considered that the external buffer (currently just vim help) needed to be highlighted, so the method of highlighting the namespace was adopted. However, almost 99% of users don’t use it. We can consider simply clearing the highlighting of the global namespace, which not only simplifies the code, but also makes it more stable. @theol0403 @justinmk

runtime/lua/vscode-neovim/highlight.lua
xiyaowong: Because setup_syntax_groups cleans up specific syntax highlighting groups, and these highlights are generally defined only after opening a file.

xiyaowong:
@ollien Additionally, I intentionally removed WinScrolled before. This is because it was already being executed on WinNew and BufferNew, setting all windows and buffers. In theory, the issue that this PR is attempting to solve should not occur.

theol0403:
I vote we keep this functionality because in the future we do want to support any external buffers and their highlightings.

xiyaowong:
I'm on the fense

ollien:
> @ollien Additionally, I intentionally removed WinScrolled before. This is because it was already being executed on WinNew and BufferNew, setting all windows and buffers. In theory, the issue that this PR is attempting to solve should not occur @_@

Yeah, I'm not really sure how treesitter is getting re-enabled (or is it? Educated guess); I didn't dive into the events here, but maybe I'll take the time soon.

xiyaowong:
> I vote we keep this functionality because in the future we do want to support any external buffers and their highlightings.

I'm on the fence

ollien:
> I'm on the fense

I agree with Theo for now, or at the very least it should be a separate PR. If, somehow, we can prevent these highlights from bleeding into actual code while preserving external buffer highlights, that would be the most ideal, but I understand where you're coming from

xiyaowong (approved):

justinmk:
> I vote we keep this functionality because in the future we do want to support any external buffers and their highlightings.

Highlighting seems like something that vscode should own. OTOH, custom plugin buffers and :help of course won't be highlighted by vscode. 

There's also an argument that as treesitter drives more functionality (such as navigation of "nodes"), it will be needed for functionality beyond highlighting. But that can be supported while ignoring the *color information* coming from Nvim. If that simplifies vscode-neovim, I don't think :help is worth the complication.

justinmk (approved):

ollien:
> But that can be supported while ignoring the color information coming from Nvim

Do you know how this might be done, or at least have a pointer in the right direction? As far as I can tell, these colors are sent as ordinary highlights. I thought about ignoring `syntax` highlight groups, but I'm sure that breaks various extensions.

ollien:
> @ollien Additionally, I intentionally removed WinScrolled before. This is because it was already being executed on WinNew and BufferNew, setting all windows and buffers. In theory, the issue that this PR is attempting to solve should not occur @_@

I spent some time digging into this, and noticed an interesting quirk! It seems that opening the peek panel does not trigger any of the events in this `AutoCmd`. I modified the script slightly to print out the event received (excluding `WinScrolled`, as the Output panel got into an infinite loop with it), and you'll notice that the events aren't triggered when opening the panel.


https://github.com/vscode-neovim/vscode-neovim/assets/977151/c3ea175d-5e53-4633-9aaf-b07e53e091ae

I don't really get why the highlights return, given we've not changed buffers (at least according to the events), but at least it helps indicate why `BufNew` and friends don't help us out here... `WinScrolled` seems to be a close enough approximation

PR #1966, opened 2024-05-15
fix(eval): try serializing the object result

xiyaowong:


src/actions_eval.ts
```diff
@@ -26,12 +26,17 @@ export async function eval_for_client(code: string, args: any): Promise<any> {
...
     void args;
 
-    const value_type = typeof result;
-    if (value_type === "object") {
-        return String(result);
-    } else if (value_type === "function") {
-        return `[Function: ${result.name}]`;
-    } else {
-        return result;
+    if (typeof result === "object") {
```
ollien, line 29: Might it make sense to remove the conditional and change this to something like

```ts
let data: string | undefined;
try {
     data = JSON.stringify(result);
} catch (e) { /* ... */ }

if (data === undefined)  {
  throw new Error(`Invalid data from eval data: ${result}`)
}

return JSON.decode(data); // or maybe even just `return result`;
```

This ensures that the data we return to Lua is serializable as JSON, and avoids the case your comment warns about

ollien (approved):

src/actions_eval.ts
xiyaowong: > or maybe even just `return result`, since we know the data can safely be serialized.

For
```js
const obj = { a() {} }
```
- JSON.stringify: ✔
- neovim transform: ❌

Directly returning the result will fail for commonly encountered objects like TextEditor and Document.

justinmk (approved):

src/actions_eval.ts
justinmk: edge cases like that would be helpful to note in the commit message

mbway:
btw the choice not to convert objects implicitly was intentional to make it clear that the value returned from `eval` is not a handle to the object from vscode, but I don't mind either approach.

With this change the README is now out of date though and there is at least one place (`ui.lua`) where a call to `JSON.stringify` is no longer necessary.

ollien:
> With this change the README is now out of date though and there is at least one place (ui.lua) where a call to JSON.stringify is no longer necessary.

If you would be so kind to submit a PR, that would be much appreciated :)  Otherwise, I can take a look in the next couple of days.

mbway:
@ollien on it now :)

PR #1971, opened 2024-05-17
fix(incsearch): win cursor not properly revealed

### LINKED ISSUE #1969: `incsearch` not works in NVIM v0.10.0
See https://github.com/neovim/neovim/issues/28800

xiyaowong:
- close #1969

Temporary solution. Need to refactor cursor manager and viewport manager. Merge the logic and redefine the necessity of some custom events, such as visual-changed and window-scroll.

Note: Handling the win_viewport event directly in the cursor manager does not work

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -111,6 +111,7 @@ export class CursorManager implements Disposable {
                     window.visibleTextEditors.forEach((e) => (e.options.cursorStyle = style));
                 },
             },
+            main.viewportManager.onCursorChanged((grid) => this.gridCursorUpdates.addForceUpdate(grid)),
         );
     }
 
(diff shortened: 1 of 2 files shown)


ollien:
Do you mind giving a before/after screencast? I'm not clear exactly what this is fixing. (I have some morbid curiosity because I am attempting to address something potentially similar right now, so I wonder if our fixes combined are a complete solution...)

xiyaowong:
It may not be the same issue.

ollien:
Ah, this explains why my CI was failing inconsistently, then, even though it worked locally :)

ollien (approved):

xiyaowong:
Maybe you are using nvim0.9

xiyaowong:
Merged. Approaching the weekend, reviews are slowing down. Not waiting this time. 🙈

ollien:
I am. Go for it, as long as it still works with 0.9

Issue #329, opened 2020-08-28
Enable incremental search and highlight while typing a search term in command mode?


bangedorrunt:
hi,
in neovim, we can do that with
```
set incsearch hl
set inccommand=split
```
but this doesn't work on *vscode-neovim*, i also installed [ haya14busa/is.vim](https://github.com/haya14busa/is.vim) but it couldn't help.

asvetliakov:
`incsearch` works just fine. Check if you've messed something with extension HL configuration

```
`set inccommand=split`
```
can't work because it creates another window in neovim. This can't be mapped to vscode

asvetliakov:
Need to force `set inccommand=nosplit` in extension options

babygau:
> Need to force `set inccommand=nosplit` in extension options

it doesn't help, i tried 

`%s/search/replace` 

but increment highlight is not shown. however doing `/search` works for me.

asvetliakov:
So few things are happening here:

1. Search mode doesn't work in command mode because it tries to insert `<tab>` for completion. (simple cases like `/abc` or `?abc` are excluded for completion). That i'll fix soon
2. I tried to enable `inccommand=nosplit` and everything is worked, even neovim sends `buf_lines_event` with removed text after typing `/` before `replace`, BUT it doesn't revert the text back. So sounds like upstream issue,  `inccommand=nosplit` can't be supported now.

asvetliakov:
Actually for `nosplit` we can ignore buf change event and just use it for HL

babygau:
thank you for your explain! hope that will be sorted 👍

awilkins:
Incremental search with `/` works well for me, but `%s/` is really jumpy - it seems to re-search for the first hit every time you type a char.

theol0403:
Closed in favor of #607

theol0403:
Currently, it seems like the new extmark support shows the replaced text, but does not remove the original. It's possible improved extmark support is needed to make this work.

theol0403:
Fixes #607 - the ability to not force config option should solve it.

ollien:
> Incremental search with `/` works well for me, but `%s/` is really jumpy - it seems to re-search for the first hit every time you type a char.

[...]

PR #1968, opened 2024-05-17
test: rework vi{ test

xiyaowong:
- Close #116

PR #1973, opened 2024-05-17
fix: reduce text jumpiness when using the substitute command

ollien:
Partially addresses #329 

Both nvim and vscode-neovim support incremental highlights while doing a substitute command (e.g. `:%s/partia`). However, vscode-neovim's support was a bit jumpy. With this change, things are much improved, but we don't avoid the initial scroll to the top (some more details can be found in my comments on that issue). 

Part of the reason for the jumpiness is every time a user types a character in the command, we completely rewrite what's in the command box. We can solve the common case by simply looking to see if the end of the text was changed, which should reduce jumpiness during simple typing

Before (I am typing `%s/poke` while the first instance is out of viewport range):

https://github.com/vscode-neovim/vscode-neovim/assets/977151/20c0344a-ad67-48d3-ac07-efb57aff2fde

After:

https://github.com/vscode-neovim/vscode-neovim/assets/977151/9a73696d-53ff-4f79-9421-d75b18850d50


src/utils/text.ts
```diff
@@ -100,3 +115,17 @@ export function normalizeInputString(str: string, wrapEnter = true): string {
     }
     return finalStr;
 }
+
+export function diffLineText(oldLine: string, newLine: string): TextChange {
```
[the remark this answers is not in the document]
ollien: Ah really? I just rewrote it, I had no idea it was ever here :sweat_smile: 

I put it in `utils` mostly so it could be unit-tested. We can't run unit tests against `command_line_manager` (due to the `vscode` import), and making a separate module just for this one function felt a bit silly.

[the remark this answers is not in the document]
xiyaowong: Perhaps no need for unit tests? 🙈

xiyaowong:
image
FYI, the GitHub package mirror source has not yet been synchronized to NVIM 0.10.0, so it will fail.

ollien:
Does this mean we have tests we expect to fail on 0.9?

xiyaowong:
Yes, we always use the latest stable version of Neovim. Test failure does not mean that related functions cannot be used on older versions

xiyaowong:
Yes, the failed test case was fixed in NVIM 0.10.0. https://github.com/vscode-neovim/vscode-neovim/pull/1968

theol0403 (approved):
Clever!

src/command_line/command_line_controller.ts
justinmk, line 1: `cmdline` is the usual name for this. if we're renaming this let's name it `cmdline_controller.ts`

though i have no idea what the difference is between "manager" and "controller"...

src/command_line/command_line_text.ts
justinmk, line 1: `cmdline_text.ts`

src/command_line/command_line_text.ts
```diff
@@ -0,0 +1,58 @@
...
+    } else if (change.action === "none") {
+        // If no change, type nothing.
+        return "";
+    } else {
+        // Rewrite the line if it's not a simple change
+        return `<C-u>${normalizeInputString(newText)}`;
+    }
+}
+
+export function diffLineText(oldLine: string, newLine: string): TextChange {
```
justinmk, line 33: we have a "diff" library, would it help here? https://github.com/vscode-neovim/vscode-neovim/blob/3d33a4ac09574ecb46602a4c8944bf346e4500b4/package.json#L2159

maybe not, it looks like this function is doing a "compare", not a "diff"

src/command_line/command_line_controller.ts
ollien: The "controller" is what handles the actual popup. It has webdev parallels, so I guess it makes sense

As for `cmdline`, sure. I'll do that for all the others too

src/command_line/command_line_text.ts
```diff
@@ -0,0 +1,58 @@
...
+    action: "none";
+}
+
+interface OtherTextChange {
+    action: "other";
+}
+
+export function calculateInputAfterTextChange(oldText: string, newText: string): string {
+    const change = diffLineText(oldText, newText);
+    if (change.action === "added") {
```
justinmk, line 20: `switch` might be appropriate here

justinmk (approved):

src/command_line/command_line_text.ts
ollien: Hm, interesting! I hadn't considered using that. My initial mental prototype was to use something like minimum edit distance to reconstruct the string, but aside from being a lot of work for such a small thing, I was worried about possible desync between our image of the command line and neovim's.

I think that might be interesting if we were ever to make this check more complicated. For such a simple "diff" check, I don't think we actually need to rely on something with so much power

PR #1981, opened 2024-05-18
ci: update homebrew formulae before running tests

### LINKED ISSUE #1980: macOS CI downloading Neovim 0.9.5
Currently, our CI is failing on macOS due to some incompatibilities in our tests on Neovim 0.9 vs 0.10. 

@xiyaowong mentioned [here](https://github.com/vscode-neovim/vscode-neovim/pull/1973#issuecomment-2118011634) that the macOS run is downloading Neovim 0.9.5 because of caching with Github and Homebrew. From what I can tell though, Github seems [perfectly aware of Neovim `0.10.0`](https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fneovim/217142283?tag=0.10.0), so I'm not sure what exactly needs to be updated...? It seems the `rhysd/action-setup-vim` action is indeed picking up that we want the latest stable version, so I don't think it's an issue with our configuration.

ollien:
Closes #1980. I must admit, I'm no Homebrew expert (I actually don't have a mac handy), but from my testing (and reading of the [man page](https://docs.brew.sh/Manpage#update-options) this forces Homebrew to update its caches every time the CI runs. The only downside (aside from time) is it also upgrade Homebrew itself, which may not be fully desirable, but unfortunately I don't think there's a way to tell homebrew to only upgrade formulae (i.e. there is no equivalent to `apt-cache update` or some such)

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -16,11 +16,17 @@ jobs:
         runs-on: ${{ matrix.os }}
         steps:
             - uses: actions/checkout@v4
+
             - name: Setup Node
               uses: actions/setup-node@v4
               with:
                   node-version: "18"
                   cache: "npm"
+
+            - name: Update Homebrew Caches
+              if: matrix.os == 'macos-latest'
+              run: brew update
+
             - name: Install Neovim
               uses: rhysd/action-setup-vim@v1
               id: vim
@@ -30,10 +36,13 @@ jobs:
                   # version: v0.9.0
             - name: npm install
               run: npm ci --silent
+
             - name: Build
               run: npm run test-compile
+
             - name: Webpack
               run: npm run webpack
+
             - name: Test
               uses: coactions/setup-xvfb@v1
               env:


xiyaowong:
Nice. I have a branch just like you to do it, and I went to bed because it was too late.

xiyaowong (approved):

.github/workflows/build_test.yml
```diff
@@ -16,11 +16,17 @@ jobs:
...
               with:
                   node-version: "18"
                   cache: "npm"
+
+            - name: Update Homebrew Caches
+              if: matrix.os == 'macos-latest'
+              run: brew update
+
             - name: Install Neovim
               uses: rhysd/action-setup-vim@v1
```
justinmk, line 31: With this action, homebrew shouldn't be needed.

Why are we using homebrew? It is very slow in CI, it can add minutes.

ollien: > With this action, homebrew shouldn't be needed.

This action _is_ using homebrew, this is not a new addition. 

> Why are we using homebrew? It is very slow in CI, it can add minutes.

Debatably, we don't need to and could just fetch from the neovim repository directly.

justinmk: > we don't need to and could just fetch from the neovim repository directly.

I thought action-setup-vim was doing that (years ago, that is how it worked...)

https://github.com/neovim/node-client/ and https://github.com/neovim/pynvim fetch from https://github.com/neovim/neovim/releases directly, and if it saves minutes in CI that's probably worth doing.

ollien: Picking a couple of random examples, it seems that on non-linux platforms, the installation takes ~10s, and this `update` step takes ~15s. Definitely not quick, but not a huge contributor towards our ~8-13 minute runtime.

PR #1982, opened 2024-05-18
chore: update eval docs and tests

mbway:
some small changes to match the changes introduced in https://github.com/vscode-neovim/vscode-neovim/pull/1966
- README updated
- added a few test cases
- removed redundant use of JSON.stringify

### DIFF
--- a/src/test/integ/actions.test.ts
+++ b/src/test/integ/actions.test.ts
@@ -64,7 +64,13 @@ describe("Eval VSCode", () => {
         output = await eval_from_nvim(client, "function foo() {}; return foo;");
         assert.equal(output, null);
 
-        await assert.rejects(() => eval_from_nvim(client, "const a = {}; a.a = a; return a"));
+        output = await eval_from_nvim(client, "function foo() {}; return {a: foo, b: 123};");
+        assert.deepEqual(output, { b: 123 });
+
+        await assert.rejects(
+            () => eval_from_nvim(client, "const a = {}; a.a = a; return a"),
+            /Error executing lua Return value of eval not JSON serializable: TypeError: Converting circular structure to JSON.*/,
+        );
 
         output = await eval_from_nvim(client, "function f(v) {return 100 + v;}; return f(2);");
         assert.equal(output, 102);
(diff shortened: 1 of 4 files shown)


mbway:
I think the test failure is unrelated to this PR

src/test/integ/actions.test.ts
```diff
@@ -64,7 +64,13 @@ describe("Eval VSCode", () => {
         output = await eval_from_nvim(client, "function foo() {}; return foo;");
         assert.equal(output, null);
 
-        await assert.rejects(() => eval_from_nvim(client, "const a = {}; a.a = a; return a"));
+        output = await eval_from_nvim(client, "function foo() {}; return {a: foo, b: 123};");
+        assert.deepEqual(output, { b: 123 });
+
+        await assert.rejects(
+            () => eval_from_nvim(client, "const a = {}; a.a = a; return a"),
+            /Error executing lua Return value of eval not JSON serializable: TypeError: Converting circular structure to JSON.*/,
```
ollien, line 72: A bit of a nitpick, but should this map so exactly to the specific TypeError? Maybe just change this to `Error executing lua Return value of eval not JSON serializable.*`, since really that's what we care about?

ollien (approved):
Looks great! And yeah, don't sweat the test failure, it's not you :)

Thanks for doing this!

ollien:
Also, I know there's a dislike of `chore` but I wasn't sure what else to label this, given the span of this PR; just wanted to give it a tag. Feel free to disagree and change it :)

theol0403 (approved):

src/test/integ/actions.test.ts
mbway: my thinking here is that we are testing that the user is given information about why the serialisation failed, but I can make the adjustment if you prefer.

xiyaowong (approved):

xiyaowong:
"chore" is fine

src/test/integ/actions.test.ts
ollien: Nah, don't sweat it :) it's a taste thing. I don't know that we need to guarantee such implementation details. No biggie either way

PR #1976, opened 2024-05-18
fix(highlight): forward search / matching in long lines

xiyaowong:
- fix: cursor position after scrolling with incsearch
- refactor: remove viewport hack when in cmdline mode

---

- Close #1975
- Fix #161
- Related:
	- #1575

Because it involves a few historical matters and dealing with hacks, I won't go into too much detail. Here are some debugging notes.

### Cursor position for incsearch

| Commit History                                         | Nvim 0.9 | Nvim 0.10 |
| ------------------------------------------------------ | -------- | --------- |
| Add test                                               | ❌       | ✅        |
| Remove hack                                            | ❌       | ✅        |
| Update cursor position on window-scroll                | ❌       | ✅        |
| Fire cursor change event on window-scroll              | ❌       | ✅        |
| custom viewport changed event                          | ❌       | ✅        |
| after searching, perform a redraw and wait for a while | ✅       | ✅        |

### Forward search matching in long line

Only tested in Nvim 0.10.

### Conclusion

Highly recommend to upgrade to Nvim 0.10.

PR #1983, opened 2024-05-19
fix: set highlight group Subsitute

### LINKED ISSUE #1979: bug: incremental highlights do not work if `inccommand` is not empty
### Check the following:

### Neovim version (nvim -v)

0.9.5

### Operating system/version

Fedora 39

### Describe the bug

Split from #329 

In #334, we disabled the `inccommand` setting, so that substitution highlights would display. However, if a user sets the configuration to `nosplit` or `nosplit`, they will no longer display.

[Per @theol0403,](https://github.com/vscode-neovim/vscode-neovim/issues/329#issuecomment-1215278530)
> Currently, it seems like the new extmark support shows the replaced text, but does not remove the original. It's possible improved extmark support is needed to make this work.

### Steps To Reproduce

1. Open a buffer with some text
2. Run `:set inccommand=nosplit`
3. Search for that text with `:%s/blah`, and notice that highlights do not occur

### Expected Behavior

vscode-neovim should respect this setting, and we should display the highlights and/or replacements as possible. I imagine `split` might have some difficulty, given it needs another window.

xiyaowong:
chore: remove the border of the custom highlights
      It will be strange when multiple decoration ranges with borders are connected.

- Fix #1979

### DIFF
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -55,6 +55,11 @@ local function setup_globals()
     Visual       = {},
     VisualNC     = {},
     VisualNOS    = {},
+    -- By default, it's linked to Search.
+    -- But Search highlighting defaults to using VSCode ThemeColor, potentially with some opacity.
+    -- When the decoration is "virtText", having a background color with opacity can
+    -- cause the virtual text to blend with the original content.
+    Substitute   = { fg='#282c34', bg='#98c379' },
     Whitespace   = {},
     LineNr       = {},
     LineNrAbove  = {},
--- a/runtime/lua/vscode-neovim/highlight.lua
+++ b/runtime/lua/vscode-neovim/highlight.lua
@@ -55,7 +55,6 @@ local function setup_globals()
     Visual       = {},
     VisualNC     = {},
     VisualNOS    = {},
-    Substitute   = {},
     Whitespace   = {},
     LineNr       = {},
     LineNrAbove  = {},
--- a/package.json
+++ b/package.json
@@ -331,16 +331,16 @@
                     },
                     "default": {
                         "IncSearch": {
-                            "backgroundColor": "theme.editor.findMatchBackground",
-                            "borderColor": "theme.editor.findMatchBorder"
+                            "backgroundColor": "theme.editor.findMatchBackground"
                         },
                         "CurSearch": {
-                            "backgroundColor": "theme.editor.findMatchBackground",
-                            "borderColor": "theme.editor.findMatchBorder"
+                            "backgroundColor": "theme.editor.findMatchBackground"
                         },
                         "Search": {
-                            "backgroundColor": "theme.editor.findMatchHighlightBackground",
-                            "borderColor": "theme.editor.findMatchHighlightBorder"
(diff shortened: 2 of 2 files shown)


xiyaowong:
`set icm=nosplit` works, but sometimes the highlighing is wrong

https://github.com/vscode-neovim/vscode-neovim/assets/47070852/48f659a9-a524-418e-918e-958fb487094d

xiyaowong:
Hard to reproduce under #1976

xiyaowong:
> `set icm=nosplit` works, but sometimes the highlighing is wrong

Hard to reproduce under #1976

runtime/lua/vscode-neovim/highlight.lua
```diff
@@ -55,7 +55,11 @@ local function setup_globals()
     Visual       = {},
     VisualNC     = {},
     VisualNOS    = {},
-    Substitute   = {},
+    -- By default, it's linked to Search.
+    -- But Search highlighting defaults to using VSCode ThemeColor, potentially with some opacity.
+    -- When the decoration is "virtText", having a background color with opacity can
+    -- cause the virtual text to blend with the original content.
+    Substitute   = { fg='#282c34', bg='#98c379' },
```
ollien, line 62: Nice! We should offer a way to customize these colors if possible, though. I can imagine this color not showing very well on some light themes.

ollien:
One thing I noticed, though I figure we can't do much about this, given VSCode owns the buffer, is that any substitutions that exceed the length of the original will write over the original text. 

Similarly (and this one I think we could fix), if your replacement exceeds the line length, it gets weirdly truncated.

image

runtime/lua/vscode-neovim/highlight.lua
xiyaowong: Now it's possible, the `setup_globals` will only execute once, and users can define it themselves, requiring some delay for custom highlighting. However, indeed, there is a need to provide a more conventional approach.

xiyaowong:
Merged, this PR has no side effects

runtime/lua/vscode-neovim/highlight.lua
ollien: Gotcha, I guess I was thinking of allowing it via our configuration, but neovim globals work too

xiyaowong: https://github.com/microsoft/vscode/issues/32813

xiyaowong:
I originally used ThemeColor, but later found that the effect was not good

https://github.com/vscode-neovim/vscode-neovim/blob/c81e64dd0f0502f79b14fe2430d7d442f347fd85/runtime/lua/vscode-neovim/highlight.lua#L58-L62

xiyaowong:
Sure. But now users can still set it in settings.json. If you want to add to this, that's fine with me.

ollien:
>  But now users can still set it in settings.json

How would users do that now? I can guess how they might do it in `init.vim`, but not `settings.json`

xiyaowong:
> > But now users can still set it in settings.json
> 
> How would users do that now? I can guess how they might do it in `init.vim`, but not `settings.json`

https://github.com/vscode-neovim/vscode-neovim#-highlights

PR #1984, opened 2024-05-19
ci: speedup tests by decreasing the default timeout

xiyaowong:


ollien (approved):

justinmk:
Definitely a welcome change. But the real solution will be to have some sort of `waitUntil()` function that retries assertions, instead of having hardcoded timeouts.

PR #1987, opened 2024-05-19
fix: inccommand highlights can exceed line length

ollien:
https://github.com/vscode-neovim/vscode-neovim/assets/977151/c35e1fae-ed65-4104-8eea-e2ddd07195a8


xiyaowong:
image
Haha, I'm dealing with this too

PR #1988, opened 2024-05-19
fix(highlight): fix eol highlight not displaying

xiyaowong:
Close #1987

### DIFF
--- a/src/test/unit/highlights/highlight_grid.test.ts
+++ b/src/test/unit/highlights/highlight_grid.test.ts
@@ -650,6 +650,26 @@ describe("processHighlightCellsEvent", () => {
                 },
             ],
         },
+        {
+            testName: "calculateEditorColFromVimScreenCol eolCol >= 0",
+            events: [
+                {
+                    row: 2,
+                    vimCol: 3,
+                    validCells: [{ hlId: 2, text: "!" }],
+                    lineText: "\t你",
+                    tabSize: 2,
+                },
+            ],
+            expectedRanges: [
+                {
+                    textType: "virtual" as const,
+                    highlights: [{ hlId: 2, text: "!", virtText: "!" }],
+                    line: 12,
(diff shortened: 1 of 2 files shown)


src/utils/text.ts
```diff
@@ -33,7 +37,18 @@ export function calculateEditorColFromVimScreenCol(line: string, screenCol: numb
         }
 
         if (currentCharIdx >= line.length) {
```
xiyaowong, line 39:
```suggestion
        if (currentCharIdx >= wcswidth(line)) {
```
@ollien Here we cannot use width; what we need is character index

xiyaowong, line 39: @ollien Here we cannot use width; what we need is character index

ollien: Nice catch. I had a test failure on one of the double-wide unit tests but must have fixed it otherwise

src/test/unit/highlights/highlight_grid.test.ts
```diff
@@ -630,6 +630,26 @@ describe("processHighlightCellsEvent", () => {
                 },
             ],
         },
```
ollien, line 632: Thanks for carrying this over. If you would be so kind, would you mind adding a case for where the `Math.max` will trigger return 0, when calculating `eolCol`?

ollien (approved):

src/utils/text.ts
```diff
@@ -17,10 +17,14 @@ export function expandTabs(line: string, tabWidth: number): string {
     return expanded;
 }
 
+/**
+ * Note: This function is for use by the highlight module only.
+ */
 export function calculateEditorColFromVimScreenCol(line: string, screenCol: number, tabSize: number): number {
```
justinmk, line 23: off topic: this could be named `getColFromNvimCol` without any loss of clarity, while reducing verbosity.

PR #1985, opened 2024-05-19
fix: Support inccommand, including "split"

xiyaowong:
fix: wildmenu sometimes not showing

https://github.com/vscode-neovim/vscode-neovim/assets/47070852/46b01ac0-09d7-4866-8c9b-a0a0b40be009

PR #1989, opened 2024-05-20
fix: wildmenu items are being set repeatedly

xiyaowong:
A mistake made in #1985


ollien (approved):
Heh, nice! 

Just because I'm curious, what was the "loop" here? I was trying to narrow it down and the best I could find was that commenting this check out seemed to stop the problem, but I couldn't figure out why.

```ts
        if (this.isDisplayed && this.completionItems.length) {
            this.input.items = this.completionItems;
        }
```

xiyaowong:
When in cmdline, nvim will continuously send cmdline_show event.

PR #1997, opened 2024-05-21
refactor(lua): rename the module vscode-neovim to vscode

xiyaowong:
I welcome all feedback! If there's a unanimous agreement that it's unnecessary, I'll close this PR and cease discussing this topic.


README.md
```diff
@@ -318,9 +318,15 @@ For more information:
...
 
 ```lua
-local vscode = require('vscode-neovim')
+local vscode = require('vscode')
 ```
 
+<!-- prettier-ignore-start -->
+> [!TIP]
+> "vscode" is an alias for "vscode-neovim" module.
+> So don't be confused if "vscode-neovim" is used in the below examples.
```
theol0403, line 327: Just change the examples. We can change all wording/internals to use `vscode` and have `vscode-neovim` as an undocumented backwards-compatibility.

justinmk (approved):
This is 100% a good idea. 

vscode-neovim is a very awkward name and should be de-emphasized as much as possible. And there is no risk of conflict with some other "vscode" lua module (we can show an error if we detect such a thing, perhaps)

README.md
```diff
@@ -309,9 +309,15 @@ For more information:
...
```
justinmk, line 318:
```suggestion
> "vscode" is an alias for "vscode-neovim" (deprecated) module.
> So don't be confused if "vscode-neovim" is used in old discussions or other resources.
```

justinmk (approved):
🚀  Names matter. Less is more. 🚀

README.md
xiyaowong: Oh, I forgot to remove this section

xiyaowong:
Obviously more concise and intuitive!

theol0403 (approved):

PR #2001, opened 2024-05-23
feat(cmdline): integrate with native neovim completions

theol0403:
This is my favourite type of PR, where bugs are fixed simply by deleting old crusty code :smile: 

- [x] Forward all cmdline bindings to nvim
- [x] Use the quickpick simply as a "render view" to nvim
- [x] Support wildmode
- [x] Support official nvim bindings for Up/Down/C-p/C-n/Tab (with clever behavior of history vs suggestions) 
- [x] Fix jitter when setting vscode selected item
- [x] Fix edge cases when typing/selecting quickly
- [x] Fix inccommand module
- [x] Make sure deleted code/timers aren't a regression

https://github.com/vscode-neovim/vscode-neovim/assets/16546293/e3217f41-83a5-4138-ba19-f2d7533daf02

Note that completions have to be triggered by wildchar, just like in nvim.


xiyaowong:
Can we merge cmdline_controller and cmdline_text into cmdline_manager?

src/main_controller.ts
```diff
@@ -149,7 +149,6 @@ export class MainController implements vscode.Disposable {
             ext_multigrid: true,
             ext_popupmenu: true,
             ext_tabline: true,
-            ext_wildmenu: true,
```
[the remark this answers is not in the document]
xiyaowong: ext_wildmenu is deprecated.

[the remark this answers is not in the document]
xiyaowong: Improvements are mainly by:
image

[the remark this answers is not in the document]
theol0403: Yeah, not technically required for this improvement, but I wanted to take the chance to move away from deprecated api.

ollien:
> Can we merge cmdline_controller and cmdline_text into cmdline_manager?

`cmdline_text` is separate to facilitate a unit test. I'd love it if we didn't have to split modules to do, but alas.

theol0403:
Really good progress made. Code refactored even more. Stupidly simple now, with very basic states. Works amazingly. Only last challenge is dealing with race conditions when rapidly selecting completions.

src/cmdline_manager.ts
```diff
@@ -28,6 +28,16 @@ export class CommandLineManager implements Disposable {
     // However, bindings are expected to cause the cmdline content to change, so we use this flag to listen to those updates.
     private redrawExpected = true;
 
+    // Much effort was put into this to make sure that race conditions behave as expected.
```
theol0403: This is difficult, and I'm not fully happy with the solution. If anyone has suggestions for this, it would be appreciated. 

The core problem is that when we spam nvim shortcuts, it triggers a cmdline_show, which we then use to update the quickpick. But then, the onChanged callback is triggered, which if we use to send anything to nvim, it breaks the completions. It is not enough to compare the onChanged text with the last sent text from nvim, because onChanged can lag behind. 

But, if we debounce the onChanged, we create a problem where typing a command and rapidly using a shortcut like C-h causes nvim to operate on outdated text. 

I haven't found a robust way to handle both cases yet.

theol0403:
Fixed it! Nothing like a good night's sleep & a walk.

runtime/modules/inccommand.lua
```diff
@@ -38,17 +38,19 @@ local function refresh_completion()
...
-  local items = get_items()
-  if items then
-    vim.rpcnotify(vim.g.vscode_channel, "redraw", { "wildmenu_show", { items } })
+  local lines = get_lines()
+  if lines then
+    local items = {}
+    for _, line in ipairs(lines) do
+      table.insert(items, { line, "", "", "" })
+    end
+    vim.rpcnotify(vim.g.vscode_channel, "redraw", { "popupmenu_show", { items, -1, 0, 0, -1 } })
```
justinmk, line 47: Interesting. Are we "faking" a Nvim redraw event?

src/utils/cmdline_text.ts
```diff
@@ -31,20 +31,6 @@ export function calculateInputAfterTextChange(oldText: string, newText: string):
     }
 }
 
-export function commandInputIsCompletable(command: string): boolean {
-    return (
-        command.charAt(0) !== "?" &&
-        command.charAt(0) !== "/" &&
-        !command.includes("s/") &&
-        !command.includes("su/") &&
```
justinmk, line 39: I was suspicious about this, it looked like a fragile hack. Good to eliminate it.

justinmk (approved):

src/utils/cmdline_text.ts
ollien: +1, it's not great but it covers a lot of common cases

src/cmdline_manager.ts
```diff
@@ -136,32 +129,24 @@ export class CommandLineManager implements Disposable {
...
-    };
-
-    private onChangeDebouncedInner = debounce(this.onChange, 100, { leading: false, trailing: true });
-
-    private onChangeDebounced = (text: string): void => {
-        this.onChangeDebouncedInner(text);
-        if (!this.redrawExpected) {
-            this.onChangeDebouncedInner.flush();
+        if (this.pendingNvimUpdates) {
+            this.pendingNvimUpdates--;
```
ollien, line 146: I don't think this could happen, but we could stand to be a bit defensive here and do `this.pendingNvimUpdates = Math.max(0, this.pendingNvimUpdates)`. If `this.pendingNvimUpdates` were to somehow get negative, we would continue taking this branch forever (since all non-zero values are truthy)

src/cmdline_manager.ts
```diff
@@ -1,128 +1,208 @@
...
+                }
+                this.redrawExpected = false;
+                this.input.show();
+                if (this.input.value !== content) {
+                    this.pendingNvimUpdates++;
+                    const activeItems = this.input.activeItems; // backup selections
+                    this.input.value = content; // update content
+                    this.input.activeItems = activeItems; // restore selections
+                }
+                break;
```
ollien, line 100: Nitpick: I might move all of this into its own method, just to keep this switch statement slim

package.json
```diff
@@ -995,6 +989,42 @@
...
+                "key": "down",
+                "when": "neovim.init && neovim.mode == cmdline",
+                "args": "<down>"
+            },
+            {
+                "command": "vscode-neovim.send-cmdline",
+                "key": "up",
+                "when": "neovim.init && neovim.mode == cmdline",
+                "args": "<up>"
+            },
```
ollien, line 1015: Huh, I would have thought this change would have allowed for up/down to go through command history, but it doesn't on my build. Is that expected?

runtime/modules/inccommand.lua
theol0403: Yeah. This isn't a core feature, but just a lua gimmick which feeds the inccommand preview buffer into the cmdline suggestions. It's self-contained so not a maintenance burden/complexity add.

package.json
theol0403: Yes, up/down works for me, so does C-n/C-p for navigating history.

src/cmdline_manager.ts
```diff
@@ -1,128 +1,208 @@
...
+            commands.registerCommand("vscode-neovim.test-cmdline", this.testCmdline),
+        );
     }
 
     public dispose() {
-        this.commandLine?.dispose();
         disposeAll(this.disposables);
     }
 
+    private reset() {
```
ollien, line 67: This feels brittle under maintenance; if someone adds a property, they'll have to remember to add it here. Might it be best to encapsulate this state in a class, and then upon the call to `reset`, just reconstruct a new object? something like

```ts
private reset () {
	this.cmdlineState = new CmdilneState();
	// and maybe still do the `this.input` resetting, since we can't encapsulate that
}
````

package.json
ollien: Hm... I wonder why it doesn't work for me.


https://github.com/vscode-neovim/vscode-neovim/assets/977151/c8b33c22-46df-436e-9689-b9d318d11628

theol0403: Do you maybe have conflicting bindings in your vscode? 

Actually, mapping the arrow keys in the quickpick might only currently be supported in vscode-insiders. It will begin to work on a later vscode update. https://github.com/microsoft/vscode/pull/211756

theol0403: Yeah, that PR has not been released yet.

ollien: ah yep, I'm not on Insiders

> no but it will be released for the May iteration. Likely released first week of June.

Sounds good :)

ollien (approved):
Nice! Lookin spiffy. Just gotta fix the commit messages :)

PR #2002, opened 2024-05-23
refactor: extends the Custom Disposable

xiyaowong:


justinmk:
what problem is this solving?

it can be named `Disposable`, there is little risk of conflict. We aren't going to use both at the same time. And imports can be "namespaced" via `import * as foo`

xiyaowong:
Reduce some boilerplate code

justinmk:
there are currently 64 usages of `Disposable` in the project. some of those would be replaced by this new class. And the new class could be namespaced if necessary, e.g. `util.Disposable`. What is the problem?

xiyaowong:
I just thought there could be some improvements here. Now I'm not feeling much improvement.

PR #2005, opened 2024-05-25
chore: package.json: add description to check vscode

erlangparasu:


theol0403:
I think it should just be `vim.g.vscode`, the nil check does the opposite.

package.json
```diff
@@ -73,19 +73,19 @@
                     "type": "string",
                     "default": "",
                     "title": "Custom init.vim path on Windows",
-                    "markdownDescription": "Path to alternative `init.vim` file on Windows, equivalent to Nvim startup option `-u`. If `useWSL` is enabled, the value of the `neovimInitVimPaths.linux` setting will be used to find the init.vim file instead. You can also use `exists('g:vscode')` in your `init.vim` to check if Nvim is running in vscode."
+                    "markdownDescription": "Path to alternative `init.vim` file on Windows, equivalent to Nvim startup option `-u`. If `useWSL` is enabled, the value of the `neovimInitVimPaths.linux` setting will be used to find the init.vim file instead. You can also use `exists('g:vscode')` in your `init.vim`, or `vim.g.vscode == nil` in your `init.lua` to check if Nvim is running in vscode."
```
justinmk, line 76:
```suggestion
                    "markdownDescription": "Path to alternative `init.vim` file on Windows, equivalent to Nvim startup option `-u`. If `useWSL` is enabled, the value of the `neovimInitVimPaths.linux` setting will be used to find the init.vim file instead. You can also use `exists('g:vscode')` in your `init.vim` (or `vim.g.vscode` in `init.lua`) to check if Nvim is running in vscode."
```

same for the others

erlangparasu: Thanks!

justinmk (approved):

ollien (approved):

PR #2011, opened 2024-05-27
refactor: set enforced options in Lua

xiyaowong:


theol0403:
What does this solve? Don't we need to set the forced options *after* the user's init has finished?

xiyaowong:
1. More timely
2. The VimEnter event is executed after loading the vimrc.
3. The purpose of "-S" doesn't belong here originally.
4. I don't see any distinction between this and other modules; there's no need to treat it as a special case within the main_controller.

theol0403:
There is a special distinction. 

The order should be:
- lua modules are loaded (and default options set)
- user config is loaded (which can require("vscode")), which can overwrite default options
- force options are loaded, which **overwrite user options**

xiyaowong:
> Additionally, the autocommand is just for buffer local options, not global options.

👍

xiyaowong:
Currently: default -> force -> user -> force

theol0403 (approved):
Please test to make sure no behavior is changed.

xiyaowong:
Did a simple test, should be ok.
image

Issue #2015, opened 2024-05-27
assets/images


xiyaowong:

xiyaowong:
![multicursor](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/72b3f2b3-6d80-4ace-b6f4-b211baad74d1)
![wildmenu](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/92d03258-0bf0-40cb-89ff-f53754675dcf)

xiyaowong:
![select-next](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/5a93c87e-626a-4a70-a9ef-5084f747c7ef)

xiyaowong:
![refactor](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/1c436b76-5c0b-42a7-8eb4-6f149761dd3c)

xiyaowong:
![image](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/816e9734-bac3-4bbb-ab7f-13bf2e364794)

PR #2016, opened 2024-05-27
chore: update image url in readme

xiyaowong:


README.md
```diff
@@ -275,7 +275,7 @@ To spawn multiple cursors from visual line/block modes type <kbd>ma</kbd>/<kbd>m
 
 See gif in action:
 
-![multicursors](/images/multicursor.gif)
+![multicursor](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/72b3f2b3-6d80-4ace-b6f4-b211baad74d1)
```
justinmk, line 278: Does this point to the image in https://github.com/vscode-neovim/vscode-neovim/issues/2015 ? How did you figure out the right URL? 

Please explain things instead of just merging PRs without explanation.

xiyaowong: I think this is a very common operation on GitHub, there's nothing much to explain. 
msedge_yBDQUWqhUs

justinmk: When I click on the image in that PR, it takes me to `https://private-user-images.githubusercontent.com/47070852/...`, which is a temporary URL that expires. But now I see that *editing* the comment does indeed expose the permanent URL. Hopefully that continues to work.

In any case, it's not obvious where the image comes from, and how to update it.

PR #2017, opened 2024-05-27
refactor: move "vim/" and merge it into "runtime/"

xiyaowong:


runtime/vscode-neovim.vim
```diff
@@ -2,12 +2,41 @@
...
-let &runtimepath = &runtimepath . ',' . s:currDir . '/vim-altercmd'
+let &runtimepath = &runtimepath . ',' . s:currDir
+
+" Load atltercmd
+" {{{
+" altercmd - Alter built-in Ex commands by your own ones
+" Version: 0.0.1
+" Copyright (C) 2009-2015 Kana Natsuno <http://whileimautomaton.net/>
+" License: MIT license
+function! s:altercmd_define(...)
```
justinmk, line 13: interesting, did you inline the only part of AlterCmd that we need? so that allowed us to delete the rest of it? that is something worth mentioning in the PR description & commit message.

runtime/vscode-neovim.vim
```diff
@@ -2,12 +2,41 @@
 let g:vscode = 1
 
 let s:currDir = fnamemodify(resolve(expand('<sfile>:p')), ':h')
-" Adjust rtp path
-let &runtimepath = &runtimepath . ',' . s:currDir . '/vim-altercmd'
+let &runtimepath = &runtimepath . ',' . s:currDir
+
+" Load atltercmd
```
justinmk, line 7: typo

runtime/vscode-neovim.vim
xiyaowong: altercmd has only this piece of code 😂

theol0403 (approved):

xiyaowong:
In case you haven't noticed, just a heads-up: I've changed "runtime/modules" to "runtime/vscode." Any thoughts on this?

justinmk:
> I've changed "runtime/modules" to "runtime/vscode." Any thoughts on this?

I think it makes sense after some thought. Is this the correct logic?:

- runtime/ is where runtime files go (obviously), thus the old top-level vim/ directory was questionable
- runtime/lua/ is for everything that is part of the require('vscode') plugin (available to users)
- runtime/vscode/ is for internal code
    - this name is better than runtime/modules/ IMO. though it might be confusing for contributors wondering why it's not in the require('vscode') module. maybe `runtime/internal/` ?

xiyaowong:
> * runtime/ is where runtime files go (obviously), thus the old top-level vim/ directory was questionable
> * runtime/lua/ is for everything that is part of the require('vscode') plugin (available to users)
> * runtime/vscode/ is for internal code

Yes

>   * this name is better than runtime/modules/ IMO. though it might be confusing for contributors wondering why it's not in the require('vscode') module. maybe `runtime/internal/` ?

- Within the `runtime/lua` directory, there are essential core files required for vscode-neovim. These files mainly interact with various managers or controllers.
- In `runtime/vscode` (previously `runtime/modules`), it's mostly independent feature implementations or patches for specific needs.
- Since the scripts in Vim don't quite resemble a module, I changed to vscode.

xiyaowong:
I vote for "runtime/vscode" because "internal" would bring some new confusion. In fact, the files here are also very good API examples.

Issue #922, opened 2022-05-26
Include neovim in extension


skyqrose:
(This is a feature suggestion, not a bug report or a request for help.)

I have VSCode installed via linux on a flatpak. Because of sandboxing issues, VSCode can't access my computer's installed neovim installation at `/usr/bin/nvim`. This headache, as well as many others (accessing it in WSL, having to install neovim separately, having to manually check that you're at least version 0.5, etc) would be solved if the extension bundled its own working installation of neovim.

Benefits:
- It just works.

Downsides to doing this:
- Bigger download size. My nvim executable is 4.3M, and I don't know how big shared libraries will be.
- Not customizable. If somebody wants to run their own custom-compiled version of neovim, they also have to have a custom version of the extension.
- Delays neovim updates. If neovim has a security vulnerability and releases a new version upstream, then the extension also needs to release a new version.
- Might have problems being cross platform? Is it possible to include different executables for each target OS/architecture? I don't know enough about how extensions are published to answer this.

Alternatives:
- Better documentation about common troubles people have with accessing their neovim installations.

---

## More detail about my specific sandboxing issues

Just so people can find this page in searches and learn from my pain, not because it's important for this feature request.

My versions and stuff:
Linux 5.17.5
PopOS! 22.04
VSCodium version 1.67.2 (A non-telemetry flatpak repackaging of VSCode)
VSCode Neovim v0.0.85

When I set `"vscode-neovim.neovimExecutablePaths.linux": "/usr/bin/nvim"` and vscode can't find the executable, I get popups saying:
```
MainController: Neovim exited with code: -2
MainController: Neovim was disconnected
MainController: Neovim spawn error: spawn /usr/bin/nvim ENOENT
Source: VSCode Neovim (Extension)
```

I set my [flatpak permissions](https://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access) to `--filesystem=host-os` (easiest via flatseal, though this step might not be necessary), and set `"vscode-neovim.neovimExecutablePaths.linux": "/run/host/usr/bin/nvim"`. But `nvim` still failed to start:
```
MainController: Neovim exited with code: 127
MainController: Neovim was disconnected
```

When I run nvim from a command line within VSCode, it gives
`/run/host/usr/bin/nvim: error while loading shared libraries: liblua5.1-luv.so.0: cannot open shared object file: No such file or directory` (and then exits with code 127)

---

## Related issues
[...]

PR #2018, opened 2024-05-27
feat: Select neovim executable

xiyaowong:
I think this is mainly beneficial for developers. Sometimes when debugging certain issues, switching to neovim isn't convenient. The main "impact" of this PR is the addition of a Statusbar, but it's just four letters, most users should be able to accept it.

Code_-_Insiders_F8vgWNiO17

### DIFF
--- a/package.json
+++ b/package.json
@@ -429,6 +429,10 @@
                 "title": "Neovim: Execute lua code"
             },
             {
+                "command": "vscode-neovim.selectNeovim",
+                "title": "Neovim: Select Neovim Executable"
+            },
+            {
                 "command": "vscode-neovim.send",
                 "title": "Neovim: Send key to neovim, syncing if in insert mode"
             },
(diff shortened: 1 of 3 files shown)


theol0403:
I wouldn't use this. Can we just download the appimage/exe for the version we want and provide it in the neovim path config? Doesn't seem too bad.

justinmk:
if this is really needed, an alternative would be to allow the `vscode-neovim.neovimExecutablePaths` to be a list of strings. Then in the vscode settings UI, the paths (I think?) can be dragged to change the order. And vscode has a way to navigate directly to a setting.

That would avoid most of the code here.

>  Can we just download the appimage/exe for the version we want and provide it in the neovim path config?

That seems like a separate issue. Was discussed in https://github.com/vscode-neovim/vscode-neovim/issues/922 , but I think it's a trap.

xiyaowong:
Yes, currently this feature is for my needs. Typically, users only download a global nvim.

Issue #1993, opened 2024-05-20
feature: integrate `vim.diagnostics` with `vscode.Diagnostics`


theol0403:
`:h diagnostic-api` has a bunch of methods to interact with buffer diagnostics. VScode also provides a [Diagnostics](https://code.visualstudio.com/api/references/vscode-api#languages) API.

The main value for this is to hook `vim.diagnostic.goto_next(severity)` and `vim.diagnostic.goto_prev(severity)`, which many people bind to `]e` and `[e`, `]d`, `[d`, etc.

Should be relatively doable.

xiyaowong:
I did think about it, too. However, I am worried that it will bring more confusion to users than benefits.

ollien:
I might give this a crack. I often find myself trying to use the minimal to find diagnostics, so I would actually use this

theol0403:
Don't we want to support as many bindings out of the box as possible? Maybe we do not need the full integration, but hooking the navigation commands will be beneficial.

xiyaowong:
How about creating a temporary nvim plugin for developing this feature? It would facilitate testing and contributions. Besides vim.diagnostics, there might be other functionalities as well. Once it’s refined, we can consider integrating it into vscode-neovim.

theol0403:
Sounds good to me. There's some functionality from multi-cursor that I want to move to core as well, we should maybe create a "staging" extension to test experimental integrations.

ollien:
How does an extension look, to you, if we have to integrate with VSCode functionality? I haven't investigated the multi-cursor extension too much, so maybe there's an "obvious" answer. Is it just "use a lot of eval"? haha

theol0403:
> I might give this a crack. I often find myself trying to use the minimap to find diagnostics, so I would actually use this 

Perhaps we can use this as a proof of concept of one of the first features that would traditionally require support code on the extension side but we implement purely through Lua.

theol0403:
Yeah traditionally we'd create a custom request and provide support on the vscode side. But we can try using only eval.

ollien:
I'd prefer to avoid eval, as I think the semantics of that might be a bit weird (especially if we're using typescript, we'd have to compile down something minified to shove through, or put it all in lua. Neither is super elegant).

I need to look more closely at the request mechanism between the lua and TS side, but maybe we can have extensions do a single eval to "register" what requests they want somehow.

theol0403:
[...]

PR #2020, opened 2024-05-27
feat: (prototype) add support for vim diagnostics

ollien:
Relates to #1993

This is a prototype I threw together while my flight was delayed... I wouldn't review it _too_ closely, but if you have any broad thoughts about approach, I'd love to hear them!

The virtual text for overlays is ... not great. I suspect there's some bug in how we draw these, but investigating that hasn't been my highest priority.

image

### DIFF
--- a/runtime/lua/vscode/diagnostics.lua
+++ b/runtime/lua/vscode/diagnostics.lua
@@ -1,0 +1,35 @@
+local api = vim.api
+local NS = api.nvim_create_namespace("vscode.highlight")
+
+local M = {}
+
+function dump(o)
+  if type(o) == "table" then
+    local s = "{ "
+    for k, v in pairs(o) do
+      if type(k) ~= "number" then
+        k = '"' .. k .. '"'
+      end
+      s = s .. "[" .. k .. "] = " .. dump(v) .. ","
+    end
+    return s .. "} "
+  else
+    return tostring(o)
+  end
+end
+
+function M.set_diagnostics(bufnr, diagnostics)
+  local vim_diagnostics = {}
+  for _i, diagnostic in pairs(diagnostics) do
+    table.insert(vim_diagnostics, {
+      col = 0,
+      lnum = diagnostic.line,
+      severity = diagnostic.severity,
+      message = diagnostic.message,
+    })
+  end
+
+  vim.diagnostic.set(NS, bufnr, vim_diagnostics)
+end
+
+return M
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -1,6 +1,7 @@
 local api, fn = vim.api, vim.fn
 
 local util = require("vscode.util")
+local diagnostics = require("vscode.diagnostics")
 
 local M = {}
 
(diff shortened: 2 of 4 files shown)


ollien:
Also, this is not really related to the PR, but I think I'm doing something wrong. `lua vim.diagnostic.goto_prev({severity=vim.diagnostic.severity.ERROR, float=false})` works, but `nnoremap ]e lua vim.diagnostic.goto_prev({severity=vim.diagnostic.severity.ERROR, float=false})` then pressing `]e` prints "Already at oldest change"...? Not sure why that is, doesn't quite make sense to me

justinmk:
Not in favor. This is non-trivial code that will grow, and need to be debugged, and currently the [justification](https://github.com/vscode-neovim/vscode-neovim/issues/1993) is that vscode doesn't have a mapping that navigates by severity...?

ollien:
I don't think it's _that_ complex, but I do understand your hesitation. There's already more loose ends popping up as I work...

xiyaowong:
> The virtual text for overlays is ... not great. I suspect there's some bug in how we draw these, but investigating that hasn't been my highest priority.

We absolutely only need to implement the navigation feature.

ollien:
> > The virtual text for overlays is ... not great. I suspect there's some bug in how we draw these, but investigating that hasn't been my highest priority.
> 
> We absolutely only need to implement the navigation feature.

Agreed. I don't want these overlays here necessarily, it's just that it may be informing of us of another bug with virtual text.

xiyaowong:
I have a use case. I set up the following mappings to navigate errors, but it jumps across all files with errors, and the order seems uncertain too.

```vim
nnoremap <silent> [g <Cmd>call VSCodeNotify('editor.action.marker.prevInFiles')<CR>
nnoremap <silent> ]g <Cmd>call VSCodeNotify('editor.action.marker.nextInFiles')<CR>
```
To reiterate, what we need is navigation functionality, not actual management of diagnostics.

Once again, what we need is the navigation function, not the real management Diagnostics

justinmk:
> I have a use case. I set up the following mappings to navigate errors,

Mappings were also the use-case for https://github.com/vscode-neovim/vscode-neovim/pull/2045 . The question is whether it is worth the risk of adding a lot of code, for a few mappings. 

A future version of vscode may gain the behavior you want, but then it's unlikely that we will know when we can untangle all of this extra code.

xiyaowong:
> A future version of vscode may gain the behavior you want, but then it's unlikely that we will know when we can untangle all of this extra code.

I didn't look carefully before; these commands have been available for a long time.
`editor.action.marker.next` `editor.action.marker.prev`

ollien:
> > A future version of vscode may gain the behavior you want, but then it's unlikely that we will know when we can untangle all of this extra code.
> 
> I didn't look carefully before; these commands have been available for a long time. `editor.action.marker.next` `editor.action.marker.prev`

We've discussed this previously (though I admit I can't find the conversation). Unfortunately this doesn't let you navigate by severity

xiyaowong:
I don't have this need. However, I have no objections to this PR. It's just a habitual update of outdated content.

PR #2022, opened 2024-05-28
feat: move output messages to a separate channel

### LINKED ISSUE #1965: When the log level is error, normal messages cannot be output
image

1. Set output channel level to error
2. Run `lua print('a\nb')`

**This is not a log, so log levels should not affect it.**

Messages should use an independent output channel
https://github.com/vscode-neovim/vscode-neovim/blob/b9ad120efac084de7e66207a385ac66af97132d9/src/messages_manager.ts#L9

And it should be `OutputChannel`, not `LogOutputChannel`

ollien:
Fixes #1965. 

I added a second channel, `vscode-neovim (messages)`, and renamed the existing one to `vscode-neovim (logs)`. Messages are sent to both locations, which preserves timestamps. While right now the `vscode-neovim (messages)` channel is a bit barebones, I'm hoping to do some tweaks to make it more useful. Stay tuned on that :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ npm run build:dev
 
 You can view the extension logs in one of three locations
 
-1. Via the `vscode-neovim` Output channel
+1. Via the `vscode-neovim (logs)` Output channel
 
     - Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
       [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ npm run build:dev
 
 You can view the extension logs in one of three locations
 
-1. Via the `vscode-neovim (logs)` Output channel
+1. Via the `vscode-neovim logs` Output channel
 
     - Note: some messages are not logged to the Output channel, to avoid infinite loop. This is decided by the
       [`logToOutputChannel` parameter](https://github.com/vscode-neovim/vscode-neovim/blob/7337ffd5009067d074af5371171f277cb522aa9b/src/logger.ts#L184).
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,7 +28,7 @@ export async function activate(context: vscode.ExtensionContext, isRestart = fal
     }
 
     const logOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (logs)`, { log: true });
-    const userMessageOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (messages)`);
+    const messageOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (messages)`);
     disposables.push(logOutputChannel);
 
     config.init();
@@ -46,7 +46,7 @@ export async function activate(context: vscode.ExtensionContext, isRestart = fal
     try {
         const plugin = new MainController(context);
         context.subscriptions.push(plugin);
-        await plugin.init(userMessageOutputChannel);
+        await plugin.init(messageOutputChannel);
     } catch (e) {
         vscode.window
             .showErrorMessage(`[Failed to start nvim] ${e instanceof Error ? e.message : e}`, "Restart")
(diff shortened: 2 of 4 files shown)


src/messages_manager.ts
```diff
@@ -54,10 +58,16 @@ export class MessagesManager implements Disposable {
...
                 this.channel.show(true);
-                this.channel.append(str);
+                this.writeMessage(str);
                 break;
             }
         }
     }
+
+    private writeMessage(msg: string) {
+        logger.info(msg);
```
xiyaowong, line 70: Do we need to add it to the log?

src/extension.ts
```diff
@@ -27,11 +27,12 @@ export async function activate(context: vscode.ExtensionContext, isRestart = fal
         verifyExperimentalAffinity();
     }
 
-    const outputChannel = vscode.window.createOutputChannel(EXT_NAME, { log: true });
-    disposables.push(outputChannel);
+    const logOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (logs)`, { log: true });
+    const userMessageOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (messages)`);
```
xiyaowong, line 31: Personal hobby: Can it be changed to `messageOutputChannel`?

src/messages_manager.ts
ollien: I waffled on this. My take is we should, as if we're ever asking a user for logs to diagnose a problem, these messages may be indicative of actions they took

src/extension.ts
ollien: I don't feel strongly. If you do, I'll change it. My thought was it was a bit clearer to say "user" because all output channels display messages...

src/messages_manager.ts
ollien: It also resolves the concern about preserving timestamps

xiyaowong: OK

src/extension.ts
xiyaowong: Nvim messages aren't necessarily user-initiated; they could come from plugin printouts. Both `log` and `message` clearly indicate their respective purposes.

ollien: Eh on second thought, I think yours is better

src/extension.ts
```diff
@@ -27,11 +27,12 @@ export async function activate(context: vscode.ExtensionContext, isRestart = fal
         verifyExperimentalAffinity();
     }
 
-    const outputChannel = vscode.window.createOutputChannel(EXT_NAME, { log: true });
-    disposables.push(outputChannel);
+    const logOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (logs)`, { log: true });
+    const messageOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} (messages)`);
```
justinmk, line 31:
```suggestion
    const logOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} logs`, { log: true });
    const messageOutputChannel = vscode.window.createOutputChannel(`${EXT_NAME} messages`);
```

CONTRIBUTING.md
```diff
@@ -47,7 +47,7 @@ npm run build:dev
 
 You can view the extension logs in one of three locations
 
-1. Via the `vscode-neovim` Output channel
+1. Via the `vscode-neovim (logs)` Output channel
```
justinmk, line 50:
```suggestion
1. Via the `vscode-neovim logs` Output channel
```

justinmk (approved):
I think having a single place to go to and refer to in discussions/docs, is very helpful for discoverability and usability. But it looks like no one else agrees :)

ollien:
>  I think having a single place to go to and refer to in discussions/docs, is very helpful for discoverability and usability. But it looks like no one else agrees :)

I actually don't disagree, but given the limitations of `LogOutputChannel`, this doesn't seem like the worst tradeoff.

xiyaowong:
> I think having a single place to go to and refer to in discussions/docs, is very helpful for discoverability and usability. But it looks like no one else agrees :)

What's this mean?

xiyaowong (approved):

justinmk:
> What's this mean?

I mean one channel instead of multiple channels. But it applies more generally (not only to channels): having one thing is usually less confusing (results in less docs, less qualifiers in discussions, etc) than spreading things out.

PR #2024, opened 2024-05-28
feat: revamp messages output channel

ollien:
This PR builds on #2022 ~(sorry for the redundant commits... couldn't find a good way to stack the PRs).~

The idea here is that the output channel can be used to emulate the bottom split window in nvim after calling `messages`, or a statusline. 

A video's worth a thousand words: 

https://github.com/vscode-neovim/vscode-neovim/assets/977151/de2b7376-1de5-4dc6-b8c5-c3416b3e6756

A few notes on what you're watching here

1. If the output panel is open, a `msg_show` will be the only thing shown
2. If the output panel is closed, single-line output will not open the panel
3. multi-line output or command history will show the output panel 

Of course, you can still view historical outputs in the logs channel.
image

### DIFF
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -67,7 +67,11 @@ export class MessagesManager implements Disposable {
     }
 
     private writeMessage(msg: string) {
+        if (msg.length === 0) {
+            return;
+        }
+
         logger.info(msg);
-        this.channel.append(msg);
+        this.channel.replace(msg);
     }
 }
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -54,7 +54,11 @@ export class MessagesManager implements Disposable {
                         for (const [commandName, content] of list) {
                             const cmdContent = content.map((c) => c[1]).join("");
 
-                            lines.push(`${commandName}: ${cmdContent}`);
+                            if (commandName.length === 0) {
+                                lines.push(cmdContent);
+                            } else {
+                                lines.push(`${commandName}: ${cmdContent}`);
+                            }
                         }
                     }
                 }
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -76,11 +76,11 @@ export class MessagesManager implements Disposable {
         }
 
         logger.info(msg);
-        const outputMsg = this.endInNewline(msg);
+        const outputMsg = this.ensureEOL(msg);
         this.channel.replace(outputMsg);
     }
 
-    private endInNewline(msg: string): string {
+    private ensureEOL(msg: string): string {
         if (msg.length === 0 || msg[msg.length - 1] === "\n") {
             return msg;
         }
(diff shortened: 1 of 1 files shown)


src/messages_manager.ts
```diff
@@ -20,54 +20,71 @@ export class MessagesManager implements Disposable {
...
+            return;
+        }
+
         logger.info(msg);
-        this.channel.append(msg);
+        const outputMsg = this.endInNewline(msg);
+        this.channel.replace(outputMsg);
+    }
+
+    private endInNewline(msg: string): string {
```
justinmk, line 83: nit: 
```suggestion
    private ensureEol(msg: string): string {
```

justinmk (approved):
no objection as long as the "logs" channel didn't lose any info it was getting before

xiyaowong (approved):

ollien:
> TODO: We should handle "msg_history_clear" event

We can open an issue to discuss this further, but is there anything we need to do here? The `:help ui` page seems to imply it just is clearing the history `:messages` uses, but I don't see anything about what the UI should do

```
["msg_history_clear"]
Clear the [:messages](https://neovim.io/doc/user/message.html#%3Amessages) history.
```

ollien:
> Just a suggestion, could we simplify "vscode-neovim messages" to "neovim messages"?

I think keeping the full name of the extension is best for clarity

justinmk:
Yes. Also it's possible we would want to put non-Nvim messages in there? Though that may be redundant with vscode's builtin notifications history.

xiyaowong:
I think we can simply `this.channel.clear()`.

ollien:
> I think we can simply `this.channel.clear()`.

I did some quick testing. It looks like calling `:messages` sends a `msg_history_clear` immediately after `msg_history_show`, so calling `clear()` when we get such an event actually just hides all output immediately. I'm not sure that it's worth the complexity to try and differentiate between standalone clears and ones following `msg_history_show`.

wyyea:
I find this feature sometimes annoying especially when you are coding and the panel suddenly pop up. Is it possible to set an option to switch on/off this feature? Thanks a lot.

ollien:
@wyyea There and been a lot of ink spilled about a config option for this. I can't find all the conversations at this moment, but https://github.com/vscode-neovim/vscode-neovim/issues/2046 has some background. 

Ultimately, if nvim needs to display output that isn't just a simple status line, it needs to go somewhere for you to see. We're reluctant to add an option for this, as a result. However, if single-line outputs are forcing the panel to show, then that may be a bug.

In any case, please open an issue about this to continue this discussion. Comments on old PRs are rather invisible to users and maintainers alike.

justinmk:
> when you are coding and the panel suddenly pop up.

This is completely useless info without saying the steps that caused the panel to show. If you invoked a command (indirectly) that prints output, why in the world should the output not display?

PR #2026, opened 2024-05-28
chore: remove redundant error message on disconnect

xiyaowong:
logger.error also generates error messages, which seems a bit redundant (previously, this confused me as to why there were two error messages). Moreover, this logging is not useful in the logs.


justinmk:
This is a fix. Another case where "chore" is being overused.

xiyaowong:
👀

justinmk (approved):

Issue #521, opened 2021-01-09
Allow write to send buffer to external program


jrpear:
In CL Neovim you can use `:w !command` to send the contents of the current buffer to the stdin of any external program `command`. I'd like to be able to do this in VSCode Neovim.

I'm not sure exactly how this is done, but it could be as easy as checking whether a `:w` command has arguments before modifying it to `:Write`

asvetliakov:
The right path would be add handling here:

https://github.com/asvetliakov/vscode-neovim/blob/532d4b8809f68fd90b9f4c18484a66ec158441d3/vim/vscode-file-commands.vim#L34

ollien:
Should be fixed by #2027

justinmk:
Related: https://github.com/vscode-neovim/vscode-neovim/issues/1260#issuecomment-1637179297

theol0403:
Fixed by https://github.com/vscode-neovim/vscode-neovim/issues/2116

PR #2027, opened 2024-05-29
refactor: remove Altercmd for "write"

ollien:
~~This is definitely _not_ fully tested, but is something I was playing around with~~ I wanted to throw out this there to see if there was any something I hadn't considered here :) It's a pretty fundamental change, so I'd love eyes from those who know more than me :D

Should also address #521. I was able to run `w !wc -c` on an unsaved buffer and get the output, without kicking up a save dialog

I've been running this for a bit with no major issues, but would love some soak testing help

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -614,8 +614,8 @@ export class BufferManager implements Disposable {
             vim.api.nvim_buf_set_var(bufId, "vscode_uri_data", docUriJson)
             vim.api.nvim_buf_set_name(bufId, bufname)
             vim.api.nvim_buf_set_option(bufId, "modifiable", not isExternalDoc)
-            -- force nofile, just in case if the buffer was created externally
-            vim.api.nvim_buf_set_option(bufId, "buftype", "nofile")
+            -- force acwrite, which is similar to nofile, but will only be written via the BufWriteCmd autocommand.
+            vim.api.nvim_buf_set_option(bufId, "buftype", "acwrite")
             vim.api.nvim_buf_set_option(bufId, "buflisted", true)
         `,
             [
--- a/src/test/integ/vscode-integration-specific.test.ts
+++ b/src/test/integ/vscode-integration-specific.test.ts
@@ -1,6 +1,6 @@
 import path from "path";
 import os from "os";
-import fs from "fs";
+import fs from "fs/promises";
 import { strict as assert } from "assert";
 
 import vscode, { Selection } from "vscode";
@@ -338,7 +338,7 @@ describe("VSCode integration specific stuff", () => {
 
     it("Opens a file through e command", async () => {
         const filePath = path.join(os.tmpdir(), Math.random().toString());
-        fs.writeFileSync(filePath, ["line 1"].join("\n"), {
+        await fs.writeFile(filePath, ["line 1"].join("\n"), {
             encoding: "utf8",
         });
 
@@ -353,7 +353,6 @@ describe("VSCode integration specific stuff", () => {
             client,
         );
     });
-
     it("Spawning command line from visual line mode produces vscode selection", async () => {
         await openTextDocument({ content: ["a1", "b1", "c1"].join("\n") });
         await sendVSCodeKeys("Vj");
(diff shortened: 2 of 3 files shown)


xiyaowong:
Maybe put the autocmd where the altercmd was?

ollien:
I can, I just put it with the other `autocmds`

xiyaowong:
Testing can be simplified. Just check the dirty flag.

ollien:
> Testing can be simplified. Just check the dirty flag.

Checking the dirty flag doesn't tell us very much, since it's explicitly set with `set nomod`, right? Checking the file on disk gives us assurance it was actually written

xiyaowong:
`TextDocument.isDirty`

ollien:
> `TextDocument.isDirty`

Sure, we could check that, but it's an integration test, is there any reason we shouldn't be checking the file system? We already use it for a few things, so we can ensure the write is working as expected.

xiyaowong:
Because we are calling the vscode command, if there is a problem, it is  vscode's problem.

ollien:
I really disagree, I think that it's completely reasonable to check the filesystem for such a critical action. It may be VSCode's problem, but we'd still want to know about it. 

Regardless: we don't need to harp on this one aspect for now :) The real thing I'm concerned about is the implications at runtime, beyond our tests.

xiyaowong:
> It may be VSCode's problem, but we'd still want to know about it.

0 chance

src/test/integ/vscode-integration-specific.test.ts
```diff
@@ -354,6 +354,42 @@ describe("VSCode integration specific stuff", () => {
...
+        const filePath = path.join(os.tmpdir(), Math.random().toString());
+        await fs.writeFile(filePath, "hello, world!\n", {
+            encoding: "utf8",
+        });
+
+        await sendVSCodeCommand("workbench.action.toggleAutoSave");
+        await openTextDocument(filePath);
+        await sendVSCodeKeys("ct,heyo");
+        await sendEscapeKey();
+        await sendVSCodeKeys(":w ! wc -c");
```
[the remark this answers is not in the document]
xiyaowong: Avoid unnecessary delay.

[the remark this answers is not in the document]
ollien: What's the difference? I don't mind changing it, but there seem to be several ways we can input things, and I'm having trouble understanding the difference between them

src/buffer_manager.ts
```diff
@@ -615,7 +615,7 @@ export class BufferManager implements Disposable {
             vim.api.nvim_buf_set_name(bufId, bufname)
             vim.api.nvim_buf_set_option(bufId, "modifiable", not isExternalDoc)
             -- force nofile, just in case if the buffer was created externally
-            vim.api.nvim_buf_set_option(bufId, "buftype", "nofile")
+            vim.api.nvim_buf_set_option(bufId, "buftype", "acwrite")
```
xiyaowong, line 618: Perhaps add some comments here, as `acwrite` is not commonly used, and this PR adds code that depends on this option.

src/test/integ/vscode-integration-specific.test.ts
```diff
@@ -354,6 +354,40 @@ describe("VSCode integration specific stuff", () => {
         );
     });
 
+    it("Writes changes to file through w command", async () => {
+        const filePath = path.join(os.tmpdir(), Math.random().toString());
+        await fs.writeFile(filePath, "hello, world!\n", {
+            encoding: "utf8",
+        });
+
+        await sendVSCodeCommand("workbench.action.toggleAutoSave");
```
xiyaowong, line 363: What's this purpose?

ollien: I probably should have added a comment, but auto save would sometimes kick in and subvert the point of the test. The idea was to disable auto save to prevent that

xiyaowong: VSCode does not enable auto-save by default. Additionally, you should modify the configuration to set 'files.autoSave' to 'off'.

src/buffer_manager.ts
xiyaowong: Additionally, the original comments are outdated.

src/test/integ/vscode-integration-specific.test.ts
ollien: Weird... I wonder why this fixed things then.

Is there a way to change the configuration just for the test host? I looked for a while and couldn't find a nice way, but I might be missing something obvious

xiyaowong: https://github.com/vscode-neovim/vscode-neovim/blob/09571212c176225ea3eca4f351296ed464b804b5/src/test/integ/composite-escape.test.ts#L18-L43

xiyaowong (approved):
Just a suggestion: change the title to `refactor: remove altercmd for "write"`

theol0403 (approved):
This is a step in the right direction, but not really the solution we need.

Ideally, we connect the vscode API with the nvim BufRead/BufWrite commands, so that file management is truly done natively (and we don't need any altercommands for files). 

Here is some playing I did https://github.com/theol0403/vscode-neovim/commit/629f6b166c92b9f5004532053be407b79948039a

Note that the BufWriteCmd contains all the information about the file being saved... so we can literally go and tell vscode to save *that* file. Then, `:w newname` and `:wall` just work. 

However, I ran into some roadblocks: how to associate a buffer with a URI? or saveas with a new URI? How to avoid saving `Untitled` files with `:wall`, when nvim triggers a save for every file individually? Do we need to remove the `w!` functionality for saveas, because there is no nvim equivalent?

We can slowly peel off the bandaid with this PR but it's not really the solution we need.

ollien:
> Note that the BufWriteCmd contains all the information about the file being saved... so we can literally go and tell vscode to save that file. Then, :w newname and :wall just work.

Oh cool! I was looking for commands to do this but I guess I can just use the JS API. I'll take a look at that commit when I get a spare cycle. 

I do think this PR has some value on its own, if only because it fixes #521, so I'll merge if it's ok with you.

theol0403:
A thought that came to me is imagine a user (say, with LazyVim) maps  `C-s` to `:wall`. Before this PR, this would have no effect, because the plumbing isn't there. Now, `files.save` is called once for every buffer you have open (10s of times).

You should add a robust check to only trigger a save if the file being saved is the active buffer, because we've changed the behavior from "when the user hits save" to "when any file is saved".

ollien:
That's a good point. I'll revert this PR until I can give this focused attention. Thanks

theol0403:
For now we just won't release.

ollien:
Sounds good. Sorry to have screwed up the release notes :sweat_smile: . Looks like release-please doesn't understand reverts

xiyaowong:
You can temporarily disable branch protection, then force push, and rerun the release-please workflow.

ollien:
@xiyaowong I don't have the permissions to do that. If you would be so kind, that would be great

justinmk:
Great to see movement in this direction, even if it was reverted for the time being :)

PR #2029, opened 2024-05-29
feat: show output when return_prompt is sent

ollien:
Just a small QoL change in face of #2024. If nvim is sending us a `return_prompt` event, the output we have is probably meaningful enough to show to a user, so we should pop out the output channel. This is useful for things like `!`, which might only have one line of output (and thus wouldn't normally display)

### DIFF
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -23,6 +23,7 @@ export class MessagesManager implements Disposable {
                 const lines: string[] = [];
                 for (const [type, content, clear] of args) {
                     if (type === "return_prompt") {
+                        this.channel.show(true);
                         continue;
                     }
 


xiyaowong:
Will it be frequent?
I personally hate extensions that automatically open the output panel.

ollien:
> Will it be frequent?

It will only be as frequent as the commands you execute would show a `Press ENTER or type command to continue` message in nvim :). I actually came across this when testing #2027, because `w ! wc -c` didn't display output. 

That said, I would be open to making a setting like "autoShowOutput", which would disable all auto-output-popping from `messages_manager`. What do you think?

xiyaowong:
> It will only be as frequent as the commands you execute would show a `Press ENTER or type command to continue` message in nvim :). 

This is fine.

> That said, I would be open to making a setting like "autoShowOutput", which would disable all auto-output-popping from `MessagesManager`. What do you think?

No, It's unnecessary

xiyaowong (approved):

Issue #2028, opened 2024-05-29
bug: cmdline commands through keymaps execute before complete


danielaird27:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

NVIM v0.9.4

### Operating system/version

Windows 11

### Describe the bug

It seems like some cmdline commands setup through keymaps execute too early and end up typing the remainder of the command into the text. I can confirm the issue is present with one keymap in my nvim config:

```
vim.keymap.set('v', '/', 'y/\\V<C-R>=escape(@",\'/\')<CR><CR>')
```

When I highlight a symbol and hit the "/" key, the "cape(@",'/')" of the keymap's command gets entered into the current buffer. It doesn't seem to happen all the time but it happens most of the time.

Highlighting the symbol before hitting the "/" key:
![Screenshot 2024-05-29 104242](https://github.com/vscode-neovim/vscode-neovim/assets/57294055/1d782947-0a45-4f67-9ca9-7a5164339cc7)

After hitting the "/" key:
![Screenshot 2024-05-29 104257](https://github.com/vscode-neovim/vscode-neovim/assets/57294055/6b5f0af4-c196-4cd8-980a-1d18b35c505a)

This problem is present in vscode-neovim 1.13.1 and 1.13.0 but is not present in 1.12.0.

Note: I could not run with vscode-neovim.neovimClean=true to test this because that would remove the keymap. However, the fact that using vscode-neovim 1.12.0 makes this problem go away leads me to believe the problem isn't in my very light config.

### Steps To Reproduce

1. Add ```vim.keymap.set('v', '/', 'y/\\V<C-R>=escape(@",\'/\')<CR><CR>')``` keymap to your nvim config.
2. Open a buffer with and fill it with the Python example given in the description.
3. Highlight funcA and hit the "/" key.

### Expected Behavior

I expect there to be no text entered into my buffer when starting a search with the provided keymap. That is the functionality shown in vscode-neovim 1.12.0.

ollien:
Thanks for the report. Indeed, I can confirm this seems to be a regression from #2001. Will investigate further.

ollien:
[...]

PR #2030, opened 2024-05-30
fix(cmdline): support nested cmdline

theol0403:
Fix https://github.com/vscode-neovim/vscode-neovim/issues/2028

`ignoreHideEvent` wasn't actually having any effect. But, when it did, it would cause issues with opening and closing the cmdline, because `cmdline_hide` would trigger after the cmdline is already hidden, setting `ignoreHideEvent` to true, and then not properly exiting the next cmdline.

This adds another state variable to make sure we know exactly when `onHide` will actually trigger (so we can be guaranteed that it clears the flag). 

Finally, the macro issue in #2028 might only be solvable with a timer (like in the old impl).

### DIFF
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -103,6 +103,9 @@ export class CommandLineManager implements Disposable {
             }
             case "cmdline_hide": {
                 logger.debug(`cmdline_hide`);
+                // in the case of rapid cmdline_hide and cmdline_show, we want to rush the behavior of onHide
+                this.input.value = "";
+                this.state.redrawExpected = true;
                 if (this.state.isDisplayed) {
                     this.state.ignoreHideEvent = true;
                     this.input.hide();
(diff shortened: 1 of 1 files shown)


ollien:
Nice! Do you think there's a way you can add an integration test around this? Maybe insert some text as a result of an evaluated expression?

theol0403:
Would be happy to but unfortunately am working on this extension on borrowed time so might need your help.

ollien:
I can give it a shot in the morning :)

theol0403:
All bugs should be fixed, please test!

There is a little bit of visual noise with the cmdline flashing open for a split second as the macro runs, but there are no more conflicts. 

This could be solved in a later PR where the cmdline showing is deferred by 10ms to allow for macros. However, this adds to the complexity with timing edge cases.

ollien:
> Finally, the macro issue in https://github.com/vscode-neovim/vscode-neovim/issues/2028 might only be solvable with a timer (like in the old impl).

Not at a computer right now to try, but could we somehow make use of the `level` argument from the `cmdline_show` event?

theol0403:
Different issue, but I didn't consider using that. It might simplify some code? Really, the quickpick should be a "UI view" over neovim, so really shouldn't care about the layer (now you are syncing states). Things should be good as is. 

My message you quoted was regarding the race conditions when executing the macro, and wasn't anything to do with the nested cmdlines. I thought a timer would be the solution, but I was able to remove the race condition without one.

theol0403:
@danielaird27 try with https://github.com/vscode-neovim/vscode-neovim/actions/runs/9295456699/artifacts/1550666108

theol0403:
Not sure if the failing tests are my fault but not enough time to investigate.

ollien:
> Not sure if the failing tests are my fault but not enough time to investigate.

I think it is due to this PR, but it's tricky. Once the help page is opened, a `<C-u>` is being sent to the main buffer (the extension assumes the cmdline is open), which gets sent to the `:help` buffer, scrolling it up

Issue #2031, opened 2024-05-30
bug: `StatusLineManager` causes spamming of `cmdline_show` events


theol0403:
When debugging the command line manager, a steady stream of debug logs are triggered due to the `StatusLineManager` poll loop. Commenting it out in `MainManager` solves the problem. This should be avoided.

PR #2032, opened 2024-05-30
fix: pulling status causes a redraw

xiyaowong:
- fix: #2031

### DIFF
--- a/src/status_line_manager.ts
+++ b/src/status_line_manager.ts
@@ -9,6 +9,7 @@ enum StatusType {
     Mode, // msg_showmode
     Cmd, // msg_showcmd
     Msg, // msg_show, msg_clear
+    StatusLine, // (custom) statusline
 }
 
 export class StatusLineManager implements Disposable {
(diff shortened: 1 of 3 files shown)


xiyaowong:
@ollien Sorry for the wrong info https://github.com/vscode-neovim/vscode-neovim/pull/1989#issuecomment-2119578688

src/status_line_manager.ts
```diff
@@ -19,9 +20,9 @@ export class StatusLineManager implements Disposable {
     private _cmdText = "";
     private _msgText = "";
 
-    private statusBar: StatusBarItem;
+    private _statusline = "";
```
ollien, line 23: nit: we don't have a getter for this, so why use the underscore prefix?

ollien (approved):

src/status_line_manager.ts
xiyaowong: Just something left from before.

PR #2033, opened 2024-05-30
fix: correctly dispose the output channels

xiyaowong:
When you execute the `vscode-neovim.restart` command, duplicate output channels appear. 
image


ollien (approved):
Nice spot!

PR #2037, opened 2024-05-30
fix: Correctly set doc version for skipping changes

### LINKED ISSUE #1624: bug: VSCode and Nvim buffers get out of sync + general unwanted behavior

- [ ] I can reproduce the issue with an empty init.vim/init.lua and no other vscode extensions (when applicable)

### Neovim version (nvim -v)

0.9.0

### Operating system/version

Windows 10 - 22H2

### Describe the bug

After the update to VSCode Neovim v1.0.1, the buffers of neovim and vscode regularly get out of sync. I think it also happens through stuff like auto formatting, but neovim also manages to do so all by itself.

The remap I added was for a purpose I don't remember right now, I think it had to do with cursor positions after using 'o' or so... but it worked in the last release of the extension, so I'll use it as an example.

### Steps To Reproduce

1. Add remap to otherwise empty init.lua: `vim.keymap.set({'n'}, 'o', 'o.<backspace>', {noremap=true})`
2. Open an empty file, press 'o' (the just defined shortcut) to go into insert mode on a new line below, type 2 lines: "asdf" and "jkl;"
3. press <ctrl+h> (while still in insert mode) to delete the last character

### Expected Behavior

I expect the file to look like this afterwards (with * showing the current cursor position):
```

asdf
jkl*
```
I should have 3 lines and the last character typed, the ';', should be deleted. The cursor should be at the end of the file.

Instead, the file looks like this:
```
*
jkl;
```
There are only 2 lines. The cursor is at the start of the file, the first typed line is removed completely. The second line is still there completely.
Additionally, the cursor won't move with hjkl commands because nvim thinks the file is empty. If I click anywhere on the second line, I get the error `nvim_win_set_cursor: Cursor position outside buffer`

### LINKED ISSUE #2023: bug: insertions after `S` become immutable
### Check the following:

### Neovim version (nvim -v)

0.10.0

### Operating system/version

MacOS 13.6.7

### Describe the bug

Some key sequences could lead to unexpected behavior when editing.

https://github.com/vscode-neovim/vscode-neovim/assets/43830077/b38a1d66-0727-4e54-a9c6-679b9f175f22

### Steps To Reproduce

1. random text
2. normal mode press O (capital o)
3. esc back to normal mode
4. press S (capital s) to insert
5. type `{`
6. esc back to normal mode

Now, the `{}` cannot handle any operations (editing and cursor moving).

### Expected Behavior

The inserted characters should be operatable.

xiyaowong:
- fix #2023
- fix #1624 (Maybe)

### DIFF
--- a/src/document_change_manager.ts
+++ b/src/document_change_manager.ts
@@ -257,9 +257,11 @@ export class DocumentChangeManager implements Disposable {
                     }
                     const oldText = doc.getText().replace(/\r\n/g, "\n");
                     const newText = newLines.join("\n");
-                    this.documentSkipVersionOnChange.set(doc, doc.version + 1);
 
                     const cursorBefore = editor.selection.active;
+
+                    // 1. Manually increment document version to avoid unexpected updates
+                    this.documentSkipVersionOnChange.set(doc, doc.version + 1);
                     const success = await editor.edit(
                         (builder) => {
                             const changes = calcDiffWithPosition(oldText, newText);
@@ -269,6 +271,11 @@ export class DocumentChangeManager implements Disposable {
                         },
                         { undoStopAfter: false, undoStopBefore: false },
                     );
+                    // 2. Set the actual version number after applying changes to prevent loss of the next update
+                    // Although successful, the document version may not actually increase.
+                    // Example: Using "S" to delete empty line does not actually change the text.
+                    this.documentSkipVersionOnChange.set(doc, doc.version);
+
                     const docPromises = this.textDocumentChangePromise.get(doc)?.splice(0, lastPromiseIdx) || [];
                     if (success) {
                         if (!editor.selection.anchor.isEqual(editor.selection.active)) {
--- a/src/test/integ/basic-editing-navigation.test.ts
+++ b/src/test/integ/basic-editing-navigation.test.ts
@@ -462,4 +462,19 @@ describe("Basic editing and navigation", () => {
         );
         assert.ok(editor.document.eol === EndOfLine.CRLF);
     });
+
+    it("Correctly set document version for skipping updates", async () => {
+        await openTextDocument({ content: "" });
+
+        await sendVSCodeKeys("S");
+        await sendVSCodeKeys("test");
+        await sendEscapeKey();
+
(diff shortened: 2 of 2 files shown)


src/document_change_manager.ts
```diff
@@ -269,6 +269,9 @@ export class DocumentChangeManager implements Disposable {
                         },
                         { undoStopAfter: false, undoStopBefore: false },
                     );
+                    // Although successful, the document version may not actually increase.
+                    // Example: Using "S" to delete empty line does not actually change the text.
+                    this.documentSkipVersionOnChange.set(doc, doc.version);
```
ollien, line 274: I haven't looked at this cod before, so forgive the silly question, but is there any reason to do the `set` on like 260, still?

xiyaowong: It's indeed a question; I am Not sure what the event-triggering cycle of VSCode is like.

xiyaowong:
Changed. It's more logical to save the version after applying the changes.

src/document_change_manager.ts
ollien: Hm, upon reflection, I guess the event handler for a document being changed could be fired before we add this to the map. We'd need some kind of mutual exclusion here, I think.

xiyaowong: > We'd need some kind of mutual exclusion here

It seems a bit tricky. How do we handle the issue of outdated data? We need to maintain the sequence of events triggered by two completely different operations. Storing the wrong version number here could result in losing changes made next time in VS Code. Correcting the version number after applying changes should work.

ollien: It does. Though I do think there's some potential for races regardless. I think it would make some sense to have a lock shared between both syncs, so that only one sync direction is handled at a time. That said, I don't think its necessary for this PR per-se.

ollien (approved):

src/document_change_manager.ts
xiyaowong: Using locks seems to disturb the event sequence instead. 

If both directions simultaneously undergo different changes, then the content is directly wrong.

PR #2041, opened 2024-05-31
fix: detect potential module conflicts

xiyaowong:


ollien (approved):

ollien:
Could we stand to add this limitation to the docs somewhere?

xiyaowong:
The documentation clearly states `local vscode = require("vscode")`. New users should understand this, but it may pose an issue for existing users.

ollien:
I don't know that I would take away from that blurb that having a file called `lua/vscode.lua` in my configuration directory would pose a problem.

xiyaowong:
I think it's enough to just detect errors. It seems a bit redundant to emphasize not using the same module name while already informing the user that the module name is vscode. Also, no plugin does it this way.

PR #2045, opened 2024-05-31
feat: override vim.lsp.buf methods

### LINKED ISSUE #2044: feature: override vim.lsp.buf methods
These methods are very common and basically the same as the behavior in vscode. My idea is to directly override these methods to call the corresponding commands in vscode.

xiyaowong:
- Resolve #2044

### DIFF
--- a/runtime/vscode/lsp/buf.lua
+++ b/runtime/vscode/lsp/buf.lua
@@ -30,7 +30,9 @@ local map = {
     end
     vscode.action(cmd)
   end,
-  list_workspace_folders = vim.NIL,
+  list_workspace_folders = function()
+    return vscode.eval("return (vscode.workspace.workspaceFolders || []).map((folder) => folder.name);")
+  end,
   add_workspace_folder = "workbench.action.addRootFolder",
   remove_workspace_folder = "workbench.action.removeRootFolder",
   workspace_symbol = function(query)
--- a/runtime/vscode/lsp/buf.lua
+++ b/runtime/vscode/lsp/buf.lua
@@ -1,3 +1,5 @@
+--- Shim vim.lsp.buf calls to vscode equivalent commands.
+
 local vscode = require("vscode")
 
 local M = {}
@@ -45,7 +47,7 @@ for method, cmd in pairs(map) do
 
   if cmd == vim.NIL then
     M[method] = function()
-      print(string.format("vim.lsp.buf.%s is not supported in vscode.", method))
+      vscode.notify(string.format("vim.lsp.buf.%s is not supported in vscode.", method), vim.log.levels.WARN)
     end
   elseif cmd_type == "string" then
     M[method] = function()
@@ -59,12 +61,13 @@ end
 vim.lsp.buf = setmetatable(M, {
   __index = function(t, method)
     t[method] = function()
-      print(
+      vscode.notify(
         string.format(
           "vim.lsp.buf.%s is not handled by vscode-neovim. %s",
           method,
-          "You can start a discussion about this in vscode-neovim"
-        )
+          "Please report this issue at [vscode-neovim](https://github.com/vscode-neovim/vscode-neovim/issues)"
+        ),
+        vim.log.levels.WARN
       )
     end
     return t[method]
--- a/runtime/vscode/lsp/buf.lua
+++ b/runtime/vscode/lsp/buf.lua
@@ -41,15 +41,13 @@ local map = {
 }
 
 for method, cmd in pairs(map) do
+  local cmd_type = type(cmd)
+
   if cmd == vim.NIL then
     M[method] = function()
       print(string.format("vim.lsp.buf.%s is not supported in vscode.", method))
     end
-    return
-  end
-
-  local cmd_type = type(cmd)
-  if cmd_type == "string" then
+  elseif cmd_type == "string" then
     M[method] = function()
       vscode.action(cmd)
     end
--- a/runtime/vscode/lsp/buf.lua
+++ b/runtime/vscode/lsp/buf.lua
@@ -1,0 +1,74 @@
+local vscode = require("vscode")
+
+local M = {}
+
+local map = {
+  hover = "editor.action.showHover",
+  declaration = "editor.action.peekDeclaration",
+  definition = "editor.action.peekDefinition",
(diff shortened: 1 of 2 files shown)


theol0403 (approved):

runtime/vscode/lsp/buf.lua
justinmk, line 1: It's not in the `overrides/` directory. What again is the difference between vscode/ and vscode/overrides/ ?

justinmk:
What problem is this solving? https://github.com/vscode-neovim/vscode-neovim/issues/2044 explains nothing. Please, always explain the problem being solved. That saves time for reviewers and saves needless discussion/questions/assumptions.

Just for reference: will this interfere with users that expect Nvim LSP (or plugins like null-ls) to work? 

This implementation looks isolated and "flat", thus low-risk and low-maintenance. If it doesn't creep into a growing collection of workarounds spread around the codebase, then it's probably fine.

If this works well, then it seems like it could avoid unnecessary key mappings/bindings. Does it allow us to remove any existing keybindings? 

If this doesn't work well, then similar to https://github.com/vscode-neovim/vscode-neovim/pull/2020 , it's risky. 

When a project is "small", it's easy to add features, because the project is easy to work with. Eventually it gets hard to work with, and contributors fade away and then it's hard to figure out which parts are needed and which were optional. It seems like we're getting a lot of minor nice-to-have feature proposals with very weak justifications:

- this PR (gives no explanation about the motivation, nor does the linked issue)
- https://github.com/vscode-neovim/vscode-neovim/pull/2020
- several other recent PRs

xiyaowong:
Isn't it more convenient to set keymaps to these methods than to find command id by yourself?

runtime/vscode/lsp/buf.lua
xiyaowong: `vscode/overrides` was actually created because I didn't know where to put the Vimscript. Later, after these Vimscript files are improved or switched to Lua, they should be relocated. Additionally, `vscode/ui.lua` and `vscode/code_actions.lua` are also considered overrides.

xiyaowong: It does get a little weird here. If you feel strongly about it, I don't mind making changes.

theol0403:
@justinmk I understand your concerns. These are all PRs designed to smooth the edges and provide an out-of-the-box experience which is as compatible with nvim configs as possible. With this PR, all bindings related to LSP provided by a distro, say LazyVim, are now supported. The code is indeed isolated and flat. 

>  will this interfere with users that expect Nvim LSP (or plugins like null-ls) to work

Nvim LSP/null-ns will NEVER be supported by this extension by design. Thus, this PR acts as a substitute so that these bindings continue to work. 

For example, currently, a mapping like https://github.com/LazyVim/LazyVim/blob/8024201e75b471eac5e399dcb6c8ec4f3f55e2a2/lua/lazyvim/plugins/lsp/keymaps.lua#L24 is a no-op in VSCode. Now, it maps to the VSCode native command.

> we're getting a lot of minor nice-to-have feature proposals with very weak justifications

#2020 is indeed going the wrong direction, but it was a prototype. The correct implementation will look similar to this PR (simply forward the command to VSCode). 

In general, the whole mantra of this extension is "if this extension were perfect, there would be no need for a readme, since all nvim behavior works out of the box". Before, many behaviours like statusline, cmdline, and other things "several other recent PRs" have touched, required custom solutions/settings/bindings to function. Now, they work "the neovim way".

justinmk:
> Additionally, this PR (vim.lsp.buf) is simple enough.

That by itself is not really a great reason for any change to be made. There needs to be a clear problem that is being solved.

> The motivation for this https://github.com/vscode-neovim/vscode-neovim/pull/133 is the same, so no explanation is provided

I don't see how #133 is related, but anyways it saves time to clearly state what problem is being solved, instead of expecting others to puzzle together the context. That helps a lot for referencing the PR (or commit message, ideally) in the future, too.

> With this PR, all bindings related to LSP provided by a distro, say LazyVim, are now supported.

That's good, though it does imply that vscode-neovim intentionally wants to support "distros". That's a lot of scope to take on.

runtime/vscode/lsp/buf.lua
justinmk: should we rename vscode/overrides/ to vscode/vimscript/ ? (in a separate PR)

runtime/vscode/lsp/buf.lua
```diff
@@ -0,0 +1,74 @@
...
+end
+
+vim.lsp.buf = setmetatable(M, {
+  __index = function(t, method)
+    t[method] = function()
+      print(
+        string.format(
+          "vim.lsp.buf.%s is not handled by vscode-neovim. %s",
+          method,
+          "You can start a discussion about this in vscode-neovim"
```
justinmk, line 68: should this link to https://github.com/vscode-neovim/vscode-neovim/issues

runtime/vscode/lsp/buf.lua
```diff
@@ -0,0 +1,74 @@
+local vscode = require("vscode")
```
justinmk, line 1: 1 or 2 lines at the top of this file explaining its purpose ?

    -- Shim vim.lsp calls to vscode equivalent functions.

justinmk (approved):

theol0403:
@justinmk 

> So that implies that vscode-neovim intentionally wants to support "distros". That's a lot of scope to take on.

Not exactly... `vscode-neovim` wants to support the full nvim API, so that distros, users, and basically anyone can have nvim configs which don't need special cases for VSCode. In this case, many distros map bindings to LSP commands (show references, go to definition, toggle inlay hints, etc), but we are not *directly* supporting that distro... we are supporting the official nvim API.

justinmk:
> we are supporting the official nvim API.

Nvim UIs are not expected to shim the Nvim LSP interface just to avoid a few mappings. It's probably low risk for now, but may be counterproductive in the future, as Nvim LSP [becomes an extensibility point](https://github.com/neovim/neovim/pull/24338#issuecomment-1636048696) for more than just "LSP servers".

xiyaowong:
To be precise, vscode-neovim is not a Neovim UI, but rather more like a script host.

However, this kind of functionality can be implemented through a separate Neovim plugin. Installing the plugin has no cost for the user, and it also allows the user to have a clearer understanding of the available features.

PR #2052, opened 2024-06-01
feat(lua): Add `vscode.into_insert`

### LINKED ISSUE #1999: feature: expose `vscode.action_into_insert` and `neovim.action_into_insert`
There are some vscode bindings which create behaviors which are incompatible with nvim. However, we would like to intelligently ease the transition and allow clever behavior from normal mode.

The best example is vscode multi cursor like C-S-L for `editor.action.selectHighlights` or C-d for `editor.action.addSelectionToNextFindMatch`.

It would be nice to be able to do the following (equivalent, though I prefer the latter):

```lua
vim.keymap.set({ "n", "x", "i" }, "<C-S-L>", function() require("vscode-neovim").action_into_insert("editor.action.selectHighlights") end)
```

or

```json
{
  "key": "ctrl+shift+l",
  "command": "neovim.action_into_insert",
  "args": "editor.action.selectHighlights",
  "when": "editorFocus && neovim.init && neovim.mode != 'insert'"
}
```

This would remove the need for https://github.com/vscode-neovim/vscode-multi-cursor.nvim/blob/main/lua/vscode-multi-cursor/wrappers.lua, because we can make these bindings default for core (and generically handle any other binding which should move to insert mode before executing). 

https://github.com/vscode-neovim/vscode-neovim/assets/16546293/8952f29c-66f4-4e16-aec4-a677e654da90

xiyaowong:
Close #1999

- [x] implement
- [x] tests
- [x] doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,7 @@ local vscode = require('vscode')
    `:h g:clipboard` for more details. Usage: `vim.g.clipboard = vim.g.vscode_clipboard`
 10. `vscode.eval()`: evaluate javascript synchronously in vscode and return the result
 11. `vscode.eval_async()`: evaluate javascript asynchronously in vscode
+12. `vscode.into_insert()`: perform operations in insert mode or switch to insert mode.
 
 ### vscode.action(name, opts)
 
@@ -588,6 +589,29 @@ Parameters:
         -   `err` is the error message, if any
         -   `ret` is the result
         -   If no callback is provided, error will be shown as a VSCode notification.
+
+### vscode.into_insert(callback)
+
+Perform operations in insert mode or switch to insert mode. If in visual mode, this function will **preserve the
+selection** after switching to insert mode.
+
+Parameters:
+
+-   `callback` (function|nil): Optional callback function to run after switching to insert mode. If not provided, it
+    will directly switch to insert mode synchronously.
+
+Examples:
+
+```lua
+-- Make `editor.action.addSelectionToNextFindMatch` work correctly in any mode.
+vim.keymap.set({ "n", "x", "i" }, "<C-d>", function()
+  vscode.into_insert(function()
+    vscode.action("editor.action.addSelectionToNextFindMatch")
+  end)
+end)
+```
+
+![select-next](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/5a93c87e-626a-4a70-a9ef-5084f747c7ef)
 
 ### VimScript
 
--- a/runtime/lua/vscode/api.lua
+++ b/runtime/lua/vscode/api.lua
@@ -418,4 +418,94 @@ do
   end
 end
 
+-----------------------
+------ utilities ------
+-----------------------
+
+---Perform operations in insert mode or switch to insert mode.
+---If in visual mode, this function will preserve the selection after
+---switching to insert mode.
+---
+---Example:
+---Make `editor.action.addSelectionToNextFindMatch` work correctly in any mode.
+---This is the behavior of the default VSCode shortcut Ctrl+d: Add Selection To Next Find Match.
+---
+---```lua
+---vim.keymap.set({ "n", "x", "i" }, "<C-d>", function()
+---  vscode.into_insert(function()
+---    vscode.action("editor.action.addSelectionToNextFindMatch")
+---  end)
+---end)
+---````
+---
+---@param callback function|nil Optional callback function to run after switching to insert mode.
+---                             If not provided, it will directly switch to insert mode synchronously.
+function M.into_insert(callback)
+  vim.validate({ callback = { callback, "f", true } })
+
+  local mode = api.nvim_get_mode().mode
+
+  --- Insert ---
+
+  if mode == "i" then
+    if callback then
+      callback()
+    end
+    return
+  end
+
+  --- Normal ---
+
+  if mode == "n" then
+    api.nvim_feedkeys("a", "n", false)
+    if callback then
+      callback()
+    else
+      vim.wait(100, function() end)
+    end
+    return
+  end
+
+  --- Visual ---
+
+  if mode:match("[vV\x16]") then
+    local A = fn.getpos("v")
+    local B = fn.getpos(".")
+    local start_pos = { A[2], A[3] - 1 }
+    local end_pos = { B[2], B[3] - 1 }
+
+    if start_pos[1] > end_pos[1] or (start_pos[1] == end_pos[1] and start_pos[2] > end_pos[2]) then
+      start_pos, end_pos = end_pos, start_pos
+    end
+
+    if mode == "V" then
+      start_pos = { start_pos[1], 0 }
+      end_pos = { end_pos[1], #fn.getline(end_pos[1]) }
+    end
+
+    local range = vim.lsp.util.make_given_range_params(start_pos, end_pos, 0, "utf-16").range
+    local ranges = { range }
+
+    api.nvim_win_set_cursor(0, end_pos)
+    api.nvim_feedkeys(api.nvim_replace_termcodes("<Esc>a", true, true, true), "n", false)
+
+    if callback then
+      M.action("start-multiple-cursors", { args = { ranges }, callback = callback })
+    else
+      M.call("start-multiple-cursors", { args = { ranges } })
+    end
+
+    return
+  end
+
+  --- Other ---
+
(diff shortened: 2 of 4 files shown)


runtime/lua/vscode/api.lua
```diff
@@ -418,4 +418,94 @@ do
...
+    if callback then
+      callback()
+    end
+    return
+  end
+
+  --- Normal ---
+
+  if mode == "n" then
+    api.nvim_feedkeys("a", "n", false)
```
theol0403, line 460: Should this be `i` instead of `a`? Which one causes no cursor update (I forget)?

runtime/lua/vscode/api.lua
```diff
@@ -418,4 +418,94 @@ do
...
+    if callback then
+      callback()
+    else
+      vim.wait(100, function() end)
```
theol0403, line 464: What is the purpose of this?

runtime/lua/vscode/api.lua
```diff
@@ -418,4 +418,94 @@ do
...
+    local range = vim.lsp.util.make_given_range_params(start_pos, end_pos, 0, "utf-16").range
+    local ranges = { range }
+
+    api.nvim_win_set_cursor(0, end_pos)
+    api.nvim_feedkeys(api.nvim_replace_termcodes("<Esc>a", true, true, true), "n", false)
+
+    if callback then
+      M.action("start-multiple-cursors", { args = { ranges }, callback = callback })
+    else
+      M.call("start-multiple-cursors", { args = { ranges } })
```
theol0403, line 495: Instead of handling these different cases, can we just default callback to noop?

runtime/lua/vscode/api.lua
```diff
@@ -418,4 +418,94 @@ do
...
+    if start_pos[1] > end_pos[1] or (start_pos[1] == end_pos[1] and start_pos[2] > end_pos[2]) then
+      start_pos, end_pos = end_pos, start_pos
+    end
+
+    if mode == "V" then
+      start_pos = { start_pos[1], 0 }
+      end_pos = { end_pos[1], #fn.getline(end_pos[1]) }
+    end
+
+    local range = vim.lsp.util.make_given_range_params(start_pos, end_pos, 0, "utf-16").range
```
theol0403, line 486: Should probably split this up into util.visual_to_range

theol0403 (changes requested):
Clever implementation! We should also add the multi cursor bindings as default, but that can be in another PR if you want.

README.md
```diff
@@ -589,6 +590,29 @@ Parameters:
         -   `ret` is the result
         -   If no callback is provided, error will be shown as a VSCode notification.
 
+### vscode.into_insert(callback)
```
justinmk, line 593: `with_` is probably more conventional than `into_`.

```suggestion
### vscode.with_insert(callback)
```

justinmk:
I don't really understand why this is needed but whatever.

https://github.com/vscode-neovim/vscode-neovim/issues/1999 mentions : 

> This would remove the need for https://github.com/vscode-neovim/vscode-multi-cursor.nvim/blob/main/lua/vscode-multi-cursor/wrappers.lua,

Is that possible now? This PR doesn't delete any code.

runtime/lua/vscode/api.lua
xiyaowong: Please read the notes and documentation.

runtime/lua/vscode/api.lua
xiyaowong: Need to wait for vscode to complete the mode switch.

README.md
xiyaowong: I wanted to use "with" at first, but considering that this function can be used as an ordinary function call without providing a callback function. "With" is not appropriate?

runtime/lua/vscode/api.lua
xiyaowong: It's not necessary for now. Although we currently have several modules that need to parse the selected range, their logic differs.

xiyaowong:
> I don't really understand why this is worth adding, but whatever.
> 

The example added in this PR is the reason. 
Also https://github.com/vscode-neovim/vscode-neovim/issues/1351#issuecomment-1788637704

> Is that possible now? This PR doesn't delete any code.

This is another plugin, but some users might only need a part of the functionality of that plugin (wrapping commands for VSCode). In fact, the content of this PR was copied from that plugin and simply improved to make it a bit more general.

xiyaowong:
> Clever implementation! We should also add the multi cursor bindings as default, but that can be in another PR if you want.

I want to ask in advance, which shortcut key do you plan to bind it to? It seems like this is another case where someone hopes to meet their own needs by changing the default behavior. :eyes:

runtime/lua/vscode/api.lua
xiyaowong: You should use "a", you can check the first test

runtime/lua/vscode/api.lua
xiyaowong: Note that this function is designed so that providing a callback function is optional; it can be called directly.

runtime/lua/vscode/api.lua
theol0403: Then why do you not need to wait before running the callback?

runtime/lua/vscode/api.lua
theol0403: Functions should usually have one function. This one has two: convert the selection to a range, and then send that range to vscode. It should be a util for readability, maintainability, and so it is available for future maintainers.

runtime/lua/vscode/api.lua
theol0403: I know the parameter is optional. But we can reduce code by saying `if not callback then callback = function() end`. And I still don't understand why we need the 100ms wait if there is no callback, but not if there isn't.

theol0403:
> Is that possible now? This PR doesn't delete any code.

Some features from that plugin can now be moved to core, but not the whole shebang. 

> I don't really understand why this is worth adding, but whatever.

Again, thinning the bridge between VSCode and Neovim. This extension currently doesn't support some VSCode-native commands in normal mode, because we haven't "let VSCode do what it does best in insert mode". This feature lets you ease the transition from normal mode to insert. 

> I want to ask in advance, which shortcut key do you plan to bind it to?

Only augment the built-in VSCode bindings (like we do for commenting, formatting, etc).

```json
{
  "key": "ctrl+shift+l",
  "command": "editor.action.selectHighlights",
  "when": "editorFocus"
}
{
  "key": "ctrl+d",
  "command": "editor.action.addSelectionToNextFindMatch",
  "when": "editorFocus"
}
{
  "key": "",
  "command": "editor.action.addSelectionToPreviousFindMatch"
}
{
  "key": "ctrl+shift+down",
  "command": "editor.action.insertCursorBelow",
  "when": "editorTextFocus"
}
{
  "key": "ctrl+shift+up",
  "command": "editor.action.insertCursorAbove",
  "when": "editorTextFocus"
}
{
  "key": "shift+alt+i",
  "command": "editor.action.insertCursorAtEndOfEachLineSelected",
  "when": "editorTextFocus"
}
```

runtime/lua/vscode/api.lua
xiyaowong: - if not callback then callback = function() end

Believe me, I have tried it


- there is no callback, but not if there isn't.

Note that I only added delays when switching modes. There are no delays in insert mode.

Without delay, executing subsequent logic before the synchronization process between vscode and neovim is complete will lead to errors.

You can modify the code and then check the "no callback" test.

runtime/lua/vscode/api.lua
xiyaowong: - For multi-cursor, we've considered blockwise and treat linewise as a range line by line.
- For selection syncing, we've considered blockwise as well, but like charwise, we only consider the start and end positions of the selected region.
- In `into_insert`, we treat all as charwise (meaning we don't support blockwise, which isn't useful).

We can refactor in another PR, but only by separating functions. Otherwise, the function will become too complex.

xiyaowong:
We shouldn't touch Ctrl+d.

runtime/lua/vscode/api.lua
xiyaowong: > And I still don't understand why we need the 100ms wait if there is no callback, but not if there isn't.

Sorry, I realized that the logic was wrong

PR #2055, opened 2024-06-02
feat(lua): Add "vscode.with_insert"

### LINKED ISSUE #1999: feature: expose `vscode.action_into_insert` and `neovim.action_into_insert`

xiyaowong:
fix #1999

Compared to #2052, the callback function is no longer optional, which makes it simpler and easier to understand and maintain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,9 +613,8 @@ Examples:
 
     ![select-next](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/5a93c87e-626a-4a70-a9ef-5084f747c7ef)
 
--   Make <kbd>Tab</kbd> work correctly in VSCode snippet mode.
-
-    e.g. Wraps `editor.action.refactor`
+-   Make "editor.action.refactor" work correctly on the selection and support snippet manipulation after entering VSCode
+    snippet mode.
 
     ```lua
     vim.keymap.set({ "n", "x" }, "<leader>r", function()
--- a/README.md
+++ b/README.md
@@ -601,16 +601,31 @@ Parameters:
 
 Examples:
 
-```lua
--- Make `editor.action.addSelectionToNextFindMatch` work correctly in any mode.
-vim.keymap.set({ "n", "x", "i" }, "<C-d>", function()
-  vscode.with_insert(function()
-    vscode.action("editor.action.addSelectionToNextFindMatch")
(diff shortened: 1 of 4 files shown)


theol0403 (approved):

README.md
```diff
@@ -589,6 +590,43 @@ Parameters:
...
+    vim.keymap.set({ "n", "x", "i" }, "<C-d>", function()
+      vscode.with_insert(function()
+        vscode.action("editor.action.addSelectionToNextFindMatch")
+      end)
+    end)
+    ```
+
+    ![select-next](https://github.com/vscode-neovim/vscode-neovim/assets/47070852/5a93c87e-626a-4a70-a9ef-5084f747c7ef)
+
+-   Make <kbd>Tab</kbd> work correctly in VSCode snippet mode.
```
theol0403, line 616: How does this make Tab work correctly?

theol0403: I understand now, but it should be worded differently. It should be something like "support snippet navigation after a refactor".

xiyaowong: All right. This is indeed an indirect benefit. Because it is in insert mode, it is normal.

PR #2059, opened 2024-06-02
set neovim working directory on every buffer load

### LINKED ISSUE #2058: bug: Working directory not set correctly with WSL installation (again)
### Check the following:

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Windows 11 w/ WSL (Ubuntu)

### Describe the bug

Getting the current working directory doesn't work when running with WSL.  !pwd gives the vscode installation directory, and so :e, :sp, :w, etc all don't work properly, nor do their tab completions.

There was an existing #520 but it has returned, because the fix in PR #1689 is no longer valid due to multi-root workspaces, etc.  Now the workspace directory is only valid in the context of a particular file.

Therefore, the working directory needs to be set upon every buffer change, instead of just once when running neovim.

I tested this approach and it seems to work.  I'll submit the PR later today.

### Steps To Reproduce

1. Install neovim within WSL, and set up the extension to use it
2. Within a WSL terminal in a project directory, run `code .`
3. open one of the files
4. hit :!pwd

### Expected Behavior

EXPECTED: project directory shows
ACTUAL: vscode installation path shows

captaincaius:
Resolve #2058

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -355,6 +355,11 @@ export class BufferManager implements Disposable {
             logger.debug(`target editor not found <check 1>, return to active editor`);
             return returnToActiveEditor();
         }
+        let uri = targetEditor.document?.uri;
+        const workspaceFolder = uri && workspace.getWorkspaceFolder(uri);
+        if(workspaceFolder) {
+            await this.client.request("nvim_set_current_dir", [workspaceFolder.uri.fsPath]);
+        }
         if (window.activeTextEditor === targetEditor) return;
         // since the event could be triggered by vscode side operations
         // we need to wait a bit to let vscode finish its internal operations
@@ -376,7 +381,7 @@ export class BufferManager implements Disposable {
         }
         if (window.activeTextEditor === targetEditor) return;
         await this.main.cursorManager.waitForCursorUpdate(targetEditor);
-        const uri = targetEditor.document.uri;
+        uri = targetEditor.document.uri;
         const { scheme } = uri;
         switch (scheme) {
             case "output": {


xiyaowong (changes requested):
- It is necessary to perform checks to avoid redundant calls.
- Ensure it works correctly during remote development.
- Can we remove this?:
https://github.com/vscode-neovim/vscode-neovim/blob/927ed955d9075f40bb5f2f40d2ad9e057f13906c/src/main_controller.ts#L209-L213

xiyaowong:
Also, can we remove this?

https://github.com/vscode-neovim/vscode-neovim/blob/927ed955d9075f40bb5f2f40d2ad9e057f13906c/src/main_controller.ts#L209-L213

ollien:
> It is necessary to perform checks to avoid redundant calls.

Why? Isn't the worst case we change the PWD to a directory we're already in, which is a no-op?

xiyaowong:
Alright, Vim has already checked it.

https://github.com/neovim/neovim/blob/3a1515bfee59710b4da9bfdf0fc4ecb5b13a00db/src/nvim/ex_docmd.c#L5789-L5790

captaincaius:
> * ~It is necessary to perform checks to avoid redundant calls.~
> * Ensure it works correctly during remote development.
> * Can we remove this?:
>   https://github.com/vscode-neovim/vscode-neovim/blob/927ed955d9075f40bb5f2f40d2ad9e057f13906c/src/main_controller.ts#L209-L213

Yeah I think you're right - it's probably not needed anymore.  I didn't try it though - I wasn't sure you folks would want it in the same PR or a future cleanup issue.

Since I need to reroll for the lint stuff anyway, I might as well add it.  Clean commit and PR coming.

PR #2060, opened 2024-06-03
fix: set working directory on every buffer change

captaincaius:
refers to #2058

redo of #2059


xiyaowong:
Thanks for your contribution! However, this is not the solution. Please refer to #2062.

Issue #2046, opened 2024-05-31
bug: output panel opens automatically for failed searches


mkaminsky:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

v0.10.0-dev-3080+gefb44e0ca

### Operating system/version

Linux Debian testing

### Describe the bug

Starting with v1.14.0 (PR #2029):  If you search for text (using / in normal mode) that does not exist in the buffer, the output panel opens automatically.  I find this somewhat jarring given how frequent it is to have no search results.

Is it possible to make that optional (as discussed in the PR)?

Thanks!

### Steps To Reproduce

1. Create / open a file.
2. Type some text, e.g., "abcdef"
3. In normal mode, type / to begin searching.
4. Enter "zzz" into the "Forward Search" field.
5. Output panel opens showing: `E486: Pattern not found: zzz`

### Expected Behavior

Don't open the output panel automatically (at least for failed searches).

mschwartz:
LOL.  Annoying bug.  I have my terminal and output tabs in the same right side panel.  It's constantly activating the output tab thus hiding the terminal one.  There's no place to move either to make my workflow satisfactory...

Maybe me and hundreds of other users are spending a lot of time looking for a solution to fix this :)

xiyaowong:
@ollien It seems that this has already been checked here.
https://github.com/vscode-neovim/vscode-neovim/blob/898ad178bd8b980fc45007f95829145ba542627b/src/messages_manager.ts#L43-L45
So ignoring the return_prompt is expected. 😂

ollien:
@xiyaowong this was a separate change from that. The thing is that if you have output that is only one line, it looks like nothing happens and you don't see your output. 

I'm surprised not finding a match sends us a `return_prompt`, that sounds inconsistent with normal nvim, but I am not able to investigate immediately.

ollien:
> The output when search match not found is two lines long. FWIW.

Interesting, because the two lines long check is not new.

mschwartz:
Maybe it’s a change to neovim itself?

I any case, can we fix this?  It is making vs code almost unusable….

thanks

ollien:
Yes, I will investigate when I get a spare cycle. You are welcome to downgrade to an earlier release in the interim :)

ollien:
[...]

PR #2068, opened 2024-06-03
fix: ensure statusline messages are shown for a minimum amount of time before clears are effective

### LINKED ISSUE #2063: bug: messages in statusline are cleared almost immediately
### Check the following:

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Fedora 40

### Describe the bug

We use the status line to show messages that are output by nvim, but they often don't show for more than a frame or two. It's a little hard to get on recording. Some investigation tells me this is due to our [handling of `return_prompt`](https://github.com/vscode-neovim/vscode-neovim/blob/2b9c9dc992273ffd4b1e7f082129af880d469c5a/src/status_line_manager.ts#L90-L92). Once we simulate a CR, we immediately get a `msg_clear`, which clears the statusline immediately

### Steps To Reproduce

1. Search for some text that isn't in a document
2. As you hit enter, look at the statusline. If your timing is right, you will see the error about the text not being found for a frame or two (or if you're unlucky, not at all).

### Expected Behavior

The messages in the status line should be readable.

ollien:
Fixes #2063 
 
A video's worth a thousand words:

https://github.com/vscode-neovim/vscode-neovim/assets/977151/a7354726-1132-4cbf-b9fa-7d2c3f1fdb7b

Notably

1. If a message is sent, we always display it immediately
2. If a clear event is sent, we ensure the message has been displayed for at least five seconds before processing the clear. Any clears received after this threshold will happen immediately

I thought about just letting the message hang around, but I felt we should respect nvim's wishes in still clearing the message. Also, because we combine statusline components into a single line (e.g. `--INSERT-- | E486: Pattern not found: sdfsd`), persisting the message forever just felt weird. I think a timeout is a good tradeoff.

There's also a driveby bugfix of not putting the `return_prompt`/`press-enter` message into the status bar.


src/utils/timer.ts
```diff
@@ -0,0 +1,76 @@
...
+// either allow the passing of a timeout *OR* a set of timer functions that can be used for manual timer firing
+export type TimerParam<K> = K extends NodeJS.Timer ? number : TimerFunctions<K>;
+
+/**
+ * Timer is an encapsulation around setTimeout, which allows a single action to be re-executed or cancelled
+ * after scheduling.
+ *
+ * This function is slightly overly generic to allow for extensibility of this timer in testing (without having to depend on time)
+ */
+export class Timer<K = NodeJS.Timer> implements Disposable {
```
ollien, line 17: I will be the first to say: this class (and its existence, for that matter) is more complex than it needs to be. However, I think by having it we gain a few notable benefits

1. We have a way to have clearable and restartable timers without having to worry about missed calls to `clearTimeout` or missed stores of the result of `setTimeout`.
2. This logic can be unit tested. I actually originally wrote this without `TimerParam` (and used `setTimeout`s in unit tests), but I've been burned by that approach before (flaky tests, arbitrarily inflating the length of unit tests), so I chose this route.
3. By making a more "generic" Timer, we can inject it into other classes to facilitate _their_ unit testing (e.g. `StautsLineMessageTimer`).

If you all don't turn your nose up at this, I'll do a follow-on PR which converts existing usage of `setTimeout` and `clearTimeout` to use `Timer` objects.

src/test/unit/status_line/status_line_message_timer.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import { StatusLineMessageTimer } from "../../../status_line/status_line_message_timer";
+import { TimerFunctions } from "../../../utils";
+
+/**
+ * Simulates a set of timer functions by keeping track of what timers are created and cleared, and allowing manual
+ * firing of known timers.
+ *
+ * The ids are always incrementing.
+ */
+class ManualTimerContainer implements TimerFunctions<number> {
```
ollien, line 12: You'll notice this is duplicated from `Timer`. I chose to do this, rather than putting it in a `utils` function, as the repetition really isn't that bad and also means that each class's test suite can change behavior as needed, without worrying about external impact. That said, I waffled on this quite a bit so I'm not tied to it

xiyaowong (changes requested):
- Reinvent debounce
- Please use `debounce` and `CancellationToken` to implement this pr

src/utils/timer.ts
ollien: Disregard this

src/test/unit/status_line/status_line_message_timer.test.ts
ollien: Disregard this

package.json
```diff
@@ -2138,7 +2138,7 @@
     },
     "devDependencies": {
         "@johnnymorganz/stylua-bin": "^0.20.0",
-        "@types/lodash-es": "^4.17.12",
+        "@types/lodash": "^4.17.4",
```
ollien, line 2142: I explained this in the commit message, but I replaced this dependency because we actually compile our TS down to JS that uses commonjs anyway.

Importing the ESModule from a unit test proved troublesome (I was getting the below error). Honestly, I'm surprised this worked at runtime at all, given we weren't using commonjs, but the JS import system is admittedly not my most well-studied area.

```
> vscode-neovim@1.15.0 test:unit
> mocha --exit ./out/test/unit/*.js ./out/test/unit/**/*.js


 Exception during run: Error [ERR_REQUIRE_ESM]: require() of ES Module /home/nick/Documents/code/vscode-neovim/node_modules/lodash-es/lodash.js from /home/nick/Documents/code/vscode-neovim/out/status_line/status_line_message_timer.js not supported.
Instead change the require of lodash.js in /home/nick/Documents/code/vscode-neovim/out/status_line/status_line_message_timer.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/home/nick/Documents/code/vscode-neovim/out/status_line/status_line_message_timer.js:4:21)
    at Object.<anonymous> (/home/nick/Documents/code/vscode-neovim/out/test/unit/status_line/status_line_message_timer.test.js:28:37)
    at async formattedImport (/home/nick/Documents/code/vscode-neovim/node_modules/mocha/lib/nodejs/esm-utils.js:9:14)
    at async exports.requireOrImport (/home/nick/Documents/code/vscode-neovim/node_modules/mocha/lib/nodejs/esm-utils.js:42:28)
    at async exports.loadFilesAsync (/home/nick/Documents/code/vscode-neovim/node_modules/mocha/lib/nodejs/esm-utils.js:100:20)
    at async singleRun (/home/nick/Documents/code/vscode-neovim/node_modules/mocha/lib/cli/run-helpers.js:125:3)
    at async exports.handler (/home/nick/Documents/code/vscode-neovim/node_modules/mocha/lib/cli/run.js:370:5) {
  code: 'ERR_REQUIRE_ESM'
}
```

If there's a better way to solve this, do let me know, but I think this is correct.

xiyaowong:
I no longer know what this PR is doing.

ollien:
Is there something I can clarify? The intention hasn't changed from the original description, I altered the implementation to be in line with your feedback (i.e. using lodash's `debounce`, which you were correct was a much simpler implementation).

xiyaowong:
```diff
diff --git a/src/status_line_manager.ts b/src/status_line_manager.ts
index 316e4a4..a49336e 100644
--- a/src/status_line_manager.ts
+++ b/src/status_line_manager.ts
@@ -1,4 +1,12 @@
-import { Disposable, StatusBarAlignment, StatusBarItem, window } from "vscode";
+import {
+    CancellationToken,
+    CancellationTokenSource,
+    Disposable,
+    StatusBarAlignment,
+    StatusBarItem,
+    window,
+} from "vscode";
+import { debounce } from "lodash-es";
 
 import { config } from "./config";
 import { EventBusData, eventBus } from "./eventBus";
@@ -15,6 +23,8 @@ enum StatusType {
 export class StatusLineManager implements Disposable {
     private disposables: Disposable[] = [];
 
+    private msgCancellationSource?: CancellationTokenSource;
+
     // ui events
     private _modeText = "";
     private _cmdText = "";
@@ -99,6 +109,7 @@ export class StatusLineManager implements Disposable {
                         }
                     }
                 }
+                this.msgCancellationSource?.cancel();
                 this.setStatus(str, StatusType.Msg);
                 break;
             }
@@ -117,7 +128,9 @@ export class StatusLineManager implements Disposable {
                 break;
             }
             case "msg_clear": {
-                this.setStatus("", StatusType.Msg);
+                this.msgCancellationSource?.cancel();
+                this.msgCancellationSource = new CancellationTokenSource();
+                this.clearMsg(this.msgCancellationSource.token);
                 break;
             }
         }
@@ -126,6 +139,15 @@ export class StatusLineManager implements Disposable {
         }
     }
 
+    private clearMsg = debounce(
+        (token: CancellationToken) => {
+            if (token.isCancellationRequested) return;
+            this.setStatus("", StatusType.Msg);
+        },
+        5000,
+        { leading: false, trailing: true },
+    );
+
     dispose() {
         disposeAll(this.disposables);
     }
```

ollien:
@xiyaowong I appreciate your help and feedback, but would really appreciate if you'd work with me a bit, instead of just posting patches without explanation. I really do want to take the time to make meaningful contributions to this project... 

The implementation your shared is indeed quite similar to the one I shared, though I didn't use a `CancellationToken` (in fact I don't think it's necessary even in your implementation, given your call to `cancel`). The only difference is our choice in whether or not to isolate that state into a class, and the unit tests surrounding that. I chose to include these tests because there are subtleties around the UX of this implementation that I think are important to document, to help future maintainers understand when they're changing the functionality of the status line

xiyaowong:
I just hope we don't complicate minor issues.

ollien:
I don't think there's much complication. Yes there's another class involved, but it gives us the ability to test the implementation quite readily. The more state we cram into each manager, the harder it is to untangle problems, and I think this is a reasonable way to factor it.

xiyaowong:
I think the timer added in this PR is essentially reinventing debounce.

ollien:
How is it re-inventing `debounce` if it uses `debounce`? Or is your qualm that the class is wrapping debounce at all? The extra class isolates the state (and clarifies the entry point), and allows us to test the changes.

xiyaowong:
> in fact I don't think it's necessary even in your implementation, given your call to `cancel`

Yes, it can be further simplified.

```diff
diff --git a/src/status_line_manager.ts b/src/status_line_manager.ts
index 316e4a4..6732de0 100644
--- a/src/status_line_manager.ts
+++ b/src/status_line_manager.ts
@@ -15,6 +15,8 @@ enum StatusType {
 export class StatusLineManager implements Disposable {
     private disposables: Disposable[] = [];
 
+    private clearMsgTimer?: NodeJS.Timeout;
+
     // ui events
     private _modeText = "";
     private _cmdText = "";
@@ -99,6 +101,7 @@ export class StatusLineManager implements Disposable {
                         }
                     }
                 }
+                clearTimeout(this.clearMsgTimer);
                 this.setStatus(str, StatusType.Msg);
                 break;
             }
@@ -117,7 +120,10 @@ export class StatusLineManager implements Disposable {
                 break;
             }
             case "msg_clear": {
-                this.setStatus("", StatusType.Msg);
+                clearTimeout(this.clearMsgTimer);
+                this.clearMsgTimer = setTimeout(() => {
+                    this.setStatus("", StatusType.Msg);
+                }, 5000);
                 break;
             }
         }

```

---

Many times, discussions are based on personal preferences and it's hard to reach a conclusion. Let's wait for others to review.

ollien:
FWIW, that implementation defers _all_ clears, even if a message has not recently been sent.

theol0403 (commented):
@ollien I appreciate your time working on this extension, it has been very beneficial to the project and it's great to have another maintainer on board!

I have some questions about this implementation:
* why, in the first place, is this all necessary? Why do we need all this logic surrounding displaying messages, when the extension strives to simply extend the nvim UI and use the vscode interface as a view?
* If the answer is that the "nvim way" is a prompt you have to dismiss and this doesn't make sense with vscode, how does noice + family do it (do they use timers + automate CR)? 
* For the slightly worse UX, does ignoring the `clear` altogether significantly reduce the code usage?
* Finally, if we get to here and this implementation still makes sense, can you explain why the custom timer class is needed compared to, say, "show the message for 5s and then clear it".  

Thanks!

ollien:
@theol0403 

Totally valid

> why, in the first place, is this all necessary? Why do we need all this logic surrounding displaying messages, when the extension strives to simply extend the nvim UI and use the vscode interface as a view?

I wish it weren't. I mentioned this briefly in the original issue, but `return_prompt` is sent extremely frequently when using `ext_messages` (in fact this inconsistency was the cause of [other bugs](https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2144175058)). When we clear the `press-enter` prompt, nvim sends us a `msg_clear` which we immediately process, hiding the message. 


> If the answer is that the "nvim way" is a prompt you have to dismiss and this doesn't make sense with vscode, how does noice + family do it (do they use timers + automate CR)?

Full disclosure: I've never even heard of noice before now, but looking quickly, yes, it seems they do also [send a `<cr>`](https://github.com/folke/noice.nvim/blob/2bf04290b32dc5b1a002a4888c95147bb91ec6f2/lua/noice/ui/msg.lua#L147-L149) upon [receiving a `return_prompt`](https://github.com/folke/noice.nvim/blob/2bf04290b32dc5b1a002a4888c95147bb91ec6f2/lua/noice/ui/msg.lua#L83-L92). I don't think they use timers, because they [replace the last message](https://github.com/folke/noice.nvim/blob/2bf04290b32dc5b1a002a4888c95147bb91ec6f2/lua/noice/ui/msg.lua#L61), and they do this every time they get any kind of `show` message (e.g. `showmode`), making its functionality a bit different than ours (we show all "message" components in the statusline at once). I'd have to play around with this extension to see how it bubbles up in UX. I felt that here, the timers were a decent tradeoff.

> For the slightly worse UX, does ignoring the clear altogether significantly reduce the code usage?

Yeah, ignoring clears is quite simple. I actually prototyped this idea originally, but I really didn't like my last error message sticking around forever until I did something else. It just felt broken.

> Finally, if we get to here and this implementation still makes sense, can you explain why the custom timer class is needed compared to, say, "show the message for 5s and then clear it".

For sure! So, a couple of points here

1. The class itself isn't strictly necessary. The class allows us to encapsulate the state, rather than adding more fields on the `StatuslineManager` (and I do want to underscore, as of the latest revision, it's not a custom timer implementation, it's just a wrapper for lodash's `debounce`). I admit, for this already small manager, two booleans probably wouldn't break the bank. When I wrote it originally, I was also dealing with timer state, and it felt cluttered and hard-to-grok to have 3-4 fields on the original class just to manage this small bit of functionality. I think some of our more complex managers suffer from a problem of having lots of properties that were added to manage state, but can be hard to grok if you don't know what's going on. I felt that by having this class, the necessity for these pieces of state could be isolated, and better yet, individually tested. I didn't see a great way to add an integration test for this, and I figured the "fast" unit test might be better anyway.
2. "show the message for 5s and then clear it" was actually my original implementation. However, it's not quite as simple as that. Consider the following sequence of events:

- `msg_show` is sent to us. We show the message and kick off a 5s timer
- We press `<cr>` to clear the `return_prompt`.
- This immediately sends a `msg_clear`  to us, clearing the statusline
- 5s later, we send a `msg_clear` to ourselves, doing nothing.

You could also do something like @xiyaowong suggested, where you kick off a 5s timer upon receiving a clear. However, this means that if nvim feels the need to clear the prompt for some _other_ reason, it's heavily delayed. What you _really_ want is a message that is shown indefinitely, but ensures a `msg_clear` does not affect it for some minimum amount of time. It is precisely this which the timer class aimed to solve, and I hope the tests can show that. This is why I mentioned there are lots of subtleties around the UX; it's actually kinda hard to get this "feeling right" (which I admit, is not my forte)!


All of that said: if the functionality is not what we want, I'm totally ok with that. I really am flexible on these things, and I hope my past comments have reflected that. I admittedly grew a bit frustrated last night, and for that I apologize.

ollien:
@justinmk 

> [setStatusBarMessage](https://code.visualstudio.com/api/references/vscode-api#window) has some interesting overloads

Those are really interesting (though the timeout alone doesn't solve the `msg_clear` concerns). I think there's a few options we have for redoing this status line, but I didn't think that the original issue warranted entirely rethinking how we handled the statusline.

xiyaowong:
If we know that a `msg_clear` will be received after inputting `<Cr>` to clear the `return_prompt` (and we want to ignore this `msg_clear`), can we use a flag to delay processing the next `msg_clear`?

ollien:
That's exactly what the timer class does

https://github.com/ollien/vscode-neovim/blob/814c7d07276b92572fc901dcdb37139b5c10ddc4/src/status_line/status_line_message_timer.ts#L47

ollien:
How do we want to proceed with this PR? The way I see it, there are three options

1. Merge this as-is (or with changes)
2. Scrap this, and smiply begin to ignore `msg_clear` events
3. Redesign the statusline entirely to make the original bug less of a problem, similar to what @justinmk suggested

xiyaowong:
> 1. Merge this as-is (or with changes)
> 2. Scrap this, and simply begin to ignore `msg_clear` events

My opinion remains unchanged. If a flag can solve the issue, why is a new class necessary? If testing is needed, you can add a command "vscode-neovim.test-statusline" to retrieve all the data from the current status_line_manager for testing purposes.

> 3. Redesign the statusline entirely to make the original bug less of a problem, similar to what @justinmk suggested

Users can implement "hideAfterTimeout," in nvim

ollien:
> If a flag can solve the issue, why is a new class necessary?

I addressed this in https://github.com/vscode-neovim/vscode-neovim/pull/2068#issuecomment-2147650392. To recap, it's not fully necessary, it's a code organization thing, and this class encapsulates the state machine necessary to provide this UX. I'm not sure what else I can say to answer your question beyond that comment. 

I'm not sure why we would build more commands around this just to expose innerworkings for testing, when it can be solved by factoring code in a way that's directly callable from a unit test. Not to mention, the integration test suite is already quite long and adding small unit tests seems like it would be beneficial, no?

theol0403 (approved):
Seems well-reasoned enough, thanks. Ultimately the status line manager is self-contained and this won't significantly increase complexity. Perhaps add a few more inline comments explaining the flow? The goal would be to have a reader understand how the timer can be removed, and where the interface between the two lie. 

I think the "If a flag can solve the issue, why is a new class necessary?" discussion is valid, and I would agree if it indeed could be that simple. If the stuff in the class could be moved to the manager in a way that reduces boilerplalte, indirection, and extra member variables, then I would suggest to do so (despite the tests being more cumbersome). But, if the exact same code were to be copied over... having it in a class is better.

ollien:
> But, if the exact same code were to be copied over... having it in a class is better.

Yeah, my thought is that this would mostly be the same code copied over. As you allude to, to remove the timer, one only has to remove the single member (and its associated invocations

ollien:
I noticed a really curious behavior, which also seems to be on master, and is probably related to this somehow. If you hit `:` then `<C-r>=`, then `<esc>` twice, the final `<esc>` does not result in a `cmdline_hide` being sent to the extension :thinking:

PR #2070, opened 2024-06-04
fix: also sync the cursor position from VSCode to Neovim in insert mode

### LINKED ISSUE #2069: In insert mode, also synchronize the cursor position from VSCode to Neovim
First: changing the selections of textEditor will break the cursor-related functions of vscode, such as placeholder jumps in snippet mode, multi-cursor selection, etc.

In insert mode, when sending keys to neovim, it is necessary to allow synchronization of cursor changes in neovim (for example, the user may bind `<C-a>` to `<Home>`), and the cursor position in neovim needs to be updated so that the cursor is in the latest position when exiting insert mode.
https://github.com/vscode-neovim/vscode-neovim/blob/62bab0597590ba4445b57b98653b2f0dd60ee002/src/typing_manager.ts#L294-L303
The problem is: sending all keystrokes will execute this logic. If the function corresponding to a key in neovim happens to call the cursor function of vscode, there will be a problem:

select
<sub> Ctrl+i calls editor.action.addSelectionToNextFindMatch </sub>

Because the cursor position is inconsistent, synchronizing the cursor changes in neovim causes vscode to lose the selected range.

Solution:

In insert mode, also update the cursor position of neovim. Also, in insert mode, after the cursor in vscode is synchronized to neovim, the cursor changes from neovim should no longer be synchronized to avoid the selection in vscode being destroyed.

xiyaowong:
fixes #2069

This PR may have some performance risks and needs more testing

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -339,25 +339,26 @@ export class CursorManager implements Disposable {
         // reset cursor style if needed
         this.updateCursorStyle(this.main.modeManager.currentMode.name);
 
-        // wait for possible layout updates first
         logger.debug(`Waiting for possible layout completion operation`);
         await this.main.bufferManager.waitForLayoutSync();
-        // wait for possible change document events
         logger.debug(`Waiting for possible document change completion operation`);
         await this.main.changeManager.getDocumentChangeCompletionLock(editor.document);
         await this.main.changeManager.documentChangeLock.waitForUnlock();
         logger.debug(`Waiting done`);
 
-        // ignore selection change caused by buffer edit
-        const selection = editor.selection;
-        const documentChange = this.main.changeManager.eatDocumentCursorAfterChange(editor.document);
-        if (documentChange && documentChange.isEqual(selection.active)) {
-            logger.debug(`Skipping onSelectionChanged event since it was selection produced by doc change`);
-        } else {
+        try {
+            const selection = editor.selection;
+            // ignore selection change caused by buffer edit
+            const documentChange = this.main.changeManager.eatDocumentCursorAfterChange(editor.document);
+            if (documentChange && documentChange.isEqual(selection.active)) {
+                logger.debug(`Skipping onSelectionChanged event since it was selection produced by doc change`);
+                return;
+            }
+
             logger.debug(
-                `Applying changed selection, kind: ${kind},  cursor: [${selection.active.line}, ${
-                    selection.active.character
-                }], isMultiSelection: ${editor.selections.length > 1}`,
+                `Applying changed selection, kind: ${kind && TextEditorSelectionChangeKind[kind]},`,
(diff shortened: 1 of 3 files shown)


src/cursor_manager.ts
```diff
@@ -361,14 +360,24 @@ export class CursorManager implements Disposable {
...
             } else {
-                if (kind != TextEditorSelectionChangeKind.Mouse || !config.disableMouseSelection)
-                    await this.updateNeovimVisualSelection(editor, selection);
+                if (selection.isEmpty) {
+                    // exit visual mode when clicking elsewhere
+                    if (this.main.modeManager.isVisualMode && kind == TextEditorSelectionChangeKind.Mouse)
+                        await this.client.input("<Esc>");
+                    await this.updateNeovimCursorPosition(editor, selection.active);
+                } else {
+                    if (kind != TextEditorSelectionChangeKind.Mouse || !config.disableMouseSelection)
```
ollien, line 378: I'm surprised the linter didn't flag this usage of `!=` vs `!==`

justinmk: need to enable "eqeqeq" : https://github.com/neovim/node-client/blob/8982da5532278f178286a11696939cbf241b4046/.eslintrc.js#L37-L38

ollien: Yeah, I assumed (incorrectly) it was in the recommended ruleset :)

xiyaowong: `npx eslint-flat-config-viewer` can see where the problem is

src/cursor_manager.ts
```diff
@@ -361,14 +360,24 @@ export class CursorManager implements Disposable {
...
+                // changes in editor selections when synchronizing the cursor
+                // position of neovim to vscode in insert mode, thus breaking
+                // the cursor-related functions of vscode, such as placeholder
+                // jumps in snippet mode, multi-cursor selection, etc.
+                this.setWantInsertCursorUpdate(editor, false);
                 await this.updateNeovimCursorPosition(editor, selection.active);
             } else {
-                if (kind !== TextEditorSelectionChangeKind.Mouse || !config.disableMouseSelection)
-                    await this.updateNeovimVisualSelection(editor, selection);
+                if (selection.isEmpty) {
```
ollien, line 372: nit: you can make this an `else if` chain of 

```ts
if (this.main.modeManager.isInsertMode) {
   // ...
}  else if (selection.isEmpty) {
  // ...
} else {
  // ...
}

xiyaowong: Style issue, I rarely (and don't like to) use `elseif` for conditions that are not in the same group, as the logical boundaries are unclear. I generally use `switch` or early return instead.

xiyaowong:
Doing too many things in insert mode is not a good idea. Only compromise.

Issue #1260, opened 2023-07-04
Issue Topic: Replace Altercommand


theol0403:
The altercommand hack is very gross. There is some useful information in https://github.com/vscode-neovim/vscode-neovim/discussions/887 that should help with removing it.

theol0403:
@justinmk some early questions.

I understand how `BufReadCmd` can be used to intercept `:edit` and let vscode handle loading the file.

However, what about commands like `:vsplit`? In this case, vscode does not sync window position with neovim. That would be too complex. Thus, `workbench.action.splitEditorRight` needs to be called at some point. 

Is there a `WindowSplitCmd` autocommand? Alternatively, is there a way for the `BufReadCmd` to know whether it was called by `:vsplit`? I suppose that would be ideal! 

Finally, I am having a little trouble understanding how to access the filename passed to `:edit`. I found you can use `v:cmdbang` to detect the bang, but I'm not sure if `v:cmdarg` actually contains the filename.

theol0403:
Additionally, what about `<C-w>v`? The keybindings are less gross, but it would be good if it would trigger an autocommand we could use to call `workbench.action.splitEditorRight`.

theol0403:
@justinmk just another ping in case this notification got lost in the noise 
I'd love to remove altercommand

theol0403:
Same question with `:saveas`.
When `BufWritePre` is called, how can I know to call the saveas command?

theol0403:
@justinmk bump.

It looks like `:vsplit` does not trigger `BufReadPre`, so not sure how to intercept that.

theol0403:
In neovim, I tried collecting as much information as possible. However, this is the data I get on a `:split` command with `WinNew` - I can't figure out how to know what kind of split it is. 

*maybe* I can try to figure out the position of the new window and calculate the most likely split, but that seems like a lot of work. However, a window position sync api does seem like it could work well with some nvim plugins.

justinmk:
> Is there a `WindowSplitCmd` autocommand? 

`WinNew` (as you noticed)

> Alternatively, is there a way for the `BufReadCmd` to know whether it was called by `:vsplit`? ... the data I get on a `:split` command with `WinNew` - I can't figure out how to know what kind of split it is.

There is `getwininfo(win_getid())` and `win_screenpos()` which can be used to intuit the layout. Vim itself doesn't know what _command_ created the window, it only knows the dimensions.

[...]

PR #2075, opened 2024-06-05
refactor(prototype): integrate BufWriteCmd for basic writing

ollien:
This is a bit of a prototype. I may make some minor changes but the goal of this PR is to be a first pass (based on [this discussion](https://github.com/vscode-neovim/vscode-neovim/issues/1260#issuecomment-2148351054) that handles the most basic `:w` cases (and also solves #521), which will have value when merged as-is.

I left some comments below about things I'm thinking about, but this is the bulk of the changes.

### DIFF
--- a/runtime/lua/vscode/files.lua
+++ b/runtime/lua/vscode/files.lua
@@ -1,0 +1,23 @@
+local api = vim.api
+local vscode = require("vscode.api")
+
+local M = {}
+
+function collect_file_info(event)
+  return {
+    buf = event.buf,
+    path = event.file,
+  }
+end
+
+function M.setup()
+  api.nvim_create_autocmd({ "BufWriteCmd" }, {
+    pattern = "*",
+    callback = function(event)
+      local info = collect_file_info(event)
+      vscode.action("save_buf", { args = { info.buf, info.path } })
+    end,
+  })
+end
+
+return M
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -280,4 +280,10 @@ function M.wslpath(path)
   return vim.trim(ret)
 end
 
+-- Get the current Neovim CWD
+-- @return string
+function M.cwd()
+  return fn.getcwd()
+end
+
 return M
--- a/runtime/vscode/overrides/vscode-file-commands.vim
+++ b/runtime/vscode/overrides/vscode-file-commands.vim
@@ -49,10 +49,11 @@ AlterCommand e[dit] Edit
 AlterCommand ex Ex
 AlterCommand ene[w] Enew
 AlterCommand fin[d] Find
-AlterCommand w[rite] Write
 AlterCommand sav[eas] Saveas
 AlterCommand wa[ll] Wall
 AlterCommand q[uit] Quit
+" TODO: Wq/Wqall need to be handled with the autocmd
+" Just putting "write" in saveAndClose doesn't work, because the writecmd is asynchronous
 AlterCommand wq Wq
 AlterCommand x[it] Xit
 AlterCommand qa[ll] Qall
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -158,6 +158,8 @@ export class BufferManager implements Disposable {
                 editor.options = { tabSize, insertSpaces, lineNumbers };
             },
         );
+
+        actions.add("save_buf", (bufID: number, filename: string) => this.handleSaveBuf(bufID, filename));
     }
 
     public dispose(): void {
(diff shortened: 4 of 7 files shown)


runtime/vscode/overrides/vscode-file-commands.vim
```diff
@@ -49,10 +49,11 @@ AlterCommand e[dit] Edit
 AlterCommand ex Ex
 AlterCommand ene[w] Enew
 AlterCommand fin[d] Find
-AlterCommand w[rite] Write
 AlterCommand sav[eas] Saveas
 AlterCommand wa[ll] Wall
```
ollien, line 53: In response to this comment https://github.com/vscode-neovim/vscode-neovim/pull/2027#issuecomment-2141140685

I still need to add a check to make sure that `wall` doesn't affect lots of other buffers

src/buffer_manager.ts
```diff
@@ -410,6 +412,74 @@ export class BufferManager implements Disposable {
...
+        }
+
+        this.saveDocument(bufID, document, filename).catch((err) => {
+            window.showErrorMessage(`Failed to save document to ${filename}: ${err}`);
+        });
+    }
+
+    private async saveDocument(bufID: number, document: TextDocument, filename: string): Promise<void> {
+        const needSaveAs = await this.needSaveAs(document, filename);
+        if (needSaveAs) {
```
ollien, line 428: The idea here is that because `saveas` is actually the complicated case, we're going to let VSCode do a standard `saveas`, and handle all of the editor renaming for us. This will have to change in the future if we want better integration, but as @theol0403 pointed out, this isn't a regression from current behavior (which always opens the dialog anyway)

runtime/vscode/overrides/vscode-file-commands.vim
```diff
@@ -49,10 +49,11 @@ AlterCommand e[dit] Edit
...
 AlterCommand q[uit] Quit
+" TODO: Wq/Wqall need to be handled with the autocmd
```
ollien, line 55: Unfortunately, this means `:w` and `:wq` have different behavior. I'm hoping to resolve this before undrafting

runtime/vscode/overrides/vscode-file-commands.vim
ollien: Actually, looks like this check is unnecessary. With our current `wall` implementation, BufWriteCmd is not fired

src/buffer_manager.ts
```diff
@@ -410,6 +412,74 @@ export class BufferManager implements Disposable {
...
+        const filepath = await this.resolveWritePath(filename);
+        return document.uri.path !== filepath;
+    }
+
+    private async resolveWritePath(filename: string): Promise<string> {
+        if (path.isAbsolute(filename)) {
+            return filename;
+        }
+
+        const cwd = await actions.lua("cwd");
```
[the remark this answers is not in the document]
theol0403: Why can't we use the workspace CWD? Or use `workspace.fs.stat` to detect if there is a new file that needs writing (if the URI exists or not).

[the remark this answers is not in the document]
ollien: You're right. I had misread the code to mean that the workspace was the CWD (most of the time)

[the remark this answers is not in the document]
xiyaowong: Vim-cwd is not necessarily accurate. For example, when in remote development, but nvim runs locally.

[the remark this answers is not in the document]
xiyaowong: In addition, you can directly use client.call client.lua to call the function. Generally, only complex operations are encapsulated in "internal".

runtime/lua/vscode/files.lua
```diff
@@ -0,0 +1,23 @@
...
+function collect_file_info(event)
+  return {
+    buf = event.buf,
+    path = event.file,
+  }
+end
+
+function M.setup()
+  api.nvim_create_autocmd({ "BufWriteCmd" }, {
+    pattern = "*",
```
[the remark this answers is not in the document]
xiyaowong: It's ok. Our operation is very cheap.

runtime/lua/vscode/files.lua
```diff
@@ -0,0 +1,23 @@
...
+    callback = function(event)
+      local info = collect_file_info(event)
+      vscode.action("save_buf", { args = { info.buf, info.path } })
```
xiyaowong, line 18: Maybe it's better to send an object.

ollien: Yeah agreed. I just did this because it was consistent with other examples I saw.

src/buffer_manager.ts
```diff
@@ -410,6 +412,74 @@ export class BufferManager implements Disposable {
...
+            }
+
+            await this.client.lua(
+                `
+                local bufId = ...
+
+                vim.bo[bufId].modified = false
+            `,
+                [bufID],
+            );
```
xiyaowong, line 459: We should handle `workspace.onDidSaveTextDocument`

src/buffer_manager.ts
```diff
@@ -410,6 +412,74 @@ export class BufferManager implements Disposable {
...
+                `
+                local bufId, docUri, docUriJson, bufname = ...
+
+                vim.api.nvim_buf_set_var(bufId, "vscode_uri", docUri)
+                vim.api.nvim_buf_set_var(bufId, "vscode_uri_data", docUriJson)
+                vim.api.nvim_buf_set_name(bufId, bufname)
+                vim.bo[bufId].modified = false
+            `,
+                [bufID, document.uri.toString(), document.uri.toJSON(), bufname],
+            );
```
xiyaowong, line 445:
```suggestion
            await workspace.saveAs(document.uri);
```

We just need to call the method, the document management is handled by `syncEditorLayout`.

src/buffer_manager.ts
```diff
@@ -612,18 +675,21 @@ export class BufferManager implements Disposable {
...
                 bufId,
                 lines,
                 makeEditorOptionsVariable(editor?.options),
-                docUri.toString(),
-                docUri.toJSON(),
+                document.uri.toString(),
+                document.uri.toJSON(),
                 bufname,
                 this.isExternalTextDocument(document),
             ],
```
xiyaowong, line 695: Perhaps we can take this opportunity to encapsulate these contents into 'lua/vscode/internal.lua'.

runtime/lua/vscode/internal.lua
```diff
@@ -280,4 +280,10 @@ function M.wslpath(path)
   return vim.trim(ret)
 end
 
+-- Get the current Neovim CWD
+-- @return string
+function M.cwd()
+  return fn.getcwd()
```
justinmk, line 286: Looks like unnecessary indirection. Why can't callers use `fn.getcwd()` directly?

ollien: It was just to make use of the `internal` module with `actions.lua` As @xiyaowong It's not really needed

justinmk: that kind of thing will get copied and perpetuated. better to avoid it

xiyaowong:
@ollien If you don't mind, I might want to continue working on this PR.

ollien:
@xiyaowong go for it. I'm a bit tied up right now

PR #2081, opened 2024-06-06
fix: do not reset cmdline state unless we are going to show a dialog

### LINKED ISSUE #2079: bug: vim-sandwich stops working since 1.14
### Check the following:

### Neovim version (nvim -v)

0.9.1

### Operating system/version

Windows 7

### Describe the bug

I have been using https://github.com/machakann/vim-sandwich in neovim and vscode-neovim since day 1. It has been working perfectly. But it stops working since v1.14 (the last version it is still working is 1.13.1).

When try to delete / replace pair of parentheses, it just alter the texts outside of the selected scope.   

### Steps To Reproduce

1. Install vim-sandwich => `Plug 'machakann/vim-sandwich'`
2. Create a *.md file with 

```
"test_1"
"test_2"
"test_3"
```
3. Run `sr"'` in normal mode. It should replace `"` pair into `'` pair. 
4. Run `sd"` in normal mode. It should delete `"` pair

vim-sandwich

### Expected Behavior

Run `sr"'` in normal mode. It should replace `"` pair into `'` pair. 
Run `sd"` in normal mode. It should delete `"` pair

ollien:
Fixes #2079 

This one is a bit of a doozy. When using the sandwich plugin and running `sd"` on `master`, these are the logs from the CmdLine logger

```
[~/Documents/code/vscode-neovim] nick@scooty-puff-jr$ tail -f /tmp/vscode-neovim.log |grep CmdLine
2024-06-06T02:18:46.502Z CmdLine: cmdline_hide
2024-06-06T02:18:46.502Z CmdLine: visible level is undefined, hiding
2024-06-06T02:18:46.548Z CmdLine: cmdline_hide
2024-06-06T02:18:46.548Z CmdLine: visible level is undefined, hiding
2024-06-06T02:18:46.753Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-06T02:18:46.753Z CmdLine: cmdline_show: setting input value: "call setpos('.', a:tail)"
2024-06-06T02:18:46.755Z CmdLine: onChange: skip updating cmdline because change originates from nvim: ""
2024-06-06T02:18:46.755Z CmdLine: cmdline_hide
2024-06-06T02:18:46.755Z CmdLine: visible level is 1, hiding
2024-06-06T02:18:46.755Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-06T02:18:46.756Z CmdLine: cmdline_show: setting input value: "call setpos('.', a:tail)"
2024-06-06T02:18:46.759Z CmdLine: cmdline_hide
2024-06-06T02:18:46.759Z CmdLine: visible level is 1, hiding
2024-06-06T02:18:46.767Z CmdLine: onChange: skip updating cmdline because change originates from nvim: "call setpos('.', a:tail)"
2024-06-06T02:18:46.771Z CmdLine: onHide: skipping event
2024-06-06T02:18:46.771Z CmdLine: onChange: sending cmdline to nvim: "call setpos('.', a:tail)" + "<C-u>" -> ""
2024-06-06T02:18:46.775Z CmdLine: onChange: sending cmdline to nvim: "" + "<C-u>call setpos('.', a:tail)" -> "call setpos('.', a:tail)"
```

It seems that because the `onChange` events are not executing until after we process our events (which is fine, we have no guarantees there). Then, because the state was reset on every `msg_show`, the `pendingUpdatesFromNvim` count was reset, meaning that any text destined for the commandline was sent to the editor. This would happen even on events we would drop due to their content being unchanged

After the change, this looks something like this

```
2024-06-06T02:43:56.416Z CmdLine: cmdline_hide
2024-06-06T02:43:56.416Z CmdLine: visible level is undefined, hiding
2024-06-06T02:43:56.416Z CmdLine: cmdline_hide
2024-06-06T02:43:56.416Z CmdLine: visible level is undefined, hiding
2024-06-06T02:43:56.622Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-06T02:43:56.623Z CmdLine: cmdline_show: setting input value: "call setpos('.', a:tail)"
2024-06-06T02:43:56.630Z CmdLine: cmdline_hide
2024-06-06T02:43:56.630Z CmdLine: visible level is 1, hiding
2024-06-06T02:43:56.631Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-06T02:43:56.631Z CmdLine: dropping cmdline_show as the content is unchanged
2024-06-06T02:43:56.632Z CmdLine: cmdline_hide
2024-06-06T02:43:56.632Z CmdLine: visible level is 1, hiding
2024-06-06T02:43:56.651Z CmdLine: onChange: skip updating cmdline because change originates from nvim: "call setpos('.', a:tail)"
2024-06-06T02:43:56.654Z CmdLine: onHide: skipping event
```

The old code actually called `reset()` when the dialog itself hid, but that is also racy. By resetting right before we show a new dialog, we ensure we have no stale state due to out of order event processing.

~~I'd love to add a test for this, but if I'm honest, I'm not sure how to trigger this bug without this plugin (which seems to be emulating some fast typing). I'll need to think on it~~

### DIFF
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -116,13 +116,18 @@ export class CommandLineManager implements Disposable {
 
         this.state.redrawExpected = false;
         const title = prompt || this.getTitle(firstc);
+        const initialInputValue = this.input.value;
         const visibilityChanged = this.showInput(level, title, content);
 
         // We only need to proceed if text has been entered. In both of these cases, they haven't.
         //
-        // 1. The dialog box is being shown for the first time, and there is no content. No one has typed anything!
+        // 1. The dialog box is being shown for the first time, and the content is the same as it was before we set
+        //    the initial value. This is basically trying to get ahead of an onChange event.
         // 2. The dialog box was already visible, but the content is identical. No one has typed anything!
-        if ((visibilityChanged && content === "") || (!visibilityChanged && this.input.value === content)) {
+        if (
+            (visibilityChanged && initialInputValue === content) ||
+            (!visibilityChanged && this.input.value === content)
+        ) {
             logger.debug("dropping cmdline_show as the content is unchanged");
             return;
         }
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -72,14 +72,6 @@ export class CommandLineManager implements Disposable {
         disposeAll(this.disposables);
     }
 
-    private reset() {
-        this.state = new CmdlineState();
-        this.input.value = "";
-        this.input.title = "";
-        this.input.items = [];
-        this.input.activeItems = [];
-    }
-
     private handleRedraw({ name, args }: EventBusData<"redraw">) {
         switch (name) {
             case "cmdline_show": {
@@ -116,29 +108,25 @@ export class CommandLineManager implements Disposable {
     }
 
     private cmdlineShow = (content: string, firstc: string, prompt: string, level: number): void => {
-        this.state.level = level;
-        this.input.title = prompt || this.getTitle(firstc);
         // only redraw if triggered from a known keybinding. Otherwise, delayed nvim cmdline_show could replace fast typing.
         if (!this.state.redrawExpected) {
             logger.debug(`cmdline_show: ignoring cmdline_show because no redraw expected: "${content}"`);
             return;
         }
 
-        this.state.redrawExpected = false;
+        const title = prompt || this.getTitle(firstc);
+        const visibilityChanged = this.showInput(level, title, content);
 
-        if (this.input.value === content) {
+        // We only need to proceed if text has been entered. In both of these cases, they haven't.
+        //
+        // 1. The dialog box is being shown for the first time, and there is no content. No one has typed anything!
+        // 2. The dialog box was already visible, but the content is identical. No one has typed anything!
+        if ((visibilityChanged && content === "") || (!visibilityChanged && this.input.value === content)) {
             logger.debug("dropping cmdline_show as the content is unchanged");
             return;
         }
 
-        if (!this.isVisible()) {
-            // Reset the state if this is a new dialog
-            logger.debug("displaying a new dialog, resetting state");
-            this.reset();
-        }
-
-        this.showInput();
-        logger.debug(`cmdline_show: setting input value: "${content}"`);
+        logger.debug(`cmdline_show: setting input value: "${content}", with level ${level}`);
         this.state.lastTypedText = content;
         this.state.pendingNvimUpdates++;
         const activeItems = this.input.activeItems; // backup selections
@@ -147,8 +135,6 @@ export class CommandLineManager implements Disposable {
     };
 
     private cmdlineHide() {
-        // The hide originated from neovim, so we don't need to listen for the hide event from the quickpick
-        this.state.ignoreHideEvent = true;
         // We expect that a cmdline_show may come through to draw this editor a second time (e.g. when level changes)
         this.state.redrawExpected = true;
 
@@ -156,6 +142,9 @@ export class CommandLineManager implements Disposable {
         // (or, defensively, if we already should be hidden)
         if (this.state.level === 1 || !this.isVisible()) {
             logger.debug(`visible level is ${this.state.level}, hiding`);
+            // The hide originated from neovim, so we don't need to listen for the hide event from the quickpick
+            this.state.ignoreHideEvent = true;
+
             this.hideInput();
         } else {
             logger.debug(`visible level is ${this.state.level}, not hiding`);
@@ -242,8 +231,32 @@ export class CommandLineManager implements Disposable {
         }
     }
 
-    private showInput() {
+    /**
+     * Ensure the input box is showing.
+     *
+     * @param level The level of this input
+     * @param title  The title to use for this input
+     * @param initialValue The value that the input should have when it is opened. If the input is already opened, this has noe ffect
+     * @returns Whether or not the visibility of the input changed
+     */
+    private showInput(level: number, title: string, initialValue: string): boolean {
+        let visibilityChanged = false;
+        // For first appearances, reset all state
+        // NOTE: we do not set the value to an empty string here as we do not want to trigger an onChange by
+        // setting it to an empty string, and then some other value.
+        if (!this.isVisible()) {
+            this.state = new CmdlineState();
+            this.input.items = [];
+            this.input.activeItems = [];
+            this.input.value = initialValue;
+            visibilityChanged = true;
+        }
+
+        this.state.level = level;
+        this.input.title = title;
         this.input.show();
+
+        return visibilityChanged;
     }
 
     private hideInput() {
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -128,6 +128,12 @@ export class CommandLineManager implements Disposable {
             (visibilityChanged && initialInputValue === content) ||
             (!visibilityChanged && this.input.value === content)
         ) {
+            // A change event will be fired for the new content in this box, given we have to show the box
+            // before we set the value
+            if (visibilityChanged && content.length > 0) {
+                this.state.pendingNvimUpdates++;
+            }
+
             logger.debug("dropping cmdline_show as the content is unchanged");
             return;
         }
@@ -254,13 +260,16 @@ export class CommandLineManager implements Disposable {
             this.state = new CmdlineState();
             this.input.items = [];
             this.input.activeItems = [];
+            // Show MUST be called before we set the value, hence this branching
+            this.input.show();
             this.input.value = initialValue;
             visibilityChanged = true;
+        } else {
+            this.input.show();
         }
 
         this.state.level = level;
         this.input.title = title;
-        this.input.show();
 
         return visibilityChanged;
     }
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -116,11 +116,6 @@ export class CommandLineManager implements Disposable {
     }
 
     private cmdlineShow = (content: string, firstc: string, prompt: string, level: number): void => {
-        if (!this.isVisible()) {
-            // Reset the state if this is a new dialog
-            this.reset();
-        }
-
         this.state.level = level;
         this.input.title = prompt || this.getTitle(firstc);
         // only redraw if triggered from a known keybinding. Otherwise, delayed nvim cmdline_show could replace fast typing.
@@ -128,16 +123,27 @@ export class CommandLineManager implements Disposable {
             logger.debug(`cmdline_show: ignoring cmdline_show because no redraw expected: "${content}"`);
             return;
         }
+
         this.state.redrawExpected = false;
-        this.showInput();
-        if (this.input.value !== content) {
-            logger.debug(`cmdline_show: setting input value: "${content}"`);
-            this.state.lastTypedText = content;
-            this.state.pendingNvimUpdates++;
-            const activeItems = this.input.activeItems; // backup selections
-            this.input.value = content; // update content
-            this.input.activeItems = activeItems; // restore selections
+
+        if (this.input.value === content) {
+            logger.debug("dropping cmdline_show as the content is unchanged");
+            return;
         }
+
+        if (!this.isVisible()) {
+            // Reset the state if this is a new dialog
+            logger.debug("displaying a new dialog, resetting state");
+            this.reset();
+        }
+
+        this.showInput();
+        logger.debug(`cmdline_show: setting input value: "${content}"`);
+        this.state.lastTypedText = content;
+        this.state.pendingNvimUpdates++;
+        const activeItems = this.input.activeItems; // backup selections
+        this.input.value = content; // update content
+        this.input.activeItems = activeItems; // restore selections
     };
 
     private cmdlineHide() {
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -22,9 +22,6 @@ class CmdlineState {
     // If this value is undefined, the input is not visible
     level?: number = undefined;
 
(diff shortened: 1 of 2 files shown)


theol0403:
The commandline isn't opening for me when I test this: likely because `show` needs to be called even on matching content (because what if the cmdline is init for "").

ollien:
You're so right. One of my earlier reshuffles of code worked fine here, but I had only tested the sandwich plugin itself. I'm surprised the tests pass.

ollien:
Hm, I was trying to put together a test for this, but I'm struggling a bit. For the `sd"` command, sandwich executes [the following](https://github.com/machakann/vim-sandwich/blob/3a7cffd07ee1c762e3abafe1ce220fc8096f9a75/autoload/operator/sandwich/act.vim#L475-L481)

```vim
function! s:delete_portion(head, tail, undojoin_cmd) abort  "{{{
  let cmd = "%ssilent noautocmd normal! \"\"dv:call setpos('\.', %s)\<CR>"
  call setpos('.', a:head)
  let @@ = ''
  execute printf(cmd, a:undojoin_cmd, 'a:tail')
  return @@
endfunction
```

I tried to write another similar motion locally to see if I could reproduce the bug, but `:nnoremap dq :<C-u>normal! $F"dl<CR>:normal!^f"dl<CR>` doesn't seem to do it. Something special must be going on here, but I'm not sure what. I may forgo the test if it doesn't come ot me soon

src/cmdline_manager.ts
```diff
@@ -116,40 +108,44 @@ export class CommandLineManager implements Disposable {
...
-            this.input.value = content; // update content
-            this.input.activeItems = activeItems; // restore selections
+        const title = prompt || this.getTitle(firstc);
+        const visibilityChanged = this.showInput(level, title, content);
+
+        // We only need to proceed if text has been entered. In both of these cases, they haven't.
+        //
+        // 1. The dialog box is being shown for the first time, and there is no content. No one has typed anything!
+        // 2. The dialog box was already visible, but the content is identical. No one has typed anything!
+        if ((visibilityChanged && content === "") || (!visibilityChanged && this.input.value === content)) {
```
theol0403, line 125: why isn't this handled by just `this.input.value === content`? When shown for the first time, this is true since both the value and the content are ""

ollien: Aha! This is subtle. Check out the log of running `sr'"` both with/without the check for `visibilityChanged`

Without:

```
2024-06-07T13:55:34.854Z CmdLine: cmdline_hide
2024-06-07T13:55:34.854Z CmdLine: visible level is undefined, hiding
2024-06-07T13:55:34.855Z CmdLine: cmdline_hide
2024-06-07T13:55:34.855Z CmdLine: visible level is undefined, hiding
2024-06-07T13:55:36.414Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-07T13:55:36.414Z CmdLine: dropping cmdline_show as the content is unchanged
2024-06-07T13:55:36.421Z CmdLine: cmdline_hide
2024-06-07T13:55:36.421Z CmdLine: visible level is 1, hiding
2024-06-07T13:55:36.421Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-07T13:55:36.422Z CmdLine: dropping cmdline_show as the content is unchanged
2024-06-07T13:55:36.424Z CmdLine: cmdline_hide
2024-06-07T13:55:36.424Z CmdLine: visible level is 1, hiding
2024-06-07T13:55:36.428Z CmdLine: onChange: sending cmdline to nvim: "" + "<C-u>call setpos('.', a:tail)" -> "call setpos('.', a:tail)"
2024-06-07T13:55:36.470Z CmdLine: onHide: skipping event
```

With:

```
2024-06-07T13:56:20.232Z CmdLine: cmdline_hide
2024-06-07T13:56:20.232Z CmdLine: visible level is undefined, hiding
2024-06-07T13:56:20.790Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-07T13:56:20.790Z CmdLine: cmdline_show: setting input value: "call setpos('.', a:tail)", with level 1
2024-06-07T13:56:20.794Z CmdLine: cmdline_hide
2024-06-07T13:56:20.794Z CmdLine: visible level is 1, hiding
2024-06-07T13:56:20.794Z CmdLine: cmdline_show: "0,call setpos('.', a:tail)"
2024-06-07T13:56:20.794Z CmdLine: cmdline_show: setting input value: "call setpos('.', a:tail)", with level 1
2024-06-07T13:56:20.797Z CmdLine: cmdline_hide
2024-06-07T13:56:20.798Z CmdLine: visible level is 1, hiding
2024-06-07T13:56:20.812Z CmdLine: onHide: skipping event
```	

You'll notice that in the second, we don't ignore the input value in this incredibly fast typing case (i.e. send by the plugin). This means that `pendingNvimChanges` isn't incremented, thus causing us to have an onChange event that gets sent to the editor.

ollien: Reading this comment over, I realize it's kinda unclear what's going on unless you've been in the trenches like I have been. Let me see if I can distill this a bit by speaking event-wise.

In the case `vim-sandwich` provides us, it executes `:setpos('.', a:tail)`. Consider the following events, if the check were just the equality check you propose, just after the extension loads (i.e. the command box has not already been used)

1. The plugin emits `:setpos('.', a:tail)`
2. We drop the event, as the initial content (`:setpos('.', a:tail)`) equals the content we get from nvim (`:setpos('.', a:tail)`). This leaves `pendingNvimEvents at zero`
3. VSCode sees a change from `""` to `":setpos('.', a:tail)"`, so it types this text.
4. We run `sr'"` again. 
5.  We drop the event, as the initial content (`:setpos('.', a:tail)`) equals the content we get from nvim (`:setpos('.', a:tail)`). This leaves `pendingNvimEvents at zero`
6. VSCode does *NOT* see a change (the input box holds `":setpos('.', a:tail)"` and we just set its content to the same), so it does not type to the editor. Thus, the command works fine.

We can also consider the case where we open a new window with the empty string.

1. We type `:`, the box opens with `value` of `""`
2. VSCode will not emit an onChange event, as we went from `""` to `""`
3. We type some text (say, `asdf`), then pres esc.
4. We type `:`, the box opens with a `value` of "".
5. This _does_ fire an onChange event, as we go from `"asdf"` to `""`. However, there's nothing for VSCode to send to the editor here, (as `lastTypedText` has been reset to `""`).

Honestly, after writing all this out, part of me wants to revert to the paradigm where we created a fully new input box for every instance of the command line. However, I have throughly tested these changes and I believe them to be correct. I might do that as a follow-on PR, but the next few days are busy for me so I won't have a chance to commit to it at a level I feel confident about.

ollien: I found some spare time for the alternate implementation, but I haven't fully tested it. Let me know what you prefer

#2089

ollien:
Added a test to demonstrate the validity of [66f192e](https://github.com/vscode-neovim/vscode-neovim/pull/2081/commits/66f192e5bba54da0575b9a948d28c0859cfe6205), and also added a driveby bug fix for something I noticed when writing that test (it's an edge case, and is present in master, but it was worth doing)

xiyaowong:
I don't pay attention to cmdline.

theol0403 (dismissed):
this.input.show() has to be called before setting the text or else we preselect the initial text (like when coming from visual mode)

ollien:
> this.input.show() has to be called before setting the text or else we preselect the initial text (like when coming from visual mode)

Oof. Thanks.

Just out of curiosity, are these quirks documented anywhere? I'm guessing not, but I figure I'd ask

ollien:
I hate how fiddly this API is... moving `show()` before setting the value causes the original bug to reappear.

ollien:
Wrote a small test for this, but unfortunately due to how input is handled for our spawned VSCode instance, it doesn't work like I'd hoped :(

Posting it here for posterity slash if anyone sees a way to make it work

```ts
diff --git a/src/test/integ/command-line.test.ts b/src/test/integ/command-line.test.ts
index 01a28fc..b8de539 100644
--- a/src/test/integ/command-line.test.ts
+++ b/src/test/integ/command-line.test.ts
@@ -221,4 +221,38 @@ describe("Command line", () => {
             client,
         );
     });
+
+    it("Should not select the text in the input when entering from visual mode with selected text", async () => {
+        await openTextDocument({ content: "hello, world\n".repeat(10) });
+
+        await sendVSCodeKeys(":3");
+        await sendVSCodeCommand("vscode-neovim.commit-cmdline");
+        await sendVSCodeKeys("Vjj");
+        // when we press : in visual mode, we should have '<,'> in the command box. If the text were selected,
+        // then this would overwrite that and we would substitute on only one line, not the three selected lines
+        //
+        // Also tricky: this cannot use "test-cmdline" for typing the first 's', otherwise the selection will
+        // not have an effect on our typing
+        await sendVSCodeKeys(":s");
+        await sendVSCodeCommand("vscode-neovim.test-cmdline", "/, world/ there");
+        await sendVSCodeCommand("vscode-neovim.commit-cmdline");
+        await assertContent(
+            {
+                content: [
+                    "hello, world",
+                    "hello, world",
+                    "hello there",
+                    "hello there",
+                    "hello there",
+                    "hello, world",
+                    "hello, world",
+                    "hello, world",
+                    "hello, world",
+                    "hello, world",
+                    "",
+                ],
+            },
+            client,
+        );
+    });
 });
```

theol0403:
Not possible to test as we do not have "puppeteer-like" control over the vscode interface. The selection of the quickpick is not something we have any visibility into.

theol0403 (commented):
The "cancel" button is no longer working

ollien:
> The "cancel" button is no longer working, neither is pressing Esc on an empty commandline

Already on it. I didn't wanna run the tests locally so I figured I'd gamble it :)

ollien:
> The "cancel" button is no longer working, neither is pressing Esc on an empty commandline
Actually,  I take it back. Neither of these specifically are happening for me. Are you just doing `:<Esc>`, and ending up with the command line still up?

theol0403:
Reproduce:
* type :
* type Esc
* type :
* type Esc
* nvim is still in cmdline mode
* type anything and it will cause the cmdline to show up again

ollien:
Wow, that's nasty. It looks like we're racing with `onHide` and `cmdline_hide`, causing the `ignoreHideEvent` flag to be sent longer than we'd like

```2024-06-10T01:29:23.784Z CmdLine: cmdline_show: "0,"
2024-06-10T01:29:23.785Z CmdLine: dropping cmdline_show as the content is unchanged
2024-06-10T01:29:25.193Z CmdLine: onHide: entering <ESC> <--- I pressed ESC here
2024-06-10T01:29:25.196Z CmdLine: cmdline_hide <--- flag set here
2024-06-10T01:29:25.197Z CmdLine: visible level is 1, hiding
2024-06-10T01:29:33.099Z CmdLine: cmdline_show: "0,"
2024-06-10T01:29:33.100Z CmdLine: dropping cmdline_show as the content is unchanged
2024-06-10T01:29:34.414Z CmdLine: onHide: skipping event <--- now we skip the event here
```

I'm sorry that this PR has turned into a Pandora's box of sorts, but I really appreciate the time you've taken to help me find the faults

xiyaowong:
Speaking as an observer: Implementing statusBar as cmdline still seems necessary.

theol0403:
Perhaps a new, first-principles approach should be considered.

Q: Why can't the VSCode UI simply reflect the cmdline state reported from nvim, with NO internal state, literally as a render view?
A: As with all two-way syncs, there needs to be debouncing. When typing, nvim should not interfere by rendering. When using bindings/completions, vscode shouldn't send events back to nvim.
A: State transitions are reported by both vscode and nvim, and either one can trigger the other. 

What if there was a reliable way of determining whether an event was triggered on the vscode side? For example, onChanged but ignored when it is triggered by nvim. Actually that's exactly what `pendingNvimUpdates` is. 

I didn't take the time to understand what the issue with sandwich is, but it seems strange to me that there is a fundamental issue with the current simplistic approach of simply rendering what nvim displays and using some logic to avoid 2-way conflicts. 

Not sure if this is contributing anything.

ollien:
> I didn't take the time to understand what the issue with sandwich is, but it seems strange to me that there is a fundamental issue with the current simplistic approach of simply rendering what nvim displays and using some logic to avoid 2-way conflicts.

Yeah, sorry for the wall-of-text I gave you. It's really just one of those enigmas.

I think the TL;DR is that the quickpick events do not fire until we yield the event loop back to VSCode, which is the heart of this problem. My thought, which I need to take the time to flesh out, is we need to really classify the code into two paths

1) changes originating from nvim, including all `cmdline_show`s and `cmdline_hide`s. 
2) changes originating from a user (e.g. `onHide`, etc).

Right now, we rely on `onChange` for _all_ updates to the command box, but given enough events, quickly enough, we will not be yielding back to VSCode, and those `onChange` events do not process until we finish processing our nvim events.

ollien:
I'm thinking out loud a bit here, but what if we generated an event queue within the `CommandlineManager`? The idea is that if we have a "slice" of cmdline show/hides all in one batch, like this, we can break them up into two groups, delimiated by the cmdline_hide


```
cmdline_show #begin group 1
cmdline_show a
cmdline_show ab
cmdline_show abc
cmdline_hide # end group 1
cmdline_show #begin group 2
cmdline_show
...
```

We could process `cmdline_show` events as normal, but if we get any `cmdline_show`s before onHide executes, they are instead adding to a `pending` queue. When `onHide` executes, it will "re-emit" the events to itself. In this example, we would run the `cmdline_show`s for group 1 as we do now, but once we execute the `cmdline_hide`, completely stop processing events, and put the remaining items into our `pending` queue. Once `onHide` fires, we begin processing the group 2.

Basically, we need some way to delimit what quickpick instance the events belong to, but because we can't just create new QuickPick instances every time (due to the races that necessitated that old timer hack), we can stage them like this.

ollien:
Closing in favor of new PR

Issue #1752, opened 2024-01-17
[Bug] `ci'`, cursor jumps to the end of line


wenfangdu:
OS: macOS 14
nvim: 0.9.5
vscode-nvim: 1.5.0

vscode-neovim behavior (cursor jumps to the end of the line):

https://github.com/vscode-neovim/vscode-neovim/assets/28700378/238e82cd-8bfc-4e7b-a25d-1cc9b2758bde

Neovim behavior (cursor stays inside the quotes):

https://github.com/vscode-neovim/vscode-neovim/assets/28700378/b71d0191-ddeb-4c1e-8063-e4bf10b39f83

Code:

```styl
$base = '../rewards'
$bgi = $proxy($base, 'bgi')

@import '../includes/highlight.styl'

.open-task-btn
  $bgi '../static/rewards/open-tasks-btn-double.png'
  $highlight '../static/rewards/open-tasks-btn-double.png'
```

xiyaowong:
#1080 Still can't reproduce

wenfangdu:
@xiyaowong I tried disabling all extensions except vscode-neovim, but the issue still persists.

xiyaowong:
OS: windows 11
All are the latest versions

KeqiZeng:
Facing same issue on MacOS 14.2.1

wenfangdu:
@xiyaowong Confirmed, both #1752 and #1080 are not reproducible on Windows 11 but reproducible on macOS 14, all other related apps are the latest.

ollien:
FWIW, I've tried to replicate this on Linux, and could not. I guess it's macOS specific?

wenfangdu:
@ollien Thanks for the experiment, looks like it is.

xiyaowong:
Please confirm if the issue still exists

wenfangdu:
@xiyaowong Still exists.

wenfangdu:
@xiyaowong I just upgraded  vscode-nvim to 1.14.1, and it's still reproducible with `ci'`, please reopen this issue:

https://github.com/vscode-neovim/vscode-neovim/assets/28700378/251235e3-2b21-4465-b4b7-6d0a11da3577

xiyaowong:
Could you provide more logs and a complete screen recording? When does the issue occur, and when does it function normally? A single screen recording like this is not helpful at all.

wenfangdu:
[...]

PR #2082, opened 2024-06-06
fix: remove maximum debounce time for normal selection

wenfangdu:
I found `"vscode-neovim.normalSelectionDebounceTime": 100` is not enough for completely getting rid of https://github.com/vscode-neovim/vscode-neovim/issues/1752.
`200` is the max limit for `vscode-neovim.mouseSelectionDebounceTime` too, might be sufficient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -320,7 +320,7 @@
                 "vscode-neovim.normalSelectionDebounceTime": {
                     "type": "integer",
                     "minimum": 20,
-                    "maximum": 100,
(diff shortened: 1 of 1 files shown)


xiyaowong (changes requested):
Just delete the maximum value limit directly.

wenfangdu:
@xiyaowong Removed.

xiyaowong (approved):

PR #2087, opened 2024-06-07
build: upgrade to ES2023

### LINKED ISSUE #2086: Upgrade TS libs to ES2023
Now that VSCode 1.90 is released ([which uses Electron 29 and Node 20](https://code.visualstudio.com/updates/v1_90#_electron-29-update)), we should take the opportunity to upgrade our Typescript `lib` to ES2023. While we probably won't use most of the features (perhaps a few of the new array functions like `findLast`), we should strive to keep this up to date to enable better DX.

ollien:
Closes #2086 

We should probably wait to merge this until 1.90 has been out a little longer, bur I wanted to put this up there so we could merge it at some other time. If there's anywhere else you think ES2023 could be put into this project, please feel free to push them to this branch :)


justinmk (approved):

ollien:
@theol0403 Thanks for fixing that conflict; I haven't had a chance to catch up here in a bit. Given we've also just bumped the project to requiring 1.90, I'm going to merge this.

PR #2089, opened 2024-06-08
fix: fully reset cmdline state every time we hide it

ollien:
This is an alternate implementation of #2081. I haven't put it thorugh its paces, but I wanted to throw it together as a proof of concept.

The main idea here is that a lot of the bugginess that #2081 aimed to fix was a result of persisting state between instances of showing the cmdline. While we can work around this, the logic to work around this was hard to understand. Instead, we just fully dispose of all state, including the input itself, every time we get a `cmdline_hide` that would result in the input box going away. 

By factoring it in this way, we fully eliminate an entire class of bug, where details are persisted between two instances of the command line input. It also pretty clearly separates the "state for neovim" and the "state for the input"

### DIFF
--- a/src/cmdline/cmdline_input.ts
+++ b/src/cmdline/cmdline_input.ts
@@ -1,0 +1,222 @@
+import { QuickPick, QuickPickItem, ThemeIcon, window, Disposable, QuickInputButton } from "vscode";
+
+import { disposeAll } from "../utils";
+import { calculateInputAfterTextChange } from "../utils/cmdline_text";
+import { createLogger } from "../logger";
+
+const logger = createLogger("CmdLineInput", false);
+
+export interface CmdlineInputListeners {
+    onAccept: () => void;
+    onChangeValue: (value: string, toInput: string) => void;
+    onChangeSelection: (selectionIndex: number) => void;
+    onHide: () => void;
+}
+
+/**
+ * The input box used for the commandline input.
+ */
+export class CmdlineInput implements Disposable {
+    private disposables: Disposable[] = [];
+
+    private input: QuickPick<QuickPickItem>;
+
+    // The last text typed in the UI, used to calculate changes
+    private lastTypedText: string = "";
+
+    // The current "level" of the cmdline we show. :help ui describes this as
+    //  > The Nvim command line can be invoked recursively, for instance by typing <c-r>= at the command line prompt.
+    //  > The level field is used to distinguish different command lines active at the same time. The first invoked
+    //  > command line has level 1, the next recursively-invoked prompt has level 2. A command line invoked from the
+    //  > cmdline-window has a higher level than the edited command line.
+    //
+    // If this value is undefined, the input is not visible
+    private level?: number = undefined;
+
+    // When updates come from nvim, we write to the input field.
+    // We don't want to send those updates back to nvim, so we use this counter to keep track of the number of onChange to ignore.
+    private pendingNvimUpdates = 0;
+
+    private disposed: boolean = false;
+
+    private listeners: CmdlineInputListeners;
+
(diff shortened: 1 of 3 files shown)


theol0403:
Not sure if this still applies but a workaround/hack I was able to remove from the old code was 

https://github.com/vscode-neovim/vscode-neovim/blob/78cf093f51a2c6b56658ca03b0a25366489c73d5/src/cmdline_manager.ts#L44

src/cmdline/cmdline_input.ts
```diff
@@ -0,0 +1,237 @@
+import { QuickPick, QuickPickItem, ThemeIcon, window, Disposable, QuickInputButton } from "vscode";
+
+import { disposeAll } from "../utils";
+import { calculateInputAfterTextChange } from "../utils/cmdline_text";
+import { createLogger } from "../logger";
+
+const logger = createLogger("CmdLineInput", false);
+
+export interface CmdlineInputListeners {
+    onAccept: () => void;
```
theol0403, line 10: You don't need this callback bolierplate. 


[`19e1412` (#2001)](https://github.com/vscode-neovim/vscode-neovim/pull/2001/commits/19e1412c4942f2b0eac9136f77329a5fc6e44e5a)

ollien:
Oh interesting... is the idea that reconstructing the input itself races with our event handling? If so, that's really unfortunate and I will close this PR.

src/cmdline/cmdline_input.ts
```diff
@@ -0,0 +1,237 @@
...
+export interface VisibilitySome {
+    visible: true;
+    level: number;
+}
+
+export interface VisibilityNone {
+    visible: false;
+}
+
+export type Visibility = VisibilitySome | VisibilityNone;
```
theol0403, line 25: Can't level just be 0 to indicate invisible? What about number | undefined?

ollien: I could. It's a personal preference of mine to have the types be a bit explicit, but if you prefer `number?`, we can do that

theol0403:
> Oh interesting... is the idea that reconstructing the input itself races with our event handling? If so, that's really unfortunate and I will close this PR.

Not sure, that was from before my time.

src/cmdline/cmdline_input.ts
ollien: You're right, you don't need it. I did this so we could decouple the interactions with neovim from the input box itself

ollien:
> > Oh interesting... is the idea that reconstructing the input itself races with our event handling? If so, that's really unfortunate and I will close this PR.
> 
> Not sure, that was from before my time.

Hm. I'll bang on it a bit

I'm currently facing some bugs with quickpick selections not working ... at all? I doubt it's related but that's what I'm staring at right now :)

src/cmdline/cmdline_input.ts
theol0403: Reduce the volume of code a reader has to grok. An optional is a known pattern.

theol0403:
> I doubt it's related but that's what I'm staring at right now :)

Quickpick selections was tricky. You gotta call show() for every update, you gotta save and restore the selection while typing, etc.

ollien:
> > Oh interesting... is the idea that reconstructing the input itself races with our event handling? If so, that's really unfortunate and I will close this PR.
> 
> Not sure, that was from before my time.

I have a theory. Given that code was written when we were still processing entire batches of commands at once, the old code may not have had a chance to yield the event loop to VSCode, but now we do (given we handle each event individually). I don't have any proof, but it's a thought

ollien:
The above timer concern is enough for me to think twice about this PR. I'm pretty sure I'm reintroducing the necessity for the timer hack :(

PR #2090, opened 2024-06-08
refactor: Add highlight group manager

xiyaowong:


src/highlights/highlight_grid.ts
```diff
@@ -1,152 +1,252 @@
...
+    // #endregion
+
+    // #region Calculate Decorations
+
+    // decoration type -> decoration options
+    public getDecorations(
+        editor: TextEditor,
+        startLine: number,
+        endLine: number,
+    ): Map<TextEditorDecorationType, DecorationOptions[]> {
```
ollien, line 95: Part of the idea of `HighlightGrid` was we could completely decouple the production of the highlighted regions from the decorations (which gave us the ability to unit test this). Do you plan on re-adding these tests for this new API?

ollien:
What's "responsible" here? I'm not sure what the goal of this PR is at a quick glance

src/highlights/highlight_grid.ts
xiyaowong: Maybe. It's not easy to add tests.

ollien: We already had a bunch of them though, and with this restructure, they're all gone. I don't mind changing up the API, but removing the tests entirely removes a lot of the safety around changing this algorithm

xiyaowong:
Not Ready for review

src/test/integ/highlight-grid.test.ts
```diff
@@ -0,0 +1,471 @@
...
+            line: number;
+            vimCol: number;
+            vimCells: VimCell[];
+            lineText: string;
+            tabSize: number;
+        }[];
+        expectedRanges: HighlightRange[];
+    }[];
+
+    testCases.forEach(({ testName, events, expectedRanges }) => {
```
ollien, line 445: These tests don't use the spawned VSCode instance - why move them from unit tests?

src/test/integ/highlight-grid.test.ts
```diff
@@ -0,0 +1,471 @@
...
+        if (testName !== "allows overlaying virtual text on an existing line") return;
```
ollien, line 446: Left over debugging measure?

xiyaowong: 😂

src/test/integ/highlight-grid.test.ts
```diff
@@ -0,0 +1,471 @@
...
+        it(testName, () => {
+            const grid = new HighlightGrid(
+                1, // arbitrary
+                // fake the group store, only for testing
+                { normalizeHighlightId: (hlId) => hlId } as HighlightGroupStore,
+            ) as any as THighlightGrid;
```
ollien, line 452: What is this cast for?

xiyaowong: No error when accessing private methods.

src/highlights/highlight_grid.ts
```diff
@@ -1,152 +1,323 @@
...
+    // In the next refresh, if a decorator is no longer used, it should be cleared.
+    private prevDecorators: Set<TextEditorDecorationType> = new Set();
+    // Flag to indicate if the grid needs to be redrawn
+    private isDirty = false;
+
+    // Used to get the editor and viewport
+    private readonly gridId: number;
+    // Normalizes highlight IDs and provides decorators
+    private readonly groupStore: HighlightGroupStore;
+    // Not available when testing, so optional
```
ollien, line 56: This, in particular, I think is a smell we should address. `HighlightGrid` was able to be fully unit tested before, because it didn't have any dependencies on our VSCode state; it was passed in as needed. As far as I can tell, these two properties are only necessary because decorations have been merged into `HighlightGrid`, which is a responsibility we can delegate either to the `Manager` or the (now removed) `Provider` like before.

src/test/integ/highlight-grid.test.ts
xiyaowong: I understand your thinking. I also anticipated that this question might be raised here. I really don't think it's necessary to separate the logic just for testing purposes; testing it directly in VSCode is also very quick. Moreover, since this is a VSCode extension, there's no need to get bogged down with this issue.

src/test/integ/highlight-grid.test.ts
ollien: Why do we need to access private methods in a test? This is a sign we should factor a code in the way that is more usable from our tests. I don't mean to suggest that unit tests are a panacea, or a perfect solution, but it is a method that I put in place primarily due to the difficulty of doing a full integration test. If we use them (which we do not have to, I just want a way to test highlights), we should ensure that we don't have to hack around our code in order to test it.

src/test/integ/highlight-grid.test.ts
ollien: Understood. Frankly, on my own, I was always quite frustrated with the limitation we had of not being able to use VSCode imports.

I don't think we necessarily need to factor our code _just_ for testing, but I do think we should separate out tests that are manipulating inputs, and those that are calling code. They're ultimately very different kinds of tests, and if for no other reason other than organization, would be a good idea to do.

src/highlights/highlight_grid.ts
xiyaowong: Alright, If it weren't for testing purposes, the code could be further simplified a lot.

I'm thinking of splitting it into separate files, like calculations.ts. What do you think?

xiyaowong:
It seems like we always run into conflicts on code organization. Sometimes, being too strict with standards can actually become limiting.

src/highlights/highlight_grid.ts
ollien: :100: 

I think we're on the same page. This is what I had hoped to do with `HighlightGrid` originally (it was a data structure which could maintain our highlight calculations, and could then be used by the `Provider` to produce decorations), and I think if we continue to maintain that separation, we will avoid complexity to just facilitate testing.

BTW, I do intend to review more than just the testing logic, I just haven't gotten to it yet :) I'm very interested in this approach

ollien:
> It seems like we always run into conflicts on code organization. Sometimes, being too strict with standards can actually become limiting.

FWIW, I don't want to be strict here, I think we just have different opinions :). In this particular case, this is less of an organization concern, and more of a way we can factor things to better promote maintenance. I'm sensitive to it with highlights in particular because the logic is so complicated, and easy to break in subtle ways. Some others, I care less about

src/highlights/highlight_grid.ts
```diff
@@ -1,152 +1,323 @@
...
+                    redrawCells.push({ text, hlId: currHlId });
+                }
+            }
+        }
+        const leftCells = prevCells.slice(0, vimCol);
+        const rightCells = prevCells.slice(vimCol + redrawCells.length);
 
-    constructor() {
-        this.grid = [];
+        this.lineCells[line] = [...leftCells, ...redrawCells, ...rightCells];
```
ollien, line 109: I like how this all comes together, here. Nice!

xiyaowong:
I'm actually torn too, because I know your reasons are more valid than mine. I'm in the process of refactoring some logic. (I'm not very fond of splitting up the logic.)

src/highlights/highlight_grid.ts
xiyaowong: Splitting the logic for testing purposes has ironically made the logic more complicated, and I'm really conflicted about it.

ollien: It's a give and take, for sure. I'm confident you'll find a good middle ground :)

xiyaowong: I'm doing something I don't like doing.

ollien: If there's any way I can help, do let me know

xiyaowong:
Done!

src/highlights/highlight_group_store.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public getDecorator(
+        hlId: number,
+    ):
+        | { decorator: TextEditorDecorationType; options: ThemableDecorationRenderOptions }
+        | { decorator: undefined; options: undefined } {
+        const decorator = this.highlighIdToDecorator.get(hlId);
+        if (decorator) return { decorator, options: this.decoratorConfigurations.get(decorator)! };
+        return { decorator: undefined, options: undefined };
```
ollien, line 147: nit: any reason to not just return `undefined`?

src/highlights/highlight_grid.ts
```diff
@@ -1,345 +1,228 @@
...
+        this.isDirty = true;
     }
 
-    cleanRow(row: number) {
-        delete this.grid[row];
+    handleRedrawFlush() {
+        if (this.isDirty && this.editor && this.viewport) {
+            this.refreshDecorations();
+            this.isDirty = false;
+        }
```
ollien, line 63: It might be worth an info log if there is no editor or viewport; it may be helpful in debugging user issues

src/highlights/grid_line.ts
```diff
@@ -0,0 +1,187 @@
+import { cloneDeep, findLast } from "lodash";
+
+import { Highlight, HighlightRange, LineCell, NormalHighlightRange, VimCell, VirtualHighlightRange } from "./types";
+import { CellIter, getWidth, isDouble, splitGraphemes } from "./util";
+
+/**
+ * A class to manage the grid lines of a grid.
+ * Handle the grid_line event and be responsible for computing the highlight ranges.
+ */
+export class GridLine {
```
ollien, line 10: just a naming thing: I think this should be plural (I thought this was only managing one line at first :) )

ollien (approved):
Nice work! Some small stuff above, but I really am liking the way this came together.

src/highlights/highlight_group_store.ts
xiyaowong: Personal preferences

```
const ret = get()
if (!ret) return
const { a, b } = ret
...

vs

const { a, b } = get()
if (!a) return
...
```

src/highlights/grid_line.ts
xiyaowong: GridLine does indeed cause confusion; typically, one would think it only manages a single row. However, using the plural form also feels awkward to me. Nevertheless, it doesn't matter much; there are comments, and no chance for reuse.

ollien: Could also do something like GridLineTracker, but I won't push on this too hard

xiyaowong: Maybe `GridLineHandler`?

ollien: Something like that sure. I was trying to emphasize that it stores data, but I could live with Handler

xiyaowong:
It needs a few more days of testing.

src/highlights/highlight_grid.ts
xiyaowong: Regarding this, I have added detailed comments. It is expected for the editor to be null here. In fact, every time a new window is created, the editor is null.

theol0403 (approved):
working for me so far, nice

src/highlights/highlight_group_store.ts
```diff
@@ -0,0 +1,157 @@
...
+    const newConfig: ThemableDecorationRenderOptions = { ...config };
+    newConfig.backgroundColor = normalizeThemeColor(newConfig.backgroundColor);
+    newConfig.borderColor = normalizeThemeColor(newConfig.borderColor);
+    newConfig.color = normalizeThemeColor(newConfig.color);
+    newConfig.outlineColor = normalizeThemeColor(newConfig.outlineColor);
+    newConfig.overviewRulerColor = normalizeThemeColor(newConfig.overviewRulerColor);
+    return newConfig;
+}
+
+export class HighlightGroupStore implements Disposable {
```
justinmk, line 79: "store" seems unnecessary. Mentioning this so it doesn't become a pattern. 

```suggestion
export class HighlightGroups implements Disposable {
```

src/highlights/highlight_group_store.ts
```diff
@@ -0,0 +1,157 @@
...
+        const decorator = window.createTextEditorDecorationType({
+            ...options,
+            rangeBehavior: DecorationRangeBehavior.ClosedClosed,
+        });
+        this.decoratorConfigurations.set(decorator, options);
+        this.highlighIdToDecorator.set(id, decorator);
+        this.disposables.push(decorator);
+    }
+
+    public addHighlightGroup(id: number, attrs: VimHighlightUIAttributes, groups: string[]): void {
```
justinmk, line 118: don't need to repeat the name of the container in its methods. but docstrings are useful.

```suggestion
    /** Adds a highlight group. */
    public add(id: number, attrs: VimHighlightUIAttributes, groups: string[]): void {
```

src/highlights/grid_line_handler.ts
```diff
@@ -0,0 +1,186 @@
+import { cloneDeep, findLast } from "lodash";
+
+import { Highlight, HighlightRange, LineCell, NormalHighlightRange, VimCell, VirtualHighlightRange } from "./types";
+import { CellIter, getWidth, isDouble, splitGraphemes } from "./util";
+
+/**
+ * Handle the grid_line event and be responsible for computing the highlight ranges.
```
justinmk, line 7:
```suggestion
 * Handles grid_line event and computes related highlight ranges.
```

src/highlights/highlight_group_store.ts
xiyaowong: My first idea was to name it like that as well, but using the plural form for a property feels a bit awkward. "store" is also used in Neovim's UI doc.

src/highlights/highlight_group_store.ts
xiyaowong: Coincide with my original idea

Issue #2000, opened 2024-05-22
bug: Nvim0.10.0 crash / neovim was disconnected


HappyRustacean:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

NVIM v0.10.0 Build type: Release LuaJIT 2.1.1713773202

### Operating system/version

Mac Os X 14.5 (23F79)

### Describe the bug

In the latest version v1.12.0, Neovim on macOS automatically exits after switching multiple VSCode tabs, causing plugins to disconnect. 

### Steps To Reproduce

1. Open VSCode on macOS with the Neovim extension.
2. Open multiple tabs within VSCode and maybe some splitted window.
3. Switch between the tabs multiple times.

### How to Solve
downgrade nvim version to

NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1713773202

xiyaowong:
So it's actually a bug of nvim0.10?

Are you sure you didn't use any plugins?

bomut:
Just adding my 5ct: I can confirm, after downgrading neovim from 0.10.0 to 0.9.5 the issue is resolved.

bomut:
No, I downgraded neovim via homebrew (following [this](https://stackoverflow.com/a/72106160/24927966) guide using [this bottle](https://github.com/Homebrew/homebrew-core/commit/63aa44faba5b5274a1a7579510cd5a570a2cca5f\#diff-4475fb60ced0426b0948706be3a2d7fae40b45620fd25acc5cbc7d5d6d8f9d34); yes, it's a bit cumbersome).

ollien:
Any kind of logs would be very helpful. I don't have a Mac to debug with (maybe someone else does), but we may be able to narrow something down. You can find instructions on how to gather them here

https://github.com/vscode-neovim/vscode-neovim/blob/master/CONTRIBUTING.md#logging

tetzng:
I have the same issue

## OS

macOS 14.5 (23F79)

## Neovim version

NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713773202

## Log

I've attached the logs that were captured during the issue. The logs are quite long, and the most relevant information is probably towards the end. However, I'm including the entire log file just to be thorough.

[...]

PR #2092, opened 2024-06-08
 docs: add configuration compatibility section to README 

ollien:
Clarifies a quirk of neovim output that changed between 0.9.x and 0.10. Thanks to @iansinnott for [pointing it out!](https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2155828299)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ mode and editor commands, making the best use of both editors.
 
 -   Install the [vscode-neovim](https://marketplace.visualstudio.com/items?itemName=asvetliakov.vscode-neovim)
     extension.
--   Install [Neovim](https://github.com/neovim/neovim/wiki/Installing-Neovim) **0.9.0** or greater.
+-   Install [Neovim](https://github.com/neovim/neovim/wiki/Installing-Neovim) **0.9.0** or greater (0.10 recommended,
+    [see below](#neovim-configuration-for-compatibility)).
     -   Set the Neovim path in the extension settings. You must specify the full path to Neovim, like
         "`C:\Neovim\bin\nvim.exe`" or "`/usr/local/bin/nvim`".
     -   The setting id is "`vscode-neovim.neovimExecutablePaths.win32/linux/darwin`", respective to your system.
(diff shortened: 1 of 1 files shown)


xiyaowong (approved):
It might be better to put this content on a wiki, making it easier to contribute and edit. Otherwise, the README will get cluttered with various special notes.

ollien:
That's a good point - maybe I'll make a wiki section for "version-specific configurations" and link that in the README

README.md
```diff
@@ -100,6 +101,13 @@ To conditionally activate plugins, `vim-plug` has a
 have built-in support for `cond = vim.g.vscode`. See
 [plugins](https://github.com/vscode-neovim/vscode-neovim/wiki/Plugins) in the wiki for tips on configuring Vim plugins.
 
+#### Neovim configuration for best compatibility
+
+Though the extension strives to be as compatible as possible with older versions of Newvim, some older versions may have
```
xiyaowong, line 106: typo

ollien: > Newvim

Well I guess it's not wrong... 😂

README.md
```diff
@@ -100,6 +101,13 @@ To conditionally activate plugins, `vim-plug` has a
 have built-in support for `cond = vim.g.vscode`. See
 [plugins](https://github.com/vscode-neovim/vscode-neovim/wiki/Plugins) in the wiki for tips on configuring Vim plugins.
 
+#### Neovim configuration for best compatibility
+
+Though the extension strives to be as compatible as possible with older versions of Neovim, some older versions may have
+quirks that are not present anymore. In light of this, certain configuration settings are recommended in some older
+versions for the best experience. These can be found
+[on the wiki](https://github.com/vscode-neovim/vscode-neovim/wiki/Version-Compatibility-Notes).
```
justinmk, line 109: We should show a message if Nvim is older than 0.10. There is no good reason for a vscode user to use an older version of Nvim.

xiyaowong: Disagree, here is just a small problem

At least not right now, because version 0.10.0 has a very serious issue #2000

justinmk: So you agree after Nvim 0.10.1 is released?

xiyaowong: I remain neutral. I usually use the nightly version. However, vscode-neovim currently works with 0.9. It just doesn't handle some edge cases well enough.

ollien: @justinmk I don't necessarily disagree with you. It's definitely worth noting that both VSCode and its extensions  auto-update, but nvim does not. Unless you're greatly invested in the nvim ecosystem, you may not even know to update.

That said, I don't think it would be a bad thing for us to develop some kind of "policy" around how long we support nvim versions for (x months after release?). In this case, the plugin still works, it just has a quirk in this older version of nvim, so I'm inclined to not say it's "broken".

PR #2096, opened 2024-06-09
style: enable `no-unsafe-assignment`, `no-unsafe-return`, and `no-unsafe-member-access`

ollien:
I passed over these when adding the type-checked rules, but I wanted to revisit them. There are some times where it is correct to disable it (as we can't know the types until runtime), but there are a few where we actually do know the types, and we should just take the time to type everything.

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -77,7 +77,8 @@ module.exports = tseslint.config(
             "@typescript-eslint/no-floating-promises": "off", // jumpy; would be nice to turn on, but we have a lot of these
             "@typescript-eslint/unbound-method": "off", // jumpy, given how vscode's API binds this. Would be good to remove.
 
-            // This codebase interacts with `any` a lot, so for now, all no-unsafe-* we violate are currently disabled
+            // I would love to disable this rule, but unfortunately it forbids spreading any[] types, which is
+            // quite useful for forwarding. I think it's a bit overzealous
             "@typescript-eslint/no-unsafe-argument": "off",
         },
     },
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -79,7 +79,6 @@ module.exports = tseslint.config(
 
             // This codebase interacts with `any` a lot, so for now, all no-unsafe-* we violate are currently disabled
             "@typescript-eslint/no-unsafe-argument": "off",
-            "@typescript-eslint/no-unsafe-member-access": "off",
         },
     },
     {
@@ -88,6 +87,7 @@ module.exports = tseslint.config(
             // Good for production code to avoid unsafe type leaks, potentially very frustrating and of little value for tests
             "@typescript-eslint/no-unsafe-assignment": "off",
             "@typescript-eslint/no-unsafe-return": "off",
+            "@typescript-eslint/no-unsafe-member-access": "off",
         },
     },
     {
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -80,7 +80,6 @@ module.exports = tseslint.config(
             // This codebase interacts with `any` a lot, so for now, all no-unsafe-* we violate are currently disabled
             "@typescript-eslint/no-unsafe-argument": "off",
             "@typescript-eslint/no-unsafe-member-access": "off",
-            "@typescript-eslint/no-unsafe-return": "off",
         },
     },
     {
@@ -88,6 +87,7 @@ module.exports = tseslint.config(
         rules: {
             // Good for production code to avoid unsafe type leaks, potentially very frustrating and of little value for tests
             "@typescript-eslint/no-unsafe-assignment": "off",
+            "@typescript-eslint/no-unsafe-return": "off",
         },
     },
     {
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -79,9 +79,15 @@ module.exports = tseslint.config(
 
             // This codebase interacts with `any` a lot, so for now, all no-unsafe-* we violate are currently disabled
             "@typescript-eslint/no-unsafe-argument": "off",
-            "@typescript-eslint/no-unsafe-assignment": "off",
             "@typescript-eslint/no-unsafe-member-access": "off",
             "@typescript-eslint/no-unsafe-return": "off",
+        },
+    },
+    {
+        files: ["src/test/**/*.ts"],
+        rules: {
+            // Good for production code to avoid unsafe type leaks, potentially very frustrating and of little value for tests
+            "@typescript-eslint/no-unsafe-assignment": "off",
         },
     },
     {
--- a/package.json
+++ b/package.json
@@ -2167,6 +2167,7 @@
         "ts-loader": "^9.5.1",
         "typescript": "^5.4.4",
         "typescript-eslint": "^7.8.0",
+        "vscode-languageserver-types": "^3.17.5",
         "webpack": "^5.91.0",
         "webpack-cli": "^5.1.4"
     },
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -85,14 +85,16 @@ class ActionManager implements Disposable {
                 return config.has(names);
             }
         });
-        this.add("get_config", (names: string | string[]) => {
+        this.add("get_config", ((names: string | string[]) => {
             const config = workspace.getConfiguration();
             if (Array.isArray(names)) {
                 return names.map((name) => config.get(name));
             } else {
+                // We can't know the type of this configuration value until runtime.
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-return
                 return config.get(names);
             }
-        });
+        }) as ((names: string) => any) & ((names: string[]) => any[]));
         this.add("update_config", async (names: string | string[], values: any, target?: "global" | "workspace") => {
             const config = workspace.getConfiguration();
             let targetConfig = null;
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -101,12 +101,11 @@ class ActionManager implements Disposable {
             if (target) {
                 targetConfig = target === "global" ? ConfigurationTarget.Global : ConfigurationTarget.Workspace;
             }
-            if (!Array.isArray(names)) {
-                names = [names];
-                values = [values];
-            }
-            for (const [idx, name] of names.entries()) {
-                await config.update(name, values[idx], targetConfig);
+
+            const namesArr = ensureArray(names);
+            const valuesArr = ensureArray(values);
+            for (const [idx, name] of namesArr.entries()) {
+                await config.update(name, valuesArr[idx], targetConfig);
             }
         });
         this.add("start-multiple-cursors", (ranges: Range[]) => {
@@ -124,6 +123,14 @@ class ActionManager implements Disposable {
                 this.client.command(`doautocmd ${e.focused ? "FocusGained" : "FocusLost"}`),
             ),
         );
+    }
+}
+
+function ensureArray<T>(x: T | T[]): T[] {
+    if (Array.isArray(x)) {
+        return x;
+    } else {
+        return [x];
     }
 }
 
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -594,12 +594,7 @@ export class BufferManager implements Disposable {
         const eol = document.eol === EndOfLine.LF ? "\n" : "\r\n";
         const lines = document.getText().split(eol);
         // We don't care about the name of the buffer if it's not a file
-        const bufname =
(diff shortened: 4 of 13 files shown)


src/document_change_manager.ts
```diff
@@ -390,12 +390,7 @@ export class DocumentChangeManager implements Disposable {
             changeArgs.push([start.line, startBytes, end.line, endBytes, text.split(eol)]);
         }
 
-        const bufTick: number = await this.client.request("nvim_buf_get_changedtick", [bufId]);
-        if (!bufTick) {
```
ollien, line 394: I hesitated slightly here, but in my testing, I don't think it's possible for `bufTick` to return a falsy value (unless it's zero, but based on the log, that doesn't seem to be the intent of the log). If we pass an unknown buffer, it throws.

src/buffer_manager.ts
```diff
@@ -722,24 +728,27 @@ export class BufferManager implements Disposable {
             // !Another hack is to retrieve cursor with delay - when we receive an external buffer the cursor pos is not immediately available
             // [1, 0]
             setTimeout(async () => {
-                const neovimCursor: [number, number] = await this.client.request("nvim_win_get_cursor", [closeWinId]);
-                if (neovimCursor) {
```
ollien, line 726: I hesitated here as well, for the same reason as the other one. However, the nvim doc doesn't say this can return anything nullish, so I'm inclined to think this was an erroneous error check as well?

eslint.config.cjs
```diff
@@ -77,11 +77,18 @@ module.exports = tseslint.config(
             "@typescript-eslint/no-floating-promises": "off", // jumpy; would be nice to turn on, but we have a lot of these
             "@typescript-eslint/unbound-method": "off", // jumpy, given how vscode's API binds this. Would be good to remove.
 
-            // This codebase interacts with `any` a lot, so for now, all no-unsafe-* we violate are currently disabled
+            // I would love to disable this rule, but unfortunately it forbids spreading any[] types, which is
+            // quite useful for forwarding. I think it's a bit overzealous
             "@typescript-eslint/no-unsafe-argument": "off",
```
justinmk, line 82: by "disable this rule" do you mean "enable"? it's already disabled :)

package.json
```diff
@@ -2167,6 +2167,7 @@
         "ts-loader": "^9.5.1",
         "typescript": "^5.4.4",
         "typescript-eslint": "^7.8.0",
+        "vscode-languageserver-types": "^3.17.5",
```
justinmk, line 2170: does not seem worth this extra dependency for a single call

eslint.config.cjs
ollien: Hehe, yes.

package.json
ollien: Yeah that's fair. I figured it was nbd because it's just types but I'll remove it

src/buffer_manager.ts
```diff
@@ -722,24 +728,27 @@ export class BufferManager implements Disposable {
...
+                const finalLine = neovimCursor[0] - 1;
+                let finalCol = neovimCursor[1];
+                try {
+                    finalCol = convertByteNumToCharNum(doc.lineAt(finalLine).text, neovimCursor[1]);
+                    logger.debug(`Adjusted cursor: [${finalLine}, ${finalCol}]`);
+                } catch (e) {
+                    logger.warn(`Unable to get cursor pos for external buffer: ${id}`);
                 }
+
+                const selection = new Selection(finalLine, finalCol, finalLine, finalCol);
```
justinmk, line 749: if `convertByteNumToCharNum` why do we still set a selection? seems like that would just result in nonsense behavior

src/cmdline_manager.ts
```diff
@@ -43,6 +43,8 @@ export class CommandLineManager implements Disposable {
     public constructor(private main: MainController) {
         eventBus.on("redraw", this.handleRedraw, this, this.disposables);
         this.input = window.createQuickPick();
+        // https://github.com/microsoft/vscode/issues/73904
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
         (this.input as any).sortByLabel = false;
```
justinmk, line 48: `as any` should be enough to avoid a warning. seems heavy handed to have to add the `// eslint-disable...` whenever we need that escape hatch. `as any` already clearly signals that it's intentional

src/logger.ts
```diff
@@ -102,7 +102,7 @@ export class Logger implements Disposable {
     }
 
     private log(level: vscode.LogLevel, scope: string, logToOutputChannel: boolean, args: any[]): void {
-        const msg = args.reduce((p, c, i) => {
+        const msg = args.reduce((p: string, c: any, i: number) => {
```
justinmk, line 105: type inference doesn't capture this already?

justinmk (approved):
Churn looks minimal 👍

xiyaowong:
Please do not add rules for boilerplate code. There are some that I deliberately removed. Why is type inference even prohibited? I really dislike such restrictions.

src/cmdline_manager.ts
ollien: Yeah I agree. I felt icky on this one but this rule offers no customization

xiyaowong:
Can we directly use some established rule packages? https://github.com/xiyaowong/eslint-config

ollien:
> Please do not add rules for boilerplate code. There are some that I deliberately removed. Why is type inference even prohibited? I really dislike such restrictions.

There are no rules for boilerplate here. The only thing that is being restricted is to return implied `any`. Type inference is fine.

src/logger.ts
ollien: It captured `c` and `i`, but not `p`, so `msg` ended up as `any`. I think the `+` with `any` confused the compiler.

ollien:
> Can we directly use some established rule packages? [xiyaowong/eslint-config](https://github.com/xiyaowong/eslint-config)

Generally, we're already using the "recommended" rules from `typescript-eslint`. These were exceptions that I disabled so I didn't change too much at once.

src/actions.ts
```diff
@@ -125,4 +126,12 @@ class ActionManager implements Disposable {
     }
 }
 
+function ensureArray<T>(x: T | T[]): T[] {
```
xiyaowong, line 129: https://lodash.com/docs/4.17.15#castArray

src/buffer_manager.ts
```diff
@@ -638,6 +633,17 @@ export class BufferManager implements Disposable {
         actions.fireNvimEvent("document_buffer_init", bufId);
     }
 
+    private async bufName(docUri: Uri): Promise<string> {
+        if (docUri.scheme !== "file") {
+            return docUri.toString();
+        } else if (config.useWsl) {
+            const wslpath = await actions.lua<string>("wslpath", docUri.fsPath);
+            return wslpath;
```
xiyaowong, line 641: Can't we set generic types on the original code?

src/buffer_manager.ts
```diff
@@ -722,24 +728,27 @@ export class BufferManager implements Disposable {
...
-                        logger.debug(`Adjusted cursor: [${finalLine}, ${finalCol}]`);
-                    } catch (e) {
-                        logger.warn(`Unable to get cursor pos for external buffer: ${id}`);
-                    }
-
-                    const selection = new Selection(finalLine, finalCol, finalLine, finalCol);
-                    editor.selections = [selection];
-                    editor.revealRange(selection, TextEditorRevealType.AtTop);
+                // Return type is known from vimdoc
+                // https://neovim.io/doc/user/api.html#nvim_win_get_cursor()
```
xiyaowong, line 732: This comment is completely unnecessary.

xiyaowong:
Don't mind it, I just woke up and didn't review it carefully.

src/main_controller.ts
```diff
@@ -257,10 +257,15 @@ export class MainController implements vscode.Disposable {
             }
             targetRange = doc.validateRange(targetRange);
             editor.selections = [new vscode.Selection(targetRange.start, targetRange.end)];
+            // This is an arbitrary action from VSCode, we can't know the return type
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
             const res = await actions.run(action, ...(options.args || []));
```
xiyaowong, line 262: I really think this kind of comment is redundant.....

src/main_controller.ts
```diff
@@ -257,10 +257,15 @@ export class MainController implements vscode.Disposable {
...
             if (options.restore_selection !== false) {
                 editor.selections = prevSelections;
             }
+
+            // This is an arbitrary action from VSCode, we can't know the return type
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
xiyaowong, line 268: redundant

src/utils/vscode.ts
```diff
@@ -12,6 +12,7 @@ import {
     TextEditor,
     commands,
 } from "vscode";
+import type * as lsp from "vscode-languageserver-types";
```
xiyaowong, line 15: Adding a package just for a Range seems unnecessary. It would be better to directly define LspRange like this:

```typescript
{
    start: { line: number; character: number };
    end: { line: number; character: number };
}
```

xiyaowong (changes requested):

src/buffer_manager.ts
xiyaowong: A community contributor wrote this part, and nobody really pays much attention to external-buffer.

src/utils/vscode.ts
ollien: Yeah I waffled on this, see above :D

src/actions.ts
ollien: TIL

src/buffer_manager.ts
ollien: Yeah I guess so. I just found the nested ternary hard to read and I did this to figure out what the problem was. But if you prefer the ternary I can do that.

src/main_controller.ts
ollien: It's a bit repetitive, sure. I did it just to justify the lint disabling, but if you would prefer it gone I can.

src/main_controller.ts
```diff
@@ -360,6 +368,8 @@ export class MainController implements vscode.Disposable {
     };
 
     private setClientInfo() {
+        // We know version will be present in package.json...
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
         const versionString = this.extContext.extension.packageJSON.version as string;
```
xiyaowong, line 373: Will this throw an error?
```suggestion
        // We know version will be present in package.json...
        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
        const versionString = (this.extContext.extension.packageJSON as any).version as string;
```

ollien: Yes, the compiler already things `packageJSON` is `any`.

xiyaowong:
Translate into English: "I agree with prohibiting unreachable conditional judgments. However, I don't think implicit 'any' poses a significant issue. In most cases, we don't perform extensive operations on implicit 'any' types (because there are no autocompletion hints, and if operations are necessary, we would explicitly specify the type). These are mostly short-lived return values."

---
I support banning unreachable conditions. However, I don't see implicit 'any' causing major issues. In most cases, we don't manipulate implicit 'any' types extensively (because there are no autocompletion, and if manipulation is required, we would explicitly specify the type). These are mostly short-lived return values.

xiyaowong:
I support banning unreachable conditions. However, I don't see implicit 'any' causing major issues. In most cases, we don't manipulate implicit 'any' types extensively (because there are no autocompletion, and if manipulation is required, we would explicitly specify the type). These are mostly short-lived return values.

src/buffer_manager.ts
xiyaowong: The ternary operator is difficult to read when not formatted. 😂

ollien:
> I support banning unreachable conditions. However, I don't see implicit 'any' causing major issues. In most cases, we don't manipulate implicit 'any' types extensively (because there are no autocompletion, and if manipulation is required, we would explicitly specify the type). These are mostly short-lived return values.

No, it's definitely not a major problem. I would not be offended if we just decided not to take this (though I would like to commit the unnecessary error checking I removed here. I think the benefit here, if anything, is that we enfroce typing on data coming back from nvim.

xiyaowong:
Alright, let "// eslint-disable ..." fill every corner.

ollien:
If you feel that way about it, I don't mind closign this PR, truly.

ollien:
Closing, due to the complexity adds. I'll probably circle around to remove those error checks when I get a spare cycle.

PR #2097, opened 2024-06-10
feat: Allow using the status bar as the cmdline

xiyaowong:
A very rough implementation. The effect is not bad. However, this is just an attempt. The cmdline in the statusbar is indeed not very conspicuous.

### DIFF
--- a/package.json
+++ b/package.json
@@ -389,6 +389,11 @@
                         }
                     },
                     "description": "Composite key bindings"
+                },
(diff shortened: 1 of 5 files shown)


ollien:
>  The cmdline in the statusbar is indeed not very conspicuous.

Yeah, this was my experience with VSCodeVim, and is ultimately why I'm not a fan.

PR #2103, opened 2024-06-12
fix(buffer): prevent concurrent syncs and stale data

xiyaowong:
Synchronous layout operations take some time, and even with debouncing and a CancelToken, it's not guaranteed that only one synchronous operation will run at a time. If multiple synchronous operations are executed simultaneously, conflicts can occur, especially when initializing the buffer, leading to errors when setting the buffer name.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -70,11 +70,26 @@ function makeEditorOptionsVariable(options?: TextEditorOptions) {
  */
 export class BufferManager implements Disposable {
     private disposables: Disposable[] = [];
+
     /**
-     * Internal sync promise
+     * Promise for other modules to wait for layout synchronization.
+     * Set this before running the debounced function, since layout is outdated.
      */
-    private syncEditorLayoutPromise?: ManualPromise;
-    private syncEditorLayoutSource?: CancellationTokenSource;
+    private syncLayoutPromise?: ManualPromise;
+    /**
+     * Cancels sync operations to avoid using outdated data
+     */
+    private syncLayoutSource?: CancellationTokenSource;
+    /**
+     * Indicates if synchronization is in progress.
+     * Make sure there is only one sync operation at a time.
+     */
+    private isSyncingLayout = false;
+    /**
+     * Indicates if the layout is outdated
+     */
+    private isLayoutOutdated = false;
+
     /**
      * Text documents originated externally, as consequence of neovim command, like :help or :PlugStatus
      */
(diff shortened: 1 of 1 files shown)


src/buffer_manager.ts
```diff
@@ -413,40 +428,46 @@ export class BufferManager implements Disposable {
...
+                await this.cleanupWindowsAndBuffers(visibleEditors);
+
+                if (token?.isCancellationRequested) continue;
+                await this.syncVisibleEditors(visibleEditors);
+
+                if (token?.isCancellationRequested) continue;
+                await this.syncActiveEditor(activeEditor);
+            }
+        } catch (e) {
+            logger.error(`Error syncing layout: ${e}`);
```
justinmk, line 460: `%O` will often produce a better result than the default toString().
```suggestion
            logger.error(`Error syncing layout: %O`, e);
```

src/buffer_manager.ts
```diff
@@ -413,40 +428,46 @@ export class BufferManager implements Disposable {
...
 
     private onEditorLayoutChanged = async () => {
-        if (!this.syncEditorLayoutPromise) {
-            this.syncEditorLayoutPromise = new ManualPromise();
+        this.syncLayoutPromise = this.syncLayoutPromise ?? new ManualPromise();
+        this.isLayoutOutdated = true;
+        this.syncLayoutSource?.cancel();
+        this.syncLayoutSource = new CancellationTokenSource();
+        if (!this.isSyncingLayout) {
+            await this.syncEditorLayoutDebounced();
```
justinmk, line 436: When we skip this because `isSyncingLayout=true`, is it possible that we lose the "latest" update? Or does the while-loop in `syncEditorLayout()` always get the latest?

src/buffer_manager.ts
```diff
@@ -413,40 +428,46 @@ export class BufferManager implements Disposable {
...
+
+                if (token?.isCancellationRequested) continue;
+                await this.syncActiveEditor(activeEditor);
+            }
+        } catch (e) {
+            logger.error(`Error syncing layout: ${e}`);
+        } finally {
+            this.isSyncingLayout = false;
+            this.syncLayoutPromise?.resolve();
+            this.syncLayoutPromise = undefined;
```
justinmk, line 464: Is it possible that `syncLayoutPromise.resolve()` throws? If so, we would miss this assignment. Could assign this before calling resolve()

src/buffer_manager.ts
xiyaowong: When an event occurs, `isLayoutOutdated` will be set to true. The ongoing loop will start a new cycle and use the latest data.

src/buffer_manager.ts
xiyaowong: Changed. Tip: The logger uses `inspect` to concatenate log information. So just pass it in as a separate parameter.

xiyaowong:
On slower devices, quickly and continuously opening new documents can reproduce this issue. Note that since errors were not caught previously, there will be no prompts, and you need to check the console for errors. The common symptoms are an immovable cursor and strange highlights appearing in the document.

PR #2105, opened 2024-06-13
fix: Align output with nvim messages and auto show based on cmdheight

### LINKED ISSUE #2099: bug: vs code now constantly auto switching to the output panel on multiline messages
### Check the following:

### Neovim version (nvim -v)

0.10.0

### Operating system/version

MacOS 14.4

### Describe the bug

In recent versions of this extension (> `v1.13.1`) the output panel gets focused whenever neovim outputs multiline messages. This behavior does not happen in v1.13.1, and is probably related to the "revamp messages output channel" PR (https://github.com/vscode-neovim/vscode-neovim/pull/2024) introduced in v1.14. 

I use a neovim plugin ([pounce](https://github.com/rlane/pounce.nvim)) with this extension, and that plugin outputs multiline messages in the course of normal use, which has made this extension near unusable because it's now constantly covering up the terminal window or otherwise refocusing things against my will.

VS code's focusing behavior is quite aggressive and is not configurable (there are lot of [longstanding issues](https://github.com/microsoft/vscode/issues/105270) about this in the vs code repo) so it would be nice to have a setting to disable this behavior in vscode-neovim

Related to: https://github.com/vscode-neovim/vscode-neovim/issues/2046 and https://github.com/vscode-neovim/vscode-neovim/issues/1812

### Steps To Reproduce

1. emit a multi-line message in neovim (`:echom "test\n"`). 
2. The ouput panel pops up, taking over focus from whatever was visible previously

### Expected Behavior

The extension behaves the same as it did in v1.13.1 and shows multiline output in the status bar area rather than the output channel

xiyaowong:
fixes #2099

### DIFF
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -3,6 +3,7 @@ import { Disposable, OutputChannel, window } from "vscode";
 import { EXT_NAME } from "./constants";
 import { EventBusData, eventBus } from "./eventBus";
 import { createLogger } from "./logger";
+import { MainController } from "./main_controller";
 import { disposeAll } from "./utils";
 
 const logger = createLogger("MessagesManager");
@@ -11,7 +12,7 @@ export class MessagesManager implements Disposable {
     private disposables: Disposable[] = [];
     private channel: OutputChannel;
 
-    public constructor() {
+    public constructor(private readonly main: MainController) {
         this.channel = window.createOutputChannel(`${EXT_NAME} messages`);
         this.disposables.push(this.channel, eventBus.on("redraw", this.handleRedraw, this));
     }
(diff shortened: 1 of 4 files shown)


src/messages_manager.ts
```diff
@@ -20,38 +21,30 @@ export class MessagesManager implements Disposable {
...
+                    // Ignore return_prompt
+                    //
+                    // A note to future readers: return_prompt is sent much more often with ui_messages. It may
+                    // not do what you expect from what :help ui says, so be careful about using these events.
+                    // See: https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2144175058
+                    if (type === "return_prompt") return acc;
+                    if (clear) return "";
+                    return acc + content.map((c) => c[1]).join("");
+                }, "");
+                const outputMsg = hasReturnPrompt ? msg.replace(/\n$/, "") : msg;
```
ollien, line 38: I know this was like this, but we may consider trimming trailing all newlines from here (`\n*$`) to minimize accidental pop outs

src/messages_manager.ts
```diff
@@ -20,38 +21,30 @@ export class MessagesManager implements Disposable {
...
+
+                this.writeMessage(outputMsg);
+
+                const lineCount = outputMsg.split("\n").length;
+                const cmdheight = (await this.main.client.getOption("cmdheight")) as number;
```
ollien, line 43: Not totally necessary, but it might be nice to say this has to be a minimum of 1, so that every message doesn't show up if someone sets it to zero. I doubt that is ever desirable

ollien (approved):
Nice idea. It might also be a good idea to document this behavior in the README so users know they can configure this

src/messages_manager.ts
xiyaowong: However, this will cause the expected line breaks to be removed.

src/messages_manager.ts
xiyaowong: I don't understand what you mean. By checking `cmdheight` to determine whether to pop up the output, the behavior is basically the same as in nvim. (In nvim, messages are in a window.)

Also, there's no need to include it in the README. If the behavior is consistent with nvim, then there's no need to mention it.

xiyaowong:
Or add this to wiki

src/messages_manager.ts
ollien: cmdheight can take on a value of zero. I'm suggesting we guard against that case

src/messages_manager.ts
ollien: It could. But if they're at the end of the string, that likely doesn't matter.

ollien:
However you wanna do it is fine. I just know the auto-show functionality has caught people off guard, and it might be nice to document `cmdheight`s correlation to it.

src/messages_manager.ts
xiyaowong: A user sets 'cmdheight' to 0 and then complains why the output always pops up?

ollien: _shrug_ you can take or leave it, it was just a thought I had

PR #2112, opened 2024-06-15
fix(highlights): wait for doc changes before updating highlights

### LINKED ISSUE #2111: bug:  Render issue when deleting line
### Check the following:

### Neovim version (nvim -v)

0.10.0

### Operating system/version

EndeavourOS 

### Describe the bug

When using dd to delete a line , the line  below is not rendered correctly.

This is a rendering issues and using `CTRL + L` seems to fix it. 
Is there some way how to fix this in general?

### Steps To Reproduce

For example see:
image

Deleteting the empty line with dd gives something like:
image

### Expected Behavior

The line should render correclty

xiyaowong:
fix #2111

We should still wait for the document synchronization to complete before calculating the highlights. Otherwise, there is a possibility of computing incorrect highlights due to inconsistent content. Previously, this logic was mistakenly deleted.

### DIFF
--- a/src/highlight_manager.ts
+++ b/src/highlight_manager.ts
@@ -30,7 +30,13 @@ export class HighlightManager implements Disposable {
         if (!this.highlightGrids.has(gridId)) {
             this.highlightGrids.set(
                 gridId,
-                new HighlightGrid(gridId, this.groupStore, this.main.bufferManager, this.main.viewportManager),
+                new HighlightGrid(
+                    gridId,
+                    this.groupStore,
+                    this.main.bufferManager,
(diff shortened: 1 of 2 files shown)


jruebsam:
> @jruebsam Can you confirm if this PR addresses your issue? (VSCode Extension -> Install from VSIX) https://github.com/vscode-neovim/vscode-neovim/actions/runs/9527929291/artifacts/1604736345

This indeed seems to fix the issue.

ollien (approved):
Nice! I noticed a quirk that seemed related to this but I didn't have time to isolate it.

PR #2116, opened 2024-06-16
feat: Integrate BufWriteCmd for "write" command

xiyaowong:
- [x] Works
- [x] Remote
remote
- [x] Add tests
- [x] Update readme

close #2075

### DIFF
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -308,7 +308,8 @@ end
 ---@field uri string
 ---@field uri_data table
 ---@field modifiable boolean
----@field bufname string|nil
+---@field bufname string
+---@field modified boolean
 
 ---@param data InitDocumentBufferData
 function M.init_document_buffer(data)
@@ -324,17 +325,26 @@ function M.init_document_buffer(data)
   api.nvim_buf_set_var(buf, "vscode_editor_options", data.editor_options)
   api.nvim_buf_set_var(buf, "vscode_uri", data.uri)
   api.nvim_buf_set_var(buf, "vscode_uri_data", data.uri_data)
-  api.nvim_buf_set_option(buf, "modifiable", data.modifiable)
   -- force acwrite, which is similar to nofile, but will only be written via the
   -- BufWriteCmd autocommand.
   api.nvim_buf_set_option(buf, "buftype", "acwrite")
   api.nvim_buf_set_option(buf, "buflisted", true)
-
-  if data.bufname and data.bufname ~= vim.NIL then
-    api.nvim_buf_set_name(buf, data.bufname)
-  end
+  api.nvim_buf_set_option(buf, "modifiable", data.modifiable)
+  api.nvim_buf_set_option(buf, "modified", false)
+  api.nvim_buf_set_name(buf, data.bufname)
 
   set_buffer_autocmd(buf)
+end
+
+---Reset undo tree for a buffer
+-- Called from extension when opening/creating new file in vscode to reset undo tree
+function M.clear_undo(buf)
+  local mod = vim.bo[buf].modified
+  local ul = vim.bo[buf].undolevels
+  api.nvim_buf_set_option(buf, "undolevels", -1)
+  api.nvim_buf_set_lines(buf, 0, 0, false, {})
+  api.nvim_buf_set_option(buf, "undolevels", ul)
+  api.nvim_buf_set_option(buf, "modified", mod)
 end
 
 --#endregion
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -284,6 +284,8 @@ end
 
 --#region Buffer management
 
+--- Implements :write and related commands, via buftype=acwrite. #521 #1260
+---
 local function set_buffer_autocmd(buf)
   api.nvim_create_autocmd({ "BufWriteCmd" }, {
     buffer = buf,
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -321,14 +321,13 @@ function M.init_document_buffer(data)
   -- set vscode controlled flag so we can check it neovim
   api.nvim_buf_set_var(buf, "vscode_controlled", true)
   -- In vscode same document can have different insertSpaces/tabSize settings
-  -- per editor
-  -- however in neovim it's per buffer. We make assumption here that these
-  -- settings are same for all editors
+  -- per editor; in Nvim it's per buffer. We assume here that these settings are
+  -- same for all editors.
   api.nvim_buf_set_var(buf, "vscode_editor_options", data.editor_options)
   api.nvim_buf_set_var(buf, "vscode_uri", data.uri)
   api.nvim_buf_set_var(buf, "vscode_uri_data", data.uri_data)
   -- force acwrite, which is similar to nofile, but will only be written via the
-  -- BufWriteCmd autocommand.
+  -- BufWriteCmd autocommand. #521 #1260
   api.nvim_buf_set_option(buf, "buftype", "acwrite")
   api.nvim_buf_set_option(buf, "buflisted", true)
   api.nvim_buf_set_option(buf, "modifiable", data.modifiable)
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -315,6 +315,8 @@ end
 function M.init_document_buffer(data)
   local buf = data.buf
 
+  -- Set bufname first so that the filetype detection can work ???
+  api.nvim_buf_set_name(buf, data.bufname)
   api.nvim_buf_set_lines(buf, 0, -1, false, data.lines)
   -- set vscode controlled flag so we can check it neovim
   api.nvim_buf_set_var(buf, "vscode_controlled", true)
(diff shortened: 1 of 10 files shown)


runtime/vscode/overrides/vscode-file-commands.vim
```diff
@@ -49,9 +47,7 @@ AlterCommand e[dit] Edit
 AlterCommand ex Ex
 AlterCommand ene[w] Enew
 AlterCommand fin[d] Find
-AlterCommand w[rite] Write
 AlterCommand sav[eas] Saveas
-AlterCommand wa[ll] Wall
```
justinmk, line 54: Can we also replace `:sav[eas]` ?

runtime/lua/vscode/internal.lua
```diff
@@ -280,4 +282,72 @@ function M.wslpath(path)
   return vim.trim(ret)
 end
 
+--#region Buffer management
+
+local function set_buffer_autocmd(buf)
+  api.nvim_create_autocmd({ "BufWriteCmd" }, {
+    buffer = buf,
+    callback = function(ev)
+      local current_name = vim.api.nvim_buf_get_name(ev.buf)
```
theol0403, line 291: Out of curiosity, would operating by uri be more robust than by name?

runtime/vscode/overrides/vscode-file-commands.vim
```diff
@@ -31,10 +31,8 @@ command! -bang -nargs=? Ex call <SID>editOrNew(<q-args>, <q-bang>)
 command! -bang Enew call <SID>editOrNew('__vscode_new__', <q-bang>)
 command! -bang Find call VSCodeNotify('workbench.action.quickOpen')
 
-command! -complete=file -bang -nargs=? Write if <q-bang> ==# '!' | call VSCodeNotify('workbench.action.files.saveAs') | else | call VSCodeNotify('workbench.action.files.save') | endif
 command! -bang Saveas call VSCodeNotify('workbench.action.files.saveAs')
 
-command! -bang Wall call VSCodeNotify('workbench.action.files.saveAll')
 command! -bang Quit if <q-bang> ==# '!' | call VSCodeNotify('workbench.action.revertAndCloseActiveEditor') | else | call VSCodeNotify('workbench.action.closeActiveEditor') | endif
 
 command! -bang Wq call <SID>saveAndClose()
```
theol0403, line 38: Can we remove these (saveAndClose/wq) too? Same with Quit? I think nvim can handle it now.

src/buffer_manager.ts
```diff
@@ -425,6 +428,104 @@ export class BufferManager implements Disposable {
...
+
+        const currentPath = normalize(current_name);
+        const targetPath = normalize(target_name);
+
+        if (currentPath === targetPath) {
+            await workspace.save(docUri);
+            return;
+        }
+
+        const vimCwd = normalize(await this.main.client.call("getcwd"));
```
theol0403, line 486: Instead of asking vim for the cwd (which we currently provide through hacks), can we get the cwd from the workspace or document?

theol0403:
I found an edge behavior, not sure if this is intentional or a minor bug.

If you make a change, then undo it, you are back to the last saved version but the vscode dirty status is still enabled. However, running `:wall` doesn't clear that dirty status (like it used to). `:w` properly clears it.

runtime/lua/vscode/internal.lua
xiyaowong: The correct one is name, which, like the autocmds data, is data in Vim. It is important to ensure that the same type of data is processed together.

runtime/vscode/overrides/vscode-file-commands.vim
xiyaowong: "quit" may cause the nvim process to exit completely. Additionally, our layout management frequently deletes windows and buffers. Using altercommand for these operations generally poses no major issues. The commonly used command is write.

src/buffer_manager.ts
xiyaowong: The approach here is to compute a generic relative file path representing the current operation, always using Vim data for computation first, and then integrating with the working directory in VSCode for subsequent operations. Vim-cwd may not always be accurate or set. A straightforward example is during remote development where nvim runs locally.


1. vim-target-filepath + vim-cwd => relative-path
2. relative-path + vscode-cwd => vscode-target-filepath

xiyaowong:
> I found an edge behavior, not sure if this is intentional or a minor bug.
> 
> If you make a change, then undo it, you are back to the last saved version but the vscode dirty status is still enabled. However, running `:wall` doesn't clear that dirty status (like it used to). `:w` properly clears it.

This needs to be done in a separate PR. Once the document modifications are synchronized, updating this status will suffice. Additionally, even if a special circumstance causes the status to be incorrect, it's okay; it's not very important.

src/buffer_manager.ts
```diff
@@ -425,6 +428,104 @@ export class BufferManager implements Disposable {
...
+        current_name: string;
+        target_name: string;
+    }) {
+        // Note: workspace.save and workspace.saveAs are smart enough to handle
+        // the documents that are not a real file (e.g. untitled, output, etc.)
+        // so we can just call them directly
+
+        const document = this.getTextDocumentForBufferId(buf);
+        if (document == null) {
+            throw new Error("Received an invalid buffer id from Neovim, cannot save");
```
justinmk, line 464: generally good to mention actual values in messages, where possible.
is the id coming "from Nvim" in this case? doesn't look like it.

```suggestion
            throw new Error(`cannot save invalid Nvim buffer id: ${buf}`);
```

src/buffer_manager.ts
```diff
@@ -425,6 +428,104 @@ export class BufferManager implements Disposable {
...
+        const docUri = document.uri;
+
+        if (document.isUntitled) {
+            await workspace.save(docUri);
+            return;
+        }
+
+        // If using Windows locally and developing on a Unix remote environment,
+        // the saved path can contain backslashes, causing folders to be treated as filenames.
+        const normalize = (p: string) => path.normalize(p).split(path.sep).join(path.posix.sep);
```
justinmk, line 476: do we need a `utils/path.ts` ?

src/buffer_manager.ts
justinmk: that comment + example seems worth adding as a code comment

src/buffer_manager.ts
```diff
@@ -425,6 +428,104 @@ export class BufferManager implements Disposable {
...
+            // Let the user choose the save location
+            // Otherwise, we would have to do too much guessing
+            await workspace.saveAs(docUri);
+            return;
+        }
+        const saveUri = Uri.joinPath(workspaceFolder.uri, vimRelativePath);
+        let fileExists = true;
+        try {
+            await workspace.fs.stat(saveUri);
+            await workspace.fs.readFile(saveUri);
```
justinmk, line 500: could live in `utils/fs.ts:exists()`

runtime/lua/vscode/internal.lua
```diff
@@ -280,4 +282,72 @@ function M.wslpath(path)
...
+
+  -- Set bufname first so that the filetype detection can work ???
+  api.nvim_buf_set_name(buf, data.bufname)
+  api.nvim_buf_set_lines(buf, 0, -1, false, data.lines)
+  -- set vscode controlled flag so we can check it neovim
+  api.nvim_buf_set_var(buf, "vscode_controlled", true)
+  -- In vscode same document can have different insertSpaces/tabSize settings
+  -- per editor
+  -- however in neovim it's per buffer. We make assumption here that these
+  -- settings are same for all editors
```
justinmk, line 326:
```suggestion
  -- In vscode same document can have different insertSpaces/tabSize settings
  -- per editor; in Nvim it's per buffer. We assume here that these settings are
  -- same for all editors.
```

runtime/lua/vscode/internal.lua
```diff
@@ -280,4 +282,72 @@ function M.wslpath(path)
...
+  api.nvim_buf_set_var(buf, "vscode_editor_options", data.editor_options)
+  api.nvim_buf_set_var(buf, "vscode_uri", data.uri)
+  api.nvim_buf_set_var(buf, "vscode_uri_data", data.uri_data)
+  -- force acwrite, which is similar to nofile, but will only be written via the
+  -- BufWriteCmd autocommand.
```
justinmk, line 331:
```suggestion
  -- BufWriteCmd autocommand. #521 #1260
```

would also be helpful if the PR mentioned these references

src/buffer_manager.ts
xiyaowong: Let's add it the next time we need it.

src/buffer_manager.ts
xiyaowong: Let's add it the next time we need it.

xiyaowong: Alright

runtime/lua/vscode/internal.lua
```diff
@@ -280,4 +282,71 @@ function M.wslpath(path)
   return vim.trim(ret)
 end
 
+--#region Buffer management
+
+local function set_buffer_autocmd(buf)
```
justinmk, line 287:
```suggestion
--- Implements :write and related commands, via buftype=acwrite. #521 #1260
local function set_buffer_autocmd(buf)
```

justinmk (approved):

runtime/lua/vscode/internal.lua
xiyaowong: Incorrect, and this comment has already been added.

xiyaowong:
I have updated the README. Please review

PR #2127, opened 2024-06-22
fix: sync dirty-state after undo/redo

### LINKED ISSUE #247: Dirty ("modified") file status
Great extension! I think this is a really nice use of neovim. I'd like to adopt this in my workflow, but I'm having a hard time getting over the fact that vscode tells me files have unsaved changes after backing out all changes with 'u' in normal mode. Is there a way around this?

xiyaowong:
fix #247


theol0403:
It seems to only work the first time for me 

https://github.com/vscode-neovim/vscode-neovim/assets/16546293/bf01cd5f-4438-4c0c-8b5c-372c9091b741

xiyaowong:
It is necessary to check whether `&modified` and `isDirty` are consistent. You can set `set stl=%m` to observe.

theol0403:
You're right... neovim is still reporting the buffer as dirty. There might be some bug going on then. 

It works as expected in nvim, but not vscode (observe the status bar). 


https://github.com/vscode-neovim/vscode-neovim/assets/16546293/ed87b6e6-8ba0-4535-8ece-278e7ba3115a

xiyaowong:
Weird, I can't reproduce it.

src/buffer_manager.ts
```diff
@@ -442,6 +443,15 @@ export class BufferManager implements Disposable {
         );
     }
 
+    // #247
+    private handleBufferModifiedSet({ buf, modified }: EventBusData<"BufModifiedSet">[0]) {
+        if (modified) return; // expected and we can't do anything about it
```
justinmk, line 448: can we go a tiny bit further and be a bit more precise, that would help a lot here. something like:

```suggestion
        if (modified) return; // Nvim buffer is always 'modified', but this is irrelevant to us.
```

would also be nice if we consistently used the `on*` prefix instead of randomly naming things `handle*()`, but that's out of scope for this PR I guess.

xiyaowong: `This was received because we modified the buffer; this is expected.`

xiyaowong: > would also be nice if we consistently used the on* prefix instead of randomly naming things handle*(), but that's out of scope for this PR I guess.

Yes, "on" is for events, "handle" is for actions. Actually, it doesn't matter; both are fine.

justinmk: > This was received because we modified the buffer; this is expected.

Nice!

justinmk: > Yes, "on" is for events, "handle" is for actions. Actually, it doesn't matter; both are fine.

"actions" are not really a well-defined concept. And I think this is more evidence that the "actions" concept is redundant.

theol0403:
> Weird, I can't reproduce it.

Can reproduce 100% of the time on Linux Fedora, `NVIM v0.11.0-dev+302-gbe999e6a0`... but ONLY on the first tab opened when vscode boots. Switching to a different tab and the behavior is expected on that tab.

xiyaowong:
@theol0403 It should be fixed. This is an "issue" with VSCode. Sometimes, when modifying content, VSCode first triggers a dirty-change event (with empty content changes) and then triggers a content-change event. This causes the modified state to be set to false before the synchronization of the changes.

theol0403 (approved):
Will test later, but code looks good.

theol0403:
Works!

src/buffer_manager.ts
```diff
@@ -442,6 +443,15 @@ export class BufferManager implements Disposable {
         );
     }
 
+    // #247
+    private handleBufferModifiedSet({ buf, modified }: EventBusData<"BufModifiedSet">[0]) {
+        if (modified) return; // This was received because we modified the buffer; this is expected.
+        const doc = this.getTextDocumentForBufferId(buf);
+        if (doc && doc.isDirty && !doc.isUntitled && !doc.isClosed) {
+            doc.save();
```
trkoch, line 451: @xiyaowong Is it possible to save without formatting? (`workbench.action.files.saveWithoutFormatting`)

theol0403: Easily achievable by editing format on save setting

xiyaowong: > @xiyaowong Is it possible to save without formatting? (`workbench.action.files.saveWithoutFormatting

You can only modify your formatOnSave settings. Additionally, since it is an undo operation and automatic formatting is enabled, the content and formatting should meet expectations.

PR #2140, opened 2024-07-05
fix: rapid cmdline_show/cmdline_hides would not be handled logically

### LINKED ISSUE #2079: bug: vim-sandwich stops working since 1.14

ollien:
Replaces #2081
Fixes #2079 

I came up with [this idea](https://github.com/vscode-neovim/vscode-neovim/pull/2081#issuecomment-2157087672) a little while ago, but haven't had the chance to sit down and implement it. Ultimately, most problems with our two-way cmdline sync originats from the fact that a cmdline_hide can be sent before the quickpick actually executes a hide event, or vice versa. Attempting to  manipulate the events to "ignore" is error prone, given this ordering problem. 

To repeat myself a bit from the linked comment, the "idea" is that we can handle cmdline events in batches. Once we receive a cmdline_hide event at level 1 (i.e. no nested command lines), we can consider the current batch "terminated", and begin queueing up events. Then, when we get a VSCode "hide" event, we can flush the batch, and continue processing events. 

A screencast showing some usage of the vim sandwich plugin, and various simple commandline usages

https://github.com/vscode-neovim/vscode-neovim/assets/977151/9fb86d8c-d180-470d-910f-ab18d046c574

### DIFF
--- a/src/cmdline/cmdline_queue.ts
+++ b/src/cmdline/cmdline_queue.ts
@@ -1,0 +1,71 @@
+import { type EventBusData } from "../eventBus";
+
+/**
+ * A queue for "batching" cmdline events.
+ *
+ * In the most simple cases, events will simply be passed through and not queued. However, there is a more complicated=
+ * case we must deal with. There is an inherent "race condition" (NB: it's not really a race condition, rather a result
+ * of how the JS event loop works, but it's easiest to think of it as a race) between quickpick hide events and cmdline_* events from nvim. If there is a
+ * cmdline_hide, followed immediately by a cmdline_show, we may call hide() on the quickpick, but onHide will not
+ * be fired until our event handlers have completed execution, leading to bizarre and confusing states. As such,
+ * we "queue" events until we know for sure that the quickpick has bene hidden, and then allow them to be flushed back
+ * to the cmdline_manager.
+ */
+export class CmdlineQueue {
+    private pendingBatches: EventBusData<"redraw">[][] = [];
+    private needFlush: boolean = false;
+    private level: number | null = null;
+
+    /**
+     * Given a newovim event, checks whether or not the caller should handle
+     * this event. If this returns false, the event is enqueued for later re-emission.
+     *
+     * @param event
+     * @returns
+     */
+    handleNvimRedrawEvent(event: EventBusData<"redraw">): boolean {
+        if (this.needFlush) {
+            this.addToBatch(event);
+            return false;
+        }
+
+        if (event.name === "cmdline_show") {
+            const [_content, _pos, _firstc, _prompt, _indent, level] = event.args[0];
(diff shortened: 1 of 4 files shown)


theol0403 (approved):
Great work on this, working for me!

There are several typos in the comments, can you go through them (and also try to make less verbose where applicable).

src/cmdline/cmdline_queue.ts
```diff
@@ -0,0 +1,71 @@
...
+ * of how the JS event loop works, but it's easiest to think of it as a race) between quickpick hide events and cmdline_* events from nvim. If there is a
+ * cmdline_hide, followed immediately by a cmdline_show, we may call hide() on the quickpick, but onHide will not
+ * be fired until our event handlers have completed execution, leading to bizarre and confusing states. As such,
+ * we "queue" events until we know for sure that the quickpick has bene hidden, and then allow them to be flushed back
+ * to the cmdline_manager.
+ */
+export class CmdlineQueue {
+    private pendingBatches: EventBusData<"redraw">[][] = [];
+    private needFlush: boolean = false;
+    private level: number | null = null;
```
theol0403, line 17: Are we tracking level in two places now?

ollien: Unfortunately, yes. This gave me a bit of an ick, but they're acutally used for different purposes; this will not always be in sync with the `level` on the cmdline manager itself. I could probably name this something better.

ollien:
> There are several typos in the comments, can you go through them (and also try to make less verbose where applicable).

Heh, sorry about that. I wrote this before breakfast so I wasn't at my sharpest writing-wise :sweat_smile:

src/cmdline/cmdline_queue.ts
ollien: Hopefully the new naming makes it a bit clearer.

theol0403 (approved):

PR #2152, opened 2024-07-09
fix(with_insert): Correct  "a" or "i" to enter insert mode

xiyaowong:


theol0403:
Why?

xiyaowong:
image

a
image

i
image

xiyaowong:
First, "i" is intuitively correct. So the question is, when do we use "a"? I have already corrected the choice, which should be easy to understand.

theol0403 (approved):

PR #2161, opened 2024-07-14
fix: Cancel UI-affecting tasks upon disposal

### LINKED ISSUE #2160: bug: "Syncing layout: Syncing visible editors" forever
### Check the following:

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Pop!_OS 22.04 LTS

### Describe the bug

As mentioned in #2157, if you try to execute the `neovim.stop` action in vscode through the neovim init or even the `vscode-neovim.afterInitConfig` setting in vscode, it works, but shows a never-ending "Syncing layout: Syncing visible editors" loading notification that cannot be closed.

P.S.: setting `vscode-neovim.neovimClean` is not relevant to this issue, since it is a problem that happens while reading the init, but even if that is set and I use the `vscode-neovim.afterInitConfig`, the same thing happens. Not setting `vscode-neovim.afterInitConfig` while `vscode-neovim.neovimClean` is set works as intended though

image

My guess to why this happens is that it keeps trying to sync the layout even after the process that wanted it to stopped (not sure how that's implemented, but it should be along those lines). Reason why I think that is because if you include a sleep call (say with `vim.cmd('sleep 5')` (even 1 second works for me, but I'm not sure if it depends on the hardware)) in the init before the actual `neovim.stop` the problem no longer happens.

### Steps To Reproduce

1. Put something like this in your neovim init's first line or in `vscode-neovim.afterInitConfig` on vscode (This should disable the plugin on vscode launch, but still allow for `neovim.restart` to start it without issues):
```lua
if vim.g.vscode then
  local vscode = require 'vscode'

  ---Check if it's the first time starting up
  ---@return boolean
  local is_first_startup = function()
    local activated = vscode.eval 'return !!vscode.vscode_neovim_activated'
    if activated then
      return false
    end
    vscode.eval 'vscode.vscode_neovim_activated = true'
    return true
  end

  if is_first_startup() then
    vscode.action 'vscode-neovim.stop'
  end
  return {}
end
```
2. Reload vscode (`workbench.action.reloadWindow`) with the extension enabled
3. Open notifications in vscode (`notifications.showList`) if they aren't shown

### Expected Behavior

The notification that it is "Syncing the layout" should not remain forever after a `neovim.stop` on init or at least be able to be closed at that point.

xiyaowong:
Close #2160

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -3,7 +3,7 @@ import { VimValue } from "neovim/lib/types/VimValue";
 import { ConfigurationTarget, Disposable, Range, commands, window, workspace } from "vscode";
 
 import { eval_for_client } from "./actions_eval";
-import { VSCodeContext, disposeAll, rangesToSelections } from "./utils";
+import { VSCodeContext, disposeAll, rangesToSelections, wait } from "./utils";
 
 function getActionName(action: string) {
     return `neovim:${action}`;
@@ -72,10 +72,7 @@ class ActionManager implements Disposable {
     private initActions() {
         // testing actions
         this.add("_ping", () => "pong");
-        this.add("_wait", async (ms = 1000) => {
-            await new Promise((resolve) => setTimeout(resolve, ms));
-            return "ok";
-        });
+        this.add("_wait", (ms = 1000) => wait(ms).then(() => "ok"));
         this.add("eval", (code: string, args: any) => eval_for_client(code, args));
         this.add("has_config", (names: string | string[]): boolean | boolean[] => {
             const config = workspace.getConfiguration();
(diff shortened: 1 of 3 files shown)


src/actions.ts
```diff
@@ -72,10 +72,7 @@ class ActionManager implements Disposable {
     private initActions() {
         // testing actions
         this.add("_ping", () => "pong");
-        this.add("_wait", async (ms = 1000) => {
-            await new Promise((resolve) => setTimeout(resolve, ms));
-            return "ok";
-        });
+        this.add("_wait", (ms = 1000) => wait(ms).then(() => "ok"));
```
xiyaowong, line 75: This part has nothing to do with PR, I just took this opportunity to change it

PR #2163, opened 2024-07-15
Ease engines requirement

antonk52:
First of all: thank you for your work!

This PR ever so slightly eases the latest VS Code version requirement as the code does not use any of the newer APIs the change is supposed to be fully backward compatible.

The reason I am creating this PR is due to some companies not allowing to update to the most recent VS Code version thus the VS Code marketplace for plugins does not allow to install the most recent version of this plugin and these users miss out on months of new functionality. I hope it is okay to make this change as it does not seem to cause any regression in functionality or add an additional maintenance burden.

The proposed minimum required VS Code version 1.89 is fairly recent [April 2024](https://code.visualstudio.com/updates/v1_89).

I have ran all precommit and `npx tsc` which exits with a 0 exit code. Thus I think this is a safe change.

Thank you

### DIFF
--- a/package.json
+++ b/package.json
@@ -2161,6 +2161,7 @@
     "dependencies": {
         "@jpwilliams/waitgroup": "^2.1.1",
         "async-mutex": "^0.5.0",
+        "core-js": "^3.37.1",
         "fast-myers-diff": "^3.2.0",
         "grapheme-splitter": "^1.0.4",
         "lodash": "^4.17.21",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
         "email": "devpieron@gmail.com"
     },
     "engines": {
-        "vscode": "^1.90.0"
+        "vscode": "^1.89.0"
     },
     "keywords": [
         "keybindings",
@@ -2140,7 +2140,7 @@
         "@types/mocha": "^10.0.7",
         "@types/node": "^20.14.9",
         "@types/sinonjs__fake-timers": "^8.1.5",
-        "@types/vscode": "^1.90.0",
+        "@types/vscode": "1.89.0",
         "@vscode/test-electron": "^2.4.0",
         "@vscode/vsce": "^2.29.0",
         "eslint": "^8.57.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
             "dependencies": {
                 "@jpwilliams/waitgroup": "^2.1.1",
                 "async-mutex": "^0.5.0",
+                "core-js": "^3.37.1",
                 "fast-myers-diff": "^3.2.0",
                 "grapheme-splitter": "^1.0.4",
                 "lodash": "^4.17.21",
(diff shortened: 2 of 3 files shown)


ollien:
Reverting this will require also reverting https://github.com/vscode-neovim/vscode-neovim/pull/2087

How quickly does your internal corporate cycle allow for updates? I am hesitant to speak on behalf of the project with a firm yes or no, but some data would help inform the decision.

antonk52:
I am not aware of consistent update cycle so I might be wrong. I presume we are usually 2 to 10 versions behind the most recent once.

theol0403:
What happens if you install with VSIX? I am hesitant to stop tracking the latest vscode version on behalf of a tiny minority of users.

antonk52:
> What happens if you install with VSIX? I am hesitant to stop tracking the latest vscode version on behalf of a tiny minority of users.

Fair point. Without going into details, unfortunately this is not allowed.

antonk52:
This might be quite an ask. As I mentioned the PR description as I do not want to add to the maintenance burden and my goal is not to for this project maintain a number of VS Code versions but instead to have one version that includes the `vscode/runtime/*.lua` files. Would it be possible to make a single plugin release with support for 1.89.x?

From my understanding the current master branch does not use any of a newer VS Code API and from the technical standpoint it should be possible. I understand that this can be a big ask and can take no for an answer.

theol0403:
Building and running seems to work fine for me, not sure about the potential node conflict here.

ollien:
@theol0403 Sorry for the delay in getting back to this. If we ease the version requirement to 1.89, while the extension itself will not have any problems necessarily, the TSConfig stating we support ES2023 will be incorrect, as we will be rolling back from Node 20 (introduced in 1.90) to Node 18. There are a couple of `Array` functions for ES2023 that are not implemented in Node 18 https://node.green/#ES2023

As I see it, we have three options (listed in order of preference)

1. Reject this PR, keeping things as they are today.
2. Accept this PR, but add a polyfill for the missing `Array` functions (e.g. ones from https://github.com/es-shims). They're generally quite small.
3. Accept this PR as-is, but revert https://github.com/vscode-neovim/vscode-neovim/pull/2087

I have a great deal of sympathy for the reporter's issue, as I've been on the other end of needlessly strict corporate requirements that break my workflow.  On the other hand, I do think that our current policy of tracking VSCode updates as soon as we can is a good one, and helps prevent some bitrot.

xiyaowong:
Two weeks have passed. Does this PR still matter?

antonk52:
> Two weeks have passed. Does this PR still matter?

Thanks for checking. Yes, this PR still matters.

Thank you for a detailed comment @ollien . One thing I can add is while the reason for this issue is still a niche of users of this extension. We have a few hundred potential users amongst my colleagues. While I cannot share an exact number I hope this shows a better ball park of the users who would appreciate a chance to use all the features as this group is still being stuck with 0.9.1 which is 50 versions behind the most recent one.

If you the maintainers are happy with either version 2 or 3, I am happy to send a follow up PR to make sure a missing Array method does not break in the runtime.

ollien:
I think I would be ok with #2, but I'll let @theol0403 make the final call. The question I have is how long we'd support 1.89 for. We can't stay there forever, but I don't think we've made use of any new features.

antonk52:
The goal of this PR was to have a chance for a single release with the all recent features. If it sticks a longer, that would be a bonus but I definitely do not want to add any maintenance burden.

theol0403:
I'm happy to merge this + make a one-time release once a polyfill is added in a PR.

antonk52:
I've updated the PR to include an Array.prototype.findLast polyfill and tested the extension with VS Code 1.89

This PR is good to get merged

theol0403 (approved):

package.json
```diff
@@ -2140,7 +2140,7 @@
         "@types/mocha": "^10.0.7",
         "@types/node": "^20.14.9",
         "@types/sinonjs__fake-timers": "^8.1.5",
-        "@types/vscode": "^1.90.0",
+        "@types/vscode": "1.89.0",
```
dustinfarris, line 2143: Should this have been `"^1.89.0"`?

ollien: OOF, that's a great catch. I think we need to revert this PR anyway to go back to our old engine requirement, making this moot, but it increases the urgency of us doing so. I'll take a look at doing this later today

antonk52: This was a correct change as the purpose of the pr was to make the extension available for 1.89. Specifying dependency version with a ^ would mean a newer types could be picked up from npm and type script would not error on a newer api usage. 

Having said that this safe to return to the previous state as the PR will be reverted for the next release.

PR #2167, opened 2024-07-18
refactor: Improve usability of afterInitConfig option in the settings

nottnottloop:
I spent around 15 minutes wondering why my Vimscript wasn't working to ignore case. Below you can see the settings for the current version of the extension, the top is the correct line and the bottom is the line I was writing:
image

Which maps to this in settings.json
image

I think I would have easily understood what to do immediately if the instructions looked like this instead:
Screenshot 2024-07-18 143434

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
                         "type": "string"
                     },
                     "title": "After init config",
-                    "markdownDescription": "Config to be sourced after init config. e.g. : `[\"echo 'hello'\", \"lua << EOF\", \"require'vscode'.notify('world')\", \"EOF\"]`"
+                    "markdownDescription": "Config to be sourced after init config. Each item is a different line of the config. For example:\n\n`set ignorecase`\n\n`echo 'hello'`\n\n`lua << EOF`\n\n`require'vscode'.notify('world')`\n\n`EOF`"
                 },
                 "vscode-neovim.neovimClean": {
                     "type": "boolean",


package-lock.json
```diff
@@ -42,7 +42,7 @@
                 "webpack-cli": "^5.1.4"
             },
             "engines": {
-                "vscode": "^1.87.0"
```
ollien, line 45: Is this necessary as part of your changeset?

ollien (commented):
LGTM, barring the errant(?) package lock changes. Thanks for the contribution!

nottnottloop:
> LGTM, barring the errant(?) package lock changes. Thanks for the contribution!

Sorry I'm not very good at this. My bad!!

nottnottloop:
Package-lock changes are removed from my branch

xiyaowong:
The modified example is incorrect.

nottnottloop:
> The modified example is incorrect. The `EOF` is missing.

Should it have the original array or should it have the original array with `set ignorecase` in it as well, what do you think

package.json
```diff
@@ -82,7 +82,7 @@
                         "type": "string"
                     },
                     "title": "After init config",
-                    "markdownDescription": "Config to be sourced after init config. e.g. : `[\"echo 'hello'\", \"lua << EOF\", \"require'vscode'.notify('world')\", \"EOF\"]`"
+                    "markdownDescription": "Config to be sourced after init config. Each item is a different line of the config. For example:\n\n`echo 'hello'`\n\n`lua << EOF`\n\n`require'vscode'.notify('world')`\n\n`set ignorecase`"
```
xiyaowong, line 85:
```suggestion
                    "markdownDescription": "Config to be sourced after init config. Each item is a different line of the config. For example:\n\n`echo 'hello'`\n\n`lua << EOF`\n\n`require'vscode'.notify('world')`\n\n`EOF`"
```

xiyaowong (changes requested):

xiyaowong:
> I wanted to demonstrate another example that might be useful

This is okay, just make sure the example works correctly.

nottnottloop:
After latest change
image

xiyaowong (approved):
How about putting `set ignorecase` on the first line? 👀

nottnottloop:
> How about putting `set ignorecase` on the first line? 👀

image
No worries, I love a good `set ignorecase`. Wait, did you mean only four lines with `echo 'hello'` removed or the whole five lines like above?

In case it isn't obvious I only use Vim motions, and have never really gone deep on Neovim configuration properly, so I didn't know what I was changing really w.r.t `EOF`.

xiyaowong:
I just realized where the original misleading aspect was. VSCode offers "Add Item". I've hardly used the settings UI. I think other options also need to consider this issue. Contributions are welcome.

PR #2169, opened 2024-07-18
fix: Fix multi-byte character selection in Nvim 0.10 blockwise visual

### LINKED ISSUE #2166: In Nvim 0.10, block visual mode cannot correctly select CJK characters.
image

upstream issue: https://github.com/neovim/neovim/issues/29786

xiyaowong:
Since Nvim 0.10.0, `virtcol2col` changed from returning the last byte of a multi-byte character to returning the first byte. However, we need the column of the last byte, which is consistent with the selected region in Nvim.

Close #2166

### DIFF
--- a/runtime/lua/vscode/util.lua
+++ b/runtime/lua/vscode/util.lua
@@ -1,21 +1,21 @@
 local M = {}
 
-local api = vim.api
+local fn, api = vim.fn, vim.api
 
 function M.is_visual_mode()
-  local mode = vim.api.nvim_get_mode().mode
+  local mode = api.nvim_get_mode().mode
   return mode == "v" or mode == "V" or mode == "\x16"
 end
 
 function M.get_char_at(line, byte_col, buf)
   if not buf or buf == 0 then
-    buf = vim.api.nvim_get_current_buf()
+    buf = api.nvim_get_current_buf()
   end
-  local line_str = vim.fn.getbufoneline(buf, line)
-  local char_idx = vim.fn.charidx(line_str, (byte_col - 1))
-  local char_nr = vim.fn.strgetchar(line_str, char_idx)
+  local line_str = fn.getbufoneline(buf, line)
+  local char_idx = fn.charidx(line_str, (byte_col - 1))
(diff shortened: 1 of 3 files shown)


xiyaowong:
Is there any questions? There is a detailed discussion in the linked Neovim issue.

runtime/lua/vscode/util.lua
```diff
@@ -79,4 +79,22 @@ function M.debounce(func, time)
   end
 end
 
+-- Since Nvim 0.10.0, `virtcol2col` changed from returning the last byte of a
+-- multi-byte character to returning the first byte. However, we need the column
+-- of the last byte, which is consistent with the selected region in Nvim.
+-- See https://github.com/neovim/neovim/issues/29786
+if fn.has("nvim-0.10.0") == 0 then
```
ollien, line 86: Will this break with nvim bumps the batch version?

ollien:
@xiyaowong I was at work and couldn't answer until now, sorry :)

runtime/lua/vscode/util.lua
xiyaowong: I didn't get what you meant.

ollien: This is checking whether or not the user is on nvim 0.10.0, right? What happens when a user upgrades to 0.10.1?

xiyaowong: It's OK, `:h has`. It's called "has", not "is" 😃

ollien: Yeah but wouldn't you want the substring to be "0.10."? (No trailing zero?) It might be better to parse this and see if it's 0.10 or greater, so we don't have to bump this when 0.11 releases

xiyaowong: has('nvim-0.10') and has('nvim-0.10.0') are the same; both indicate version 0.10.0 or higher. I prefer to use the full version number.

ollien: Ah, I misunderstood `has` and assumed it was a substring match!

ollien (approved):

PR #2177, opened 2024-07-23
refactor: follow nvim ui-messages protocol

### LINKED ISSUE #2134: Output channel: `ui-messages` handling, `MessagesManager` logging
Hi, I know this has been an area that's changed a lot recently, so hopefully this isn't stirring up too much trouble, but I noticed that the `ui-messages` handling seems to be slightly inconsistent at times, and I think I know why. I can probably open a PR to address it, but I wanted to check that the approach seems agreeable first.

## The issue

A simple example that shows the kind of thing I'm talking about. With `neovimClean` set, I source this file:
```lua
local i = 0

function do_echo()
    vim.api.nvim_command('echom "' .. vim.inspect(i) .. ' line 1\nline 2\nline 3"')
    i = i + 1

    if i < 5 then
        vim.defer_fn(do_echo, 1500)
    end
end

do_echo()
```

The `vscode-neovim messages` channel only appears to show one update, which (maybe comes from modeline?) is displayed as `1 line 1^@line 2^@line 3`. It's possible to see the messages with `:messages`, but I would expect the output panel to be updated as well (as in `nvim`).

The `vscode-neovim logs` panel is interesting here. The messages don't appear to be logged, except for the visible one:
```
2024-06-30 17:30:57.363 [info] MessagesManager: 1 line 1^@line 2^@line 3
2024-06-30 17:31:07.665 [debug] BufferManager: Editor viewColumn: undefined, winId: 1005, closing
2024-06-30 17:31:07.665 [debug] BufferManager: Document: output:asvetliakov.vscode-neovim.vscode-neovim%20logs, bufId: 5, deleting
2024-06-30 17:31:07.666 [debug] BufferManager: Visible editor, viewColumn: 1, doc: file:///Users/ianchamberlain/Downloads/repro.lua
2024-06-30 17:31:07.666 [debug] BufferManager: Visible editor, viewColumn: 2, doc: file:///Users/ianchamberlain/Documents/Development/vscode-neovim/src/messages_manager.ts
2024-06-30 17:31:07.669 [warning] ViewportManager: Unable to update scrolled view. No grid for winId: 1006
2024-06-30 17:31:11.963 [debug] BufferManager: Editor viewColumn: undefined, winId: 1006, closing
2024-06-30 17:31:11.964 [debug] BufferManager: Visible editor, viewColumn: 1, doc: file:///Users/ianchamberlain/Downloads/repro.lua
2024-06-30 17:31:11.964 [debug] BufferManager: Visible editor, viewColumn: 2, doc: file:///Users/ianchamberlain/Documents/Development/vscode-neovim/src/messages_manager.ts
2024-06-30 17:31:12.035 [warning] ViewportManager: Unable to update scrolled view. No grid for winId: 1007
````

However I can see them after running `:messages`, all concatenated together:

```
2024-06-30 17:33:59.180 [debug] ModeManager: Setting mode context to normal
2024-06-30 17:33:59.180 [info] MessagesManager: echomsg: 0 line 1
line 2
line 3
echomsg: 1 line 1
line 2
line 3
echomsg: 2 line 1
line 2
line 3
echomsg: 3 line 1
line 2
line 3
2024-06-30 17:33:59.182 [debug] StatusLineManager: Skipping statusline clear as a message is currently pending
```

## Why I think it happens

Reading `:help ui`, I think there are a couple potential problems here working together:
1. Messages are written immediately when `msg_show` is received, instead of being collected until a neovim `flush` event.
2. `msg_show` events have a boolean `replace_last`, which if set means:
  	> Replace the message in the most-recent `msg_show` call, but any other visible message should still remain.

  	Looking at the implementation `messages_manager.ts`, it seems like all previous messages in the notification would be cleared, instead of just the last one. 
3. From what I can tell, `msg_clear` is unimplemented except for statusline

For one example, looking at a `/` search that doesn't match, `nvim` normally shows two lines and then a prompt:
```
/asdfklj
E486: Pattern not found: asdfklj
Press ENTER or type command to continue
```

`vscode-neovim` only shows the second line, (third one omitted, related to https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2144175058 I think):

Screenshot 2024-06-30 at 17 52 44

## Proposed change

I believe it's possible to behave closer to vanilla `nvim`, if the message manager worked closer to how `ui-events` are describe. Effectively:
* Maintain a new internal message buffer
* `redraw`:
	* `msg_show`: append message string to a `string[]` buffer
		* if `replace_last`: replace last message in the buffer with the current one
		* Maybe do something different with `return_prompt` here too?
	* `msg_history_show`: replace internal buffer with incoming history messages. Maybe this should use a separate buffer even?
	* `msg_clear`: clear internal buffer
* `flush-redraw`:
	* append internal buffer to output channel (and logs)
	* clear internal buffer

If all that sounds reasonable, I am happy to work on a PR for it! I think it's possible without too much refactoring outside of `messages_manager.ts`, and maybe will also help make the experience more consistent with `nvim`. Since this has been troublesome lately, it would need a lot of testing around things like `cmdheight` and the changes between 0.9 and 0.10.

Thanks! Let me know what y'all think.

### PR DESCRIPTION
**Pull request by ian-h-chamberlain**
Closes #2134 

`:help ui-messages` shows a lot more details about how these messages are meant to be handled. This change makes our handling match more closely to the expected behavior and now obeys *_clear types of messages as well.

This changes the behavior of message handling pretty significantly, so it probably deserves some more serious testing before merging into a release — especially given some of the issues that have shown up in the past re: `return_prompt` + `cmdheight`, etc.

Keeping this as a draft for now as I'm still seeing an issue in some case where `:messages` (i.e. `msg_history_show`) gets cleared immediately after writing, resulting in no visible output. Seems to be sent due to https://github.com/neovim/neovim/issues/12929 maybe we should just ignore it, or treat it differently?

- [x] make sure `:messages` works

Feel free to give feedback while this is still in draft though, I would like to check that the approach seems agreeable too.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,4 +1,4 @@
-export const NVIM_MIN_VERSION = "0.9.0";
+export const NVIM_MIN_VERSION = "0.10.0";
 
 export const GlyphChars = {
     COMMAND: "\u2318",
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -1,3 +1,5 @@
+import { inspect } from "util";
+
 import { Disposable, OutputChannel, window } from "vscode";
 
 import { EXT_NAME } from "./constants";
@@ -12,9 +14,28 @@ export class MessagesManager implements Disposable {
     private disposables: Disposable[] = [];
     private channel: OutputChannel;
 
+    private redrawing = Promise.resolve();
+
+    private revealOutput: boolean = false;
+    private replaceOutput: boolean = false;
+    private displayHistory: boolean = false;
+    private didChange: boolean = false;
+
+    private messageBuffer: string[] = [];
+    private historyBuffer: string[] = [];
+
     public constructor(private readonly main: MainController) {
         this.channel = window.createOutputChannel(`${EXT_NAME} messages`);
-        this.disposables.push(this.channel, eventBus.on("redraw", this.handleRedraw, this));
+
+        // Prevent concurrent redraw / flush by chaining them on a single promise:
+        const redrawHandler = eventBus.on("redraw", (e) => {
+            this.redrawing = this.redrawing.then(() => this.handleRedraw(e));
+        });
+        const flushHandler = eventBus.on("flush-redraw", () => {
+            this.redrawing = this.redrawing.then(() => this.handleFlush());
+        });
+
+        this.disposables.push(redrawHandler, flushHandler);
     }
 
     public dispose(): void {
@@ -24,62 +45,106 @@ export class MessagesManager implements Disposable {
     private async handleRedraw({ name, args }: EventBusData<"redraw">): Promise<void> {
         switch (name) {
             case "msg_show": {
-                const hasReturnPrompt = args.some(([type]) => type === "return_prompt");
-                const msg = args.reduce((acc, [type, content, clear]) => {
+                let lineCount = 0;
+
+                for (const [type, content, replaceLast] of args) {
                     // Ignore return_prompt
                     //
                     // A note to future readers: return_prompt is sent much more often with ui_messages. It may
                     // not do what you expect from what :help ui says, so be careful about using these events.
                     // See: https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2144175058
-                    if (type === "return_prompt") return acc;
-                    if (clear) return "";
-                    return acc + content.map((c) => c[1]).join("");
-                }, "");
-                const outputMsg = hasReturnPrompt ? msg.replace(/\n$/, "") : msg;
+                    if (type === "return_prompt") {
+                        this.replaceOutput = true;
+                        continue;
+                    }
 
-                this.writeMessage(outputMsg);
+                    // NOTE: we could also potentially handle e.g. `echoerr` differently here,
+                    // like logging at error level or displaying a toast etc.
 
-                const lineCount = outputMsg.split("\n").length;
-                const cmdheight = (await this.main.client.getOption("cmdheight")) as number;
-                // Before Nvim 0.10, cmdheight is unchangeable, and it's always 0.
-                if (lineCount > Math.max(1, cmdheight)) {
-                    this.channel.show(true);
+                    const text = content.map(([_attrId, chunk]) => chunk).join("");
+                    if (replaceLast) {
+                        this.messageBuffer.pop();
+                    }
+                    this.messageBuffer.push(text);
+
+                    lineCount += text.split("\n").length;
                 }
 
+                const cmdheight = (await this.main.client.getOption("cmdheight")) as number;
+                // Before Nvim 0.10, cmdheight is unchangeable, and it's always 0.
+                this.revealOutput ||= lineCount > Math.max(1, cmdheight);
+                break;
+            }
+
+            case "msg_clear": {
+                this.messageBuffer = [];
                 break;
             }
 
             case "msg_history_show": {
-                const lines = [];
                 for (const arg of args) {
                     for (const list of arg) {
                         for (const [commandName, content] of list) {
-                            const cmdContent = content.map((c) => c[1]).join("");
+                            const cmdContent = content.map(([_attrId, chunk]) => chunk).join("");
 
                             if (commandName.length === 0) {
-                                lines.push(cmdContent);
+                                this.historyBuffer.push(cmdContent);
                             } else {
-                                lines.push(`${commandName}: ${cmdContent}`);
+                                this.historyBuffer.push(`${commandName}: ${cmdContent}`);
                             }
                         }
                     }
                 }
 
-                this.channel.show(true);
-                this.writeMessage(lines.join("\n"));
+                this.displayHistory = true;
+                this.replaceOutput = true;
+                this.revealOutput = true;
                 break;
             }
+
+            case "msg_history_clear": {
+                this.historyBuffer = [];
+                this.replaceOutput = true;
+                break;
+            }
+
+            default:
+                return;
         }
+
+        this.didChange = true;
+        logger.trace(name, inspect(args, { depth: 5, compact: 3 }));
+    }
+
+    private handleFlush(): void {
+        if (!this.didChange) return;
+
+        const messages = this.displayHistory ? this.historyBuffer : this.messageBuffer;
+        logger.trace(`Flushing ${this.displayHistory ? "history " : ""}message buffer: ${inspect(messages)}`);
+        const msg = this.ensureEOL(messages.join("\n"));
+
+        this.writeMessage(msg);
+        if (this.revealOutput) {
+            this.channel.show(true);
+        }
+
+        // Reset all the state for the next batch of redraw messages
+        this.didChange = false;
+        this.displayHistory = false;
+        this.revealOutput = false;
+        this.replaceOutput = false;
     }
 
     private writeMessage(msg: string): void {
-        if (msg.length === 0) {
-            return;
+        if (msg.length > 0) {
+            logger.info(inspect(msg));
         }
 
-        logger.info(msg);
-        const outputMsg = this.ensureEOL(msg);
-        this.channel.replace(outputMsg);
+        if (this.replaceOutput) {
+            this.channel.replace(msg);
+        } else if (msg.length !== 0) {
+            this.channel.append(msg);
+        }
     }
 
     private ensureEOL(msg: string): string {
--- a/src/messages_manager.ts
+++ b/src/messages_manager.ts
@@ -17,7 +17,6 @@ export class MessagesManager implements Disposable {
     private redrawing = Promise.resolve();
 
     private revealOutput: boolean = false;
-    private replaceOutput: boolean = false;
     private displayHistory: boolean = false;
     private didChange: boolean = false;
 
@@ -42,18 +41,17 @@ export class MessagesManager implements Disposable {
         disposeAll(this.disposables);
     }
 
-    private async handleRedraw({ name, args }: EventBusData<"redraw">): Promise<void> {
+    private handleRedraw({ name, args }: EventBusData<"redraw">): void {
         switch (name) {
             case "msg_show": {
-                let lineCount = 0;
                 for (const [type, content, replaceLast] of args) {
-                    // Ignore return_prompt
-                    //
                     // A note to future readers: return_prompt is sent much more often with ui_messages. It may
                     // not do what you expect from what :help ui says, so be careful about using these events.
                     // See: https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2144175058
                     if (type === "return_prompt") {
-                        this.replaceOutput = true;
+                        // This kinda mimics normal neovim behavior, but it's not exactly the
+                        // same because we still don't require a keypress/hide the panel afterwards.
+                        this.revealOutput = true;
                         continue;
                     }
 
@@ -65,13 +63,7 @@ export class MessagesManager implements Disposable {
                         this.messageBuffer.pop();
                     }
                     this.messageBuffer.push(text);
-
-                    lineCount += text.split("\n").length;
                 }
-
-                const cmdheight = (await this.main.client.getOption("cmdheight")) as number;
-                // Before Nvim 0.10, cmdheight is unchangeable, and it's always 0.
-                this.revealOutput ||= lineCount > Math.max(1, cmdheight);
                 break;
             }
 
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/messages_manager.ts [hunk 1]
```diff
@@ -12,9 +14,28 @@ export class MessagesManager implements Disposable {
...
+    private messageBuffer: string[] = [];
+    private historyBuffer: string[] = [];
+
     public constructor(private readonly main: MainController) {
         this.channel = window.createOutputChannel(`${EXT_NAME} messages`);
-        this.disposables.push(this.channel, eventBus.on("redraw", this.handleRedraw, this));
+
+        // Prevent concurrent redraw / flush by chaining them on a single promise:
+        const redrawHandler = eventBus.on("redraw", (e) => {
+            this.redrawing = this.redrawing.then(() => this.handleRedraw(e));
```
ollien, line 32: I don't think you need this promise. The JS event loop works such that if a `redraw` and `flush-redraw` come in one after the other, one callback will complete execution before the other begins.

Clever way of doing mutal exclusion, though!

src/messages_manager.ts
```diff
@@ -24,62 +45,106 @@ export class MessagesManager implements Disposable {
...
                     // A note to future readers: return_prompt is sent much more often with ui_messages. It may
                     // not do what you expect from what :help ui says, so be careful about using these events.
                     // See: https://github.com/vscode-neovim/vscode-neovim/issues/2046#issuecomment-2144175058
-                    if (type === "return_prompt") return acc;
-                    if (clear) return "";
-                    return acc + content.map((c) => c[1]).join("");
-                }, "");
-                const outputMsg = hasReturnPrompt ? msg.replace(/\n$/, "") : msg;
+                    if (type === "return_prompt") {
+                        this.replaceOutput = true;
```
ollien, line 57: I think you might be falling victim to the same bug that I fell victim to (which caused me to write this comment). Reading the issue you linked in the PR description, it seems that the `msg_history_clear` is sent following a `return_prompt`, yeah?

I'd also be curious to know how this affects something like "Pattern not found" messages. Are they still visible at the console at all, or are they immediately cleared?

src/messages_manager.ts
ian-h-chamberlain: Is that true? I guess `handleFlush()` will always complete execution, but `handleRedraw()` being async, wouldn't it normally yield at the first `await` point and potentially allow a flush to come in while waiting?

In a previous iteration I had both handlers as `async` so I think it was even more susceptible to this — I would expect execution to long potentially something like this? https://jsfiddle.net/mwboatyk/

src/messages_manager.ts
ian-h-chamberlain: > Reading the issue you linked in the PR description, it seems that the `msg_history_clear` is sent following a `return_prompt`, yeah?

From what I can tell, that's usually true? I guess maybe the clear would normally be sent after the user pressed Enter in response to that message. Actually, that's an interesting find — maybe it would makes sense to have the `return_prompt` "cancel out" a subsequent `msg_clear`. 

I kept iterating on this a little after I pushed the draft, and I found that it worked fairly well to simply skip setting `didChange = true` when a clear comes in — so effectively, we'd still clear the buffer, but avoid redrawing on the next flush, so the user could still see the output in the channel. Maybe limiting that to only happen after a `return_prompt` would be more correct.


> I'd also be curious to know how this affects something like "Pattern not found" messages. Are they still visible at the console at all, or are they immediately cleared?

Yeah, these are followed by a clear command almost immediately as well (with the same `return_prompt`):
```
2024-07-24T13:50:25.500Z MessagesManager: msg_show [
  [ 'wmsg', [ [ 98, 'search hit BOTTOM, continuing at TOP' ] ], true ],
  [
    'emsg',
    [ [ 78, 'E486: Pattern not found: aslkdjfhlasdk' ] ],
    false
  ],
  [
    'return_prompt',
    [ [ 153, 'Press ENTER or type command to continue' ] ],
    false
  ]
]
extensionHostProcess.js:148
2024-07-24T13:50:25.501Z MessagesManager: Flushing message buffer: [
  'search hit BOTTOM, continuing at TOP',
  'E486: Pattern not found: aslkdjfhlasdk'
]
extensionHostProcess.js:148
2024-07-24T13:50:25.502Z MessagesManager: 'search hit BOTTOM, continuing at TOP\nE486: Pattern not found: aslkdjfhlasdk\n'
extensionHostProcess.js:148
2024-07-24T13:50:25.503Z MessagesManager: msg_clear [ [] ]
extensionHostProcess.js:148
2024-07-24T13:50:25.505Z MessagesManager: Flushing message buffer: []
```

src/messages_manager.ts
ollien: You're totally right. I had a false memory that this was synchronous (it was when I was last here).

src/test/integ/messages.test.ts
```diff
@@ -0,0 +1,123 @@
...
+        await wait();
+        outputEditor = findOutputChannel();
+        assert.equal(outputEditor?.document.getText(), "");
+    });
+
+    it("should reveal for return prompt", async () => {
+        await sendNeovimKeys(client, "/foobar\n");
+        await wait();
+        let outputEditor = findOutputChannel();
+        assert.equal(outputEditor?.document.getText(), "/foobar             \nE486: Pattern not found: foobar\n");
```
ollien, line 116: I totally get why you made this change (I did it, too), but this is likely going to frustrate a lot of users. https://github.com/vscode-neovim/vscode-neovim/issues/2046

Unfortunately, it seems sane to do what you're doing (`return_prompt` implies we should have something meaningful displayed to the user), but when `ext_messages` is enabled, nvim sends us lots of `return_prompt`s that one wouldn't otherwise expect (e.g, after a missing pattern). This is ultimately why I just ignored `return_prompt` in the end :/

ian-h-chamberlain: Oops, you're right, in my arrogance I thought these changes would work around that issue but I forgot to actually compare `/`-search behavior with nvim itself and also prioor version of the extension. A couple cases I should probably test for both 0.9 and 0.10:

- [ ] search with default nvim configuration
- [ ] search with misc. `cmdheight` settings
- [ ] maybe try different `shortmess` settings? 
- [ ] Are there other easy ways of triggering the hit-enter prompt??

In #2029 you suggested possibly adding a configuration setting for hit-enter prompt behavior... given this keeps coming up, do you think it makes sense to consider adding this now? For this PR I will try to maintain the behavior as before (to prevent another #2046), but maybe a follow-up could include allowing users to opt-in to hit-enter prompts revealing the output panel, plus maybe some docs about `shortmess` would help?

ollien: > In https://github.com/vscode-neovim/vscode-neovim/pull/2029 you suggested possibly adding a configuration setting for hit-enter prompt behavior... given this keeps coming up, do you think it makes sense to consider adding this now?

I'm not sure, but I'm leaning towards no. If nvim's return_prompt behavior was consistent between clients, maybe (which was my understanding when I wrote that), but I suspect most users don't want an output panel for literally every output.

ian-h-chamberlain: Fair enough... the good news is it looks like upstream they are planning to overhaul this and reduce the frequency significantly: https://github.com/neovim/neovim/issues/22478

I'm not sure what the exact implications for external `ui-messages` consumers will be, but maybe it will help with this problem in general.

ian-h-chamberlain: hoo boy, okay. It turns out even after removing the `return_prompt` handling, this PR will still reveal on failed searches for anyone whose `cmdheight` is < 2, because there's no `msg_clear` between the `/<pattern>` message and the following error message, and `replace_last` is also set to false 😭 

```
msg_clear [ [] ]
msg_show [ [ '', [ [ 0, '/asdkfjhqwe             ' ] ], false ] ]
Flushing message buffer: [ '/asdkfjhqwe             ' ]
msg_show [
  [ 'emsg', [ [ 90, 'E486: Pattern not found: asdkfjhqwe' ] ], false ],
  [
    'return_prompt',
    [ [ 166, 'Press ENTER or type command to continue' ] ],
    false
  ]
]
Flushing message buffer: [ '/asdkfjhqwe             ', 'E486: Pattern not found: asdkfjhqwe' ]
```

I guess it's possible to set the default `cmdheight` back to 2, kinda like the behavior before 4ca8fbf70edfaaf080823d583e8d2105ace0c06e ? It might still be a breaking change for people who've got `cmdheight` explicitly set in their configuration in that case... 

The only other option I can think of is trying to display a hint the first time to let people know they can increase cmdheight to suppress the output panel, or something like that, to help avoid the influx of unhappy users filing issues... I'll keep thinking about different possible options I guess

ollien: I've been thinking about this, and I think that if we're going to tell users to customize cmdheight, is there much difference between it and us making a configuration value? I'm not sure which I prefer. I think `cmdheight` makes the most sense, because then the solution is fully in nvim's control.

ian-h-chamberlain: Ok, I tried to implement this configuration hint popup and it mostly works, but I realized while testing against 0.9.5 that since `cmdheight` isn't configurable for the attached UI until 0.10, we always end up with `cmdheight=0` and there's no way to suppress the output reveal. I suppose that might be an argument in favor of a `vscode-neovim` configuration? 

Alternatively, maybe just suggest they upgrade to 0.10+ to make `cmdheight` work as expected is simpler, idk.

ollien: Maybe a good compromise is to make an option that accepts a minimum number of lines before output is revealed, with a default of "cmdheight", (or null if necessary), which will fall back to the nvim value

ollien: Also, I hope it goes without saying, but I appreciate the care you've put into this!!

ollien (commented):
This all seems fine to me. I'd like to get your thoughts on [this](https://github.com/vscode-neovim/vscode-neovim/pull/2177/files#r1720825017) before I approve, but if you don't think it's worth it, that's totally cool and I won't block on it :)

justinmk:
Please, let's not add new options. Especially when there is already an existing Nvim option. 

> Alternatively, maybe just suggest they upgrade to 0.10+ to make `cmdheight`

Yes, for pete's sake. Supporting users that can't bother to do a one-time update is not worth our time. They are constantly updating vscode but they can't bother to update Nvim?

ollien (approved):

src/messages_manager.ts
```diff
@@ -8,78 +10,190 @@ import { disposeAll } from "./utils";
...
+        const picked = await window.showInformationMessage(
+            `The vscode-neovim output panel will be revealed for output longer than
+            ${cmdheight} line(s). To suppress output for messages from searching with /,
+            use neovim version 0.10+ and configure cmdheight=2 or greater.`,
+            "Ok",
+            suppress,
+        );
+
+        if (picked === suppress) {
+            this.context.globalState.update(SUPPRESS_CMDHEIGHT_MSG, true);
```
justinmk, line 183: can we just bump our required Nvim version? that eliminates this code and avoids noise in globalState.

theol0403 (commented):
Yes, I would rather just require nvim 0.10 than dance around old versions.

justinmk:
Seems this PR is nearly ready to merge, we're just waiting for the old-Nvim-version code to be deleted?

You can also bump the required Nvim version to 0.10 here: https://github.com/vscode-neovim/vscode-neovim/blob/e9d6584f324e6a5725305c27610dcbf722b1e9f8/src/constants.ts#L1

src/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const NVIM_MIN_VERSION = "0.9.0";
+export const NVIM_MIN_VERSION = "0.10.0";
```
ollien, line 1: Can you also update the README?

ollien (approved):

theol0403:
Thanks for the work here, let me know when it is all ready to merge

ian-h-chamberlain:
> Thanks for the work here, let me know when it is all ready to merge

@theol0403 should be good now I think, I just had to push some fixes for the test / commit message format. Sorry for the delay getting it across the finish line!

cgjosephlee:
Thanks all your efforts. Isn't dropping support of nvim 0.9 worth a greater version bumping (e.g. -> 0.19) and a PSA?

justinmk:
> Isn't dropping support of nvim 0.9 worth a greater version bumping (e.g. -> 1.19) and a PSA?

We should have mentioned it in the changelog. But other than that, the "PSA" is already present in the form of an automated message that is displayed at startup?

cgjosephlee:
I was using nvim 0.9.4 and I first encountered issue #2227 before seeing any automated message. Then I reverted to v1.18.9.

dotai2012:
Sorry but ever since I updated to this version 1.18.10, my VScode has been popping messages from neovim non stop.

image

ollien:
@dotai2012 Please file an issue if you're experiencing trouble. That said, please be mindful of the nvim version requirement introduced by this PR.

PR #2178, opened 2024-07-24
fix: directly replace buffer lines for auto-generated docs

### LINKED ISSUE #2175: bug: nvim_win_set_cursor: Cursor position outside buffer
### Check the following:

### Neovim version (nvim -v)

0.9.5

### Operating system/version

macOS 14

### Describe the bug

When using with GitHub Copilot, clicking the answer box constantly throws `nvim_win_set_cursor: Cursor position outside buffer`.

### Steps To Reproduce

https://github.com/user-attachments/assets/1fc2cb50-5a5f-40b1-becd-b6b76d57eeae

### Expected Behavior

No errors.

xiyaowong:
For auto-generated documents, we directly replace the content. These documents frequently change rapidly and in large volumes, which can lead to synchronization issues and errors such as cursor misalignment. Additionally, we do not process dotrepeat or cursor updates, as these are unnecessary.

fix #2175


theol0403:
I think this is much too much of a hack. Why are there synchronization issues? Is nvim_buf_set_text too slow? I don't think so. Why are there cursor issues? This seems like a larger problem. Why do we need to explicitly skip processing dotrepeat? In these editors, there are no insert modes, so dotrepeat hack is never run anyway. Why skip cursor updates? If the only communication is vscode -> nvim, the default behavior is fine.

xiyaowong:
> I think this is much too much of a hack. Why are there synchronization issues? Is nvim_buf_set_text too slow? I don't think so. Why are there cursor issues? This seems like a larger problem. 

I don't think this is a hack; it's just a different synchronization strategy for specific documents. However, I agree that the current issue hasn't been identified yet.

The cursor error is caused by content inconsistency, leading to the cursor set for nvim going out of range. Having debugged similar issues before, I know it's due to content synchronization errors. Initially, I considered checking the nvim_buf_set_text issue, but I realized that dotrepeat and cursor updates are unnecessary, so I came up with this solution.

> In these editors, there are no insert modes.

Incorrect. You can switch to this type of editor in insert mode, and in vscode-chat-code-block, it is possible to switch to insert mode.

> Why skip cursor updates? If the only communication is vscode -> nvim, the default behavior is fine.

Since this type of editor changes very quickly, it frequently triggers cursor synchronization between VSCode and Neovim. In the cursor synchronization logic, we also need to wait for the synchronization process to complete.

PR #2183, opened 2024-07-25
fix: queue text document change events

### LINKED ISSUE #2175: bug: nvim_win_set_cursor: Cursor position outside buffer

xiyaowong:
During the initialization process, the document may continue to change, so by the end of initialization, our synchronized content is outdated and has lost the change events. This PR always caches document change events, filters out the outdated events after initialization, and then performs synchronization.

Please also check the code comments.

fix #2175


xiyaowong:
Added description :)

src/document_change_manager.ts
```diff
@@ -405,16 +461,18 @@ export class DocumentChangeManager implements Disposable {
 
         await actions.lua("handle_changes", bufId, changeArgs);
 
+        if (!isDirty) {
+            await this.client.request("nvim_buf_set_option", [bufId, "modified", false]);
```
theol0403, line 465: Why is this line also in 412?

theol0403 (approved):

src/document_change_manager.ts
xiyaowong: The following events may occur:

1. The version remains unchanged, `contentChanges` is empty, and `isDirty` is false (the logic will not reach line 465, so `modified` needs to be set).
2. The version changes, `contentChanges` is not empty, and `isDirty` is false (this is an undo operation; therefore, after applying the changes, `modified` needs to be set manually).

PR #2196, opened 2024-08-09
Revert "Ease engines requirement"

ollien:
Reverts vscode-neovim/vscode-neovim#2163

The singular support release has been completed, so we can roll back to our higher engine version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
             "dependencies": {
                 "@jpwilliams/waitgroup": "^2.1.1",
                 "async-mutex": "^0.5.0",
-                "core-js": "^3.37.1",
                 "fast-myers-diff": "^3.2.0",
                 "grapheme-splitter": "^1.0.4",
                 "lodash": "^4.17.21",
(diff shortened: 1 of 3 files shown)


xiyaowong:
Isn't this reintroducing the issue described in #2163? I don't think it's necessary to update to the latest version of VS Code since we aren't using the new APIs. However, I suggest removing core-js and using findLast from lodash instead.

ollien:
Yes, you're correct. At minimum, we need to change the version requirement to `^1.89.0`, instead of just `1.89.0` However, as @theol0403 suggested, this was meant to be a one-time thing. I think in general it's likely a good thing for us to stay as up to date as possible

That said, I wouldn't mind staying on `^1.89.0` until we need a new feature, so long as we revert https://github.com/vscode-neovim/vscode-neovim/pull/2087 so our tsconfig is still correct. We shouldn't declare that we can compile with ES2023 if we can't actually.

xiyaowong:
It should be modified in a few months, and upgrading is not necessary unless absolutely required. As for the version requirement to ^1.89.0 you mentioned, this is just a devDependence, and for now, changing it doesn't have any impact.

ollien:
> It should be modified in a few months, and upgrading is not necessary unless absolutely required.

If that's our stance, we should roll back ES2023 support, Node 18 (which 1.89 runs) does not support all of its features

theol0403 (approved):

theol0403:
In the future, we can upgrade to newer VSCode engines more slowly. For now, we will target the newest version because we have code which depends on it.

PR #2215, opened 2024-08-26
docs: update README.md with snap installation details

jamesikanos:
I've installed nvim as a snap package and VSCode could not detect automatically. This is because the Ubuntu 2204 package is 0.6, which is further behind than this extension recommends. 

A bit of googling found that snap stores the binary at: /snap/nvim/current/usr/bin/nvim

Updated the README as people may install from Ubuntu Store which uses snap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,13 @@ Install [Neovim](https://github.com/neovim/neovim/wiki/Installing-Neovim) **0.9.
 "`vscode-neovim.neovimExecutablePaths.win32/linux/darwin`", respective to your system. For example,
 "`C:\Neovim\bin\nvim.exe`" or "`/usr/local/bin/nvim`".
 
-> **Note:** If you want to use Neovim from WSL, set the `useWSL` configuration toggle and specify the Linux path to the
+> **WSL Users:** If you want to use Neovim from WSL, set the `useWSL` configuration toggle and specify the Linux path to the
 > nvim binary. `wsl.exe` Windows binary and `wslpath` Linux binary are required for this. `wslpath` must be available
 > through `$PATH` Linux env setting. Use `wsl --list` to check for the correct default Linux distribution.
+>
+> **Snap Users:** If you want to use Neovim from Snap, the Neovim path must be resolved to the snap binary location. On some
+> systems it might be "`/snap/nvim/current/usr/bin/nvim`". To check if you're running as a snap package, see if `which nvim`
+> resolves to `/usr/bin/snap`.
 
 ### Neovim configuration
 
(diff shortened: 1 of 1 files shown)


ollien:
Hi! Thanks for the PR :)

We don't support nvim 0.6, as you point out. Is this knowledge generally useful as snap upgrades their nvim version? I must admit I'm not the most familiar with it

jamesikanos:
Yes, I believe that snap packages are upgraded regularly. The "Snap Store" is available by default in Ubuntu and is probably a common mechanism for Ubuntu users to install Neovim.

So, even though the "nvim" binary is available in the $PATH (and is symlinked within /usr/bin), it doesn't get picked up by this extension because of the way that nvim resolves to the snap binary.

Of course, like many Open Source PR's, this is selfish as it would have helped me solve the issue if the Readme had mentioned what to do for Snap users. But, I could not comment as to how many people will have my configuration.

Either way, I'm happy with the extension and works as advertised!

theol0403 (approved):

theol0403:
I'm happy to merge once the formatting issues are fixed!

README.md
```diff
@@ -79,9 +79,13 @@ Install [Neovim](https://github.com/neovim/neovim/wiki/Installing-Neovim) **0.9.
 "`vscode-neovim.neovimExecutablePaths.win32/linux/darwin`", respective to your system. For example,
 "`C:\Neovim\bin\nvim.exe`" or "`/usr/local/bin/nvim`".
 
-> **Note:** If you want to use Neovim from WSL, set the `useWSL` configuration toggle and specify the Linux path to the
+> **WSL Users:** If you want to use Neovim from WSL, set the `useWSL` configuration toggle and specify the Linux path to the
 > nvim binary. `wsl.exe` Windows binary and `wslpath` Linux binary are required for this. `wslpath` must be available
 > through `$PATH` Linux env setting. Use `wsl --list` to check for the correct default Linux distribution.
+>
+> **Snap Users:** If you want to use Neovim from Snap, the Neovim path must be resolved to the snap binary location. On some
+> systems it might be "`/snap/nvim/current/usr/bin/nvim`". To check if you're running as a snap package, see if `which nvim`
```
justinmk, line 87: node-client could check this in `findNvim()`, then users wouldn't need to manually configure it: https://github.com/neovim/node-client/issues/267

jamesikanos:
> I'm happy to merge once the formatting issues are fixed!

Have run this through Prettier and it made some modifications. Also, update the PR title in accordance with your conventions, please let me know if this is acceptable.

Issue #2136, opened 2024-07-01
vscode-neovim becomes unresponsive but nvim did *not* crash/disconnect


ian-h-chamberlain:
To be clear: this is a separate issue from #2000 and similar, because no "neovim disconnected" errors is displayed but the extension otherwise stops working. There are many similar / duplicate issues about neovim 0.10.0 crashing, please see those for workarounds etc.

---

Ok, last night I was reproducing this more consistently, on both `0.9.5` and `0.10.0`. Here are some of my notes about it, maybe they will help:

* No error message / notification is displayed when the problem occurs. Most of the time I was simply stuck in normal mode and unable to move the cursor or change modes, etc. 
  * In one case, I was stuck in macro-recording mode.
  * I have seen insert-mode before, but not typically
* Keybinds become unresponsive, but the extension is still active (logging, etc.) and the `neovim` process is still running:
   ```
   0.0  0.4  0:51.00 │  ├─ /Users/ianchamberlain/Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper (Plugin).app/Contents/MacOS/Code Helper (Plugin) --type=utility --utility-sub-type=node.mojom.NodeService --lang=en-US --service-sandbox-type=none --dns-resu
   0.0  0.1  0:00.00 │  │  └─ /opt/homebrew/bin/nvim -N --embed --cmd source /Users/ianchamberlain/.vscode/extensions/asvetliakov.vscode-neovim-1.17.2/runtime/vscode-neovim.vim
   ```
* `Developer: Toggle Keyboard Shortcuts Troubleshooting` showed it was still running `vscode-neovim.send` when pressing e.g. `hjkl`, so the context keys must have been set as for normal mode.

Extension logs (I believe the problem started around `2024-06-30T03:15:23.949Z`): [vscode-app-1719717602613.log](https://github.com/user-attachments/files/16046393/vscode-app-1719717602613.log)

Reproduction steps that were working for me almost 100% of the time, but don't seem to be working anymore. I'm not sure what changed, but this is basically what I had at the time:
1. 10-15 tabs open total, including several `:help` tabs (external buffers) 
2. search for phrase in workspace using VSCode shortcut (<kbd>cmd+shift+f</kbd> on macOS)    3. click a search result that opens a new editor (buffer that was not previously open)
4. click in the new editor
5. Observe stuck cursor / unresponsive, run `Neovim: Restart Extension` to go back to step 1

[...]

PR #2230, opened 2024-09-07
fix(buffers): don't await window/buffer cleanup

ian-h-chamberlain:
Workaround for #2136

This isn't really fixing the problem, since I haven't been able to figure out the root cause yet, but this should at least allow the cleanup handler to move on and restore control to the user after a few seconds.

Also got rid of the lua layer for the window cleanup, since we can just call the nvim APIs directly without an extra indirection (I had hoped that itself might fix the issue but alas did not).

Also ensure messages output channel is cleaned up by registering its `dipsose()` call. I think I broke this in #2177 (oops!)

----

Edit: for testing, I've been driving this a version of this change to see if it helped the UX and it did in fact help! My editor remained usable after the timeout, instead of forcing me to restart the extension like I usually do:
```
2024-09-07 12:39:06.197 [warning] BufferManager: [Error: nvim_win_close timed out after 5000ms
	at /Users/ianchamberlain/.vscode/extensions/asvetliakov.vscode-neovim-1.18.10-dev/dist/extension.js:2:334915]
2024-09-07 12:39:06.198 [debug] BufferManager: Visible editor, viewColumn: 2, doc: file:///Users/ianchamberlain/Documents/Development/lix/build/subprojects/rnix-0.11.0/meson.build
2024-09-07 12:39:06.198 [debug] BufferManager: Visible editor, viewColumn: 1, doc: file:///Users/ianchamberlain/Documents/Development/lix/doc/manual/src/release-notes/rl-2.90.md
2024-09-07 12:39:06.198 [debug] BufferManager: Document not known, init in neovim
2024-09-07 12:39:06.199 [debug] BufferManager: Init buffer for 90, doc: file:///Users/ianchamberlain/Documents/Development/lix/doc/manual/src/release-notes/rl-2.90.md
```

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -1,5 +1,4 @@
 import path from "path";
-import { inspect } from "util";
 
 import { debounce } from "lodash";
 import { Buffer, NeovimClient } from "neovim";
@@ -582,6 +581,7 @@ export class BufferManager implements Disposable {
 
                 if (token?.isCancellationRequested) continue;
                 this.syncLayoutProgress.report("Cleaning up windows and buffers");
+                // Intentionally not `awaited`, see comment in function for details:
                 this.cleanupWindowsAndBuffers(visibleEditors);
 
                 if (token?.isCancellationRequested) continue;
@@ -625,17 +625,16 @@ export class BufferManager implements Disposable {
             unusedBuffers.push(bufId);
         });
 
-        // We don't await the result of closing these as a workaround for #2136,
-        // so that the user isn't blocked forever in case the nvim client gets stuck.
-        if (unusedWindows.length) {
+        if (unusedWindows.length || unusedBuffers.length) {
+            const toCleanup = { windows: unusedWindows, buffers: unusedBuffers };
+            // Log if cleanup takes a long time, in case the request is never fulfilled
+            const logSlowCleanup = setTimeout(() => logger.warn("Cleanup took longer than 5s: ", toCleanup), 5000);
+            // We don't await the result of this cleanup as a workaround for #2136,
+            // so that the user isn't blocked forever in case the nvim client gets stuck.
             actions
-                .lua("close_windows", unusedWindows)
-                .catch((err) => logger.warn(`Failed to cleanup windows ${inspect(unusedWindows)}: ${err}`));
-        }
-        if (unusedBuffers.length) {
-            actions
-                .lua("delete_buffers", unusedBuffers)
-                .catch((err) => logger.warn(`Failed to cleanup buffers ${inspect(unusedBuffers)}: ${err}`));
+                .lua("cleanup_windows_and_buffers", toCleanup)
+                .catch((err) => logger.warn("Failed to cleanup", toCleanup, err))
+                .finally(() => clearTimeout(logSlowCleanup));
         }
     }
 
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -1,4 +1,5 @@
 import path from "path";
+import { inspect } from "util";
 
 import { debounce } from "lodash";
 import { Buffer, NeovimClient } from "neovim";
@@ -581,7 +582,7 @@ export class BufferManager implements Disposable {
 
                 if (token?.isCancellationRequested) continue;
                 this.syncLayoutProgress.report("Cleaning up windows and buffers");
-                await this.cleanupWindowsAndBuffers(visibleEditors);
+                this.cleanupWindowsAndBuffers(visibleEditors);
 
                 if (token?.isCancellationRequested) continue;
                 this.syncLayoutProgress.report("Syncing visible editors");
@@ -603,7 +604,7 @@ export class BufferManager implements Disposable {
 
     private syncEditorLayoutDebounced = debounce(this.syncEditorLayout, 100, { leading: false, trailing: true });
 
-    private async cleanupWindowsAndBuffers(visibleEditors: TextEditor[]): Promise<void> {
+    private cleanupWindowsAndBuffers(visibleEditors: TextEditor[]): void {
         const unusedWindows: number[] = [];
         const unusedBuffers: number[] = [];
         // close windows
@@ -614,6 +615,7 @@ export class BufferManager implements Disposable {
             this.winIdToEditor.delete(winId);
             unusedWindows.push(winId);
         });
+
         // delete buffers
         [...this.textDocumentToBufferId.entries()].forEach(([document, bufId]) => {
             if (!document.isClosed) return;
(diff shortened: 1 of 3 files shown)


src/buffer_manager.ts
```diff
@@ -622,8 +624,39 @@ export class BufferManager implements Disposable {
...
             unusedBuffers.push(bufId);
         });
-        unusedWindows.length && (await actions.lua("close_windows", unusedWindows));
-        unusedBuffers.length && (await actions.lua("delete_buffers", unusedBuffers));
+
+        logger.debug(`Closing windows: ${inspect(unusedWindows)}...`);
+        for (const winId of unusedWindows) {
+            logger.debug(`Closing window ${winId}...`);
+            try {
+                await this.request("nvim_win_close", [winId, true], 5000);
```
justinmk, line 632: The Lua code runs in the Nvim server. With your changes, now multiple rpc requests would be made (one request per window).

src/buffer_manager.ts
```diff
@@ -622,8 +624,39 @@ export class BufferManager implements Disposable {
...
+            }
+        }
+    }
+
+    // Workaround for #2136 - this request sometimes hangs indefinitely so we
+    // just hardcode it to return after some time, to restore functionality to the user.
+    private async request(name: string, args: any[], delayMs: number): Promise<void> {
+        return Promise.race([
+            this.client.request(name, args),
+            wait(delayMs).then(() => {
```
justinmk, line 656: Hmm. Maybe the [client](https://github.com/neovim/node-client/) `request()` should have a builtin way to support a "timeout" parameter.

I guess there's no way to cancel the request :/

src/buffer_manager.ts
ian-h-chamberlain: Ah, I guess that makes sense as a reason to reduce the overhead of the call. I guess one downside of batching like that would be that it is harder to tell _which_ call caused the timeout, unless the timeout mechanism itself were implemented in Lua and e.g. a list of results was returned...

src/buffer_manager.ts
```diff
@@ -622,8 +624,39 @@ export class BufferManager implements Disposable {
             this.textDocumentToBufferId.delete(document);
             unusedBuffers.push(bufId);
         });
-        unusedWindows.length && (await actions.lua("close_windows", unusedWindows));
```
justinmk, line 625: instead of a timeout parameter on `request()`, any reason to simply do the call asynchronoously? we don't use the result for anything.

    unusedWindows.length && (actions.lua("close_windows", unusedWindows).catch(e => log(e));

src/buffer_manager.ts
justinmk: > harder to tell _which_ call caused the timeout

well, `nvim_win_close` cannot hang or timeout. so the only thing that could be timing out is the RPC call.

ian-h-chamberlain: Hmm, I thought that when I had added prints on the Lua side (before bypassing it) that even the `pcall(vim.api.nvim_win_close, win, true)` was getting stuck, but it was a while ago and I don't know much about the nvim internals that are doing this work... 

Assuming it is the RPC, does that imply we should anticipate a possible hang like this on _any_ call and write the client to handle those cases? Would that be a neovim core bug or a `node-client` bug?

src/buffer_manager.ts
ian-h-chamberlain: I thought about this too, but I wasn't sure if it would actually be safer to handle the errors more "correctly" by keeping the editors/buffers around in the `this.textEditorToWinId` etc. maps (as I wrote in the TODO comments here). 

The other issue I thought of was that these promises might never resolve nor reject if they get stuck forever, so there could be a memory leak from just firing and forgetting them... I'm not sure how big a concern that is in practice, or if the GC is smart enough to deal with it. I guess even with `race()` this might be a problem, now that I think about it...

justinmk: Yeah, the `.catch()` approach seems simpler and clearer until we find a more robust solution.

src/buffer_manager.ts
justinmk: > does that imply we should anticipate a possible hang like this on _any_ call and write the client to handle those cases?

consumers like vscode-neovim shouldn't generally worry about that. I think https://github.com/neovim/node-client/ or Nvim should eventually try to provide a generalized timeout/watchdog mechanism.

src/buffer_manager.ts
```diff
@@ -622,8 +624,19 @@ export class BufferManager implements Disposable {
...
-        unusedBuffers.length && (await actions.lua("delete_buffers", unusedBuffers));
+
+        // We don't await the result of closing these as a workaround for #2136,
+        // so that the user isn't blocked forever in case the nvim client gets stuck.
+        if (unusedWindows.length) {
+            actions
+                .lua("close_windows", unusedWindows)
+                .catch((err) => logger.warn(`Failed to cleanup windows ${inspect(unusedWindows)}: ${err}`));
```
xiyaowong, line 633:
```suggestion
                .catch((err) => logger.warn("Failed to cleanup windows", unusedWindows, err));
```
The logger has already used inspect.

xiyaowong (changes requested):

xiyaowong:
I'm not sure if there are any side effects, but logically, the cleanup must be completed, and the window must be cleaned up before the buffer.

ian-h-chamberlain:
> I'm not sure if there are any side effects, but logically, the cleanup must be completed, and the window must be cleaned up before the buffer.

@xiyaowong I tried to resolve this by combining the buffer + window cleanup into one call so they would be executed in order, let me know what you think.

I also added back a simple log in case it takes >5s to do the cleanup, so it's not just stuck forever silently at least.

src/buffer_manager.ts
```diff
@@ -622,8 +623,18 @@ export class BufferManager implements Disposable {
...
-        unusedBuffers.length && (await actions.lua("delete_buffers", unusedBuffers));
+
+        if (unusedWindows.length || unusedBuffers.length) {
+            const toCleanup = { windows: unusedWindows, buffers: unusedBuffers };
+            // Log if cleanup takes a long time, in case the request is never fulfilled
+            const logSlowCleanup = setTimeout(() => logger.warn("Cleanup took longer than 5s: ", toCleanup), 5000);
+            // We don't await the result of this cleanup as a workaround for #2136,
+            // so that the user isn't blocked forever in case the nvim client gets stuck.
+            actions.lua("cleanup_windows_and_buffers", toCleanup).then(
+                () => clearTimeout(logSlowCleanup),
```
justinmk, line 634: nice approach. does this run on error or do we also need clearTimeout for the error case?

ian-h-chamberlain: Good point, moved this to a `finally` to ensure it's cleared in both cases

src/buffer_manager.ts
```diff
@@ -581,7 +581,7 @@ export class BufferManager implements Disposable {
 
                 if (token?.isCancellationRequested) continue;
                 this.syncLayoutProgress.report("Cleaning up windows and buffers");
-                await this.cleanupWindowsAndBuffers(visibleEditors);
+                this.cleanupWindowsAndBuffers(visibleEditors);
```
xiyaowong, line 584: Here might be an issue; it is recommended to add a comment.

xiyaowong (approved):

justinmk (approved):

src/buffer_manager.ts
ian-h-chamberlain: Sure, what should the comment say? I guess I'm not sure what issue you are referring to here

ian-h-chamberlain: Ah, I thought that the comment I added inside `cleanupWindowsAndBuffers` would explain that, and I think I removed `await` here because the linter didn't like that I was awaiting a non-Promise. I can add something though

Issue #2184, opened 2024-07-25
bug: "Syncing layout: Syncing active editor" keeps running forever


wenfangdu:
### Check the following:

- [X] I have checked the behavior after setting "vscode-neovim.neovimClean" in VSCode settings.
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing issues.

### Neovim version (nvim -v)

0.9.5

### Operating system/version

macOS 14

### Describe the bug

Occasionally, when VSCode is idle, this notification pops up, and keeps running forever:
![image](https://github.com/user-attachments/assets/a5157491-3444-43af-a5a6-a74c1d766769)

### Expected Behavior

No notification.

xiyaowong:
Without reproduction steps, logs, screen recordings, or any information that can help identify and resolve the issue, I will close this issue.

wenfangdu:
@xiyaowong Indeed, it happens occasionally and automatically when idle, next time it happens, I'll collect the logs and paste them here.

stefan-bc-ts:
Here is what happend to me:
1. neovim diconected and I restart
2. get this msg
<img width="381" alt="Screenshot 2024-07-25 at 10 36 41 AM" src="https://github.com/user-attachments/assets/10597faf-7779-4adc-af94-3a58daa709a8">

pwnz22:
I'm also having same bug

dakesan:
> Here is what happend to me:
> 
> 1. neovim diconected and I restart
> 2. get this msg
> 
[...]

PR #2260, opened 2024-09-24
bug: Add configuration to skip syncing extension visible windows

andrewDoing:
Issue: https://github.com/vscode-neovim/vscode-neovim/issues/2184

This code change fixes the issue at the cost of being able to navigate the generated code windows in Copilot Chat. There is probably a more ideal solution where these windows are handled/synced with neovim in a faster way, but in my experience, I don't navigate these with vim anyway, and I'd rather have the neovim instance functioning.


xiyaowong:
This is not the solution. What we need to solve is why there will be problems when there are many editors.

andrewDoing:
I understand that there's a better solution, but what if I add a configuration field like "disableCopilotChat" that toggles this behavior? Once the performance of many windows is solved (maybe by me but probably by someone else) this code is easily findable and deletable.

xiyaowong:
I don't think there are performance issues here, whether it's the code blocks in Copilot chat or the large Git diffs; we've tested them before.

justinmk:
https://github.com/vscode-neovim/vscode-neovim/issues/2267#issuecomment-2381694486 may be a good place to start. It looks like we have an `OptionSet` handler that can trigger an infinite loop.

xiyaowong:
The editors were originally no different, but treating them separately would only increase the maintenance burden.

PR #2271, opened 2024-09-28
feat: ignoreSync

Malix-Labs:


justinmk:
Wow, didn't know about this vscode feature! This definitely looks useful. Thanks for pointing to the upstream docs too.

package.json
```diff
@@ -43,37 +43,43 @@
                     "type": "string",
                     "default": "nvim",
                     "title": "Neovim executable path on Windows",
-                    "markdownDescription": "Path to Nvim executable if running VS Code on Windows. \n- Example: `C:\\tools\\neovim\\Neovim\\bin\\nvim.exe` (if you installed using Chocolatey)\n- If you want to use Windows Subsystem for Linux, enable the `useWSL` setting and set `neovimExecutablePaths.linux` instead"
+                    "markdownDescription": "Path to Nvim executable if running VS Code on Windows. \n- Example: `C:\\tools\\neovim\\Neovim\\bin\\nvim.exe` (if you installed using Chocolatey)\n- If you want to use Windows Subsystem for Linux, enable the `useWSL` setting and set `neovimExecutablePaths.linux` instead",
+                    "ignoreSync": true
```
justinmk, line 47: for each case, would you mind moving the added line up so that the existing lines aren't changed. that makes the diff less noisy.
```suggestion
                    "ignoreSync": true,
                    "markdownDescription": "Path to Nvim executable if running VS Code on Windows. \n- Example: `C:\\tools\\neovim\\Neovim\\bin\\nvim.exe` (if you installed using Chocolatey)\n- If you want to use Windows Subsystem for Linux, enable the `useWSL` setting and set `neovimExecutablePaths.linux` instead"
```

Malix-Labs:
Meanwhile, I also discovered this:
https://code.visualstudio.com/docs/editor/settings-sync#_configuring-synced-data

Malix-Labs:
I couldn't find which to use, so I will first ask the vscode team for clarification
Reverted to draft

package.json
Malix-Labs: Good idea !

Will do it if `ignoreSync` is preferable to the machine scope

Issue #2267, opened 2024-09-26
bug: vscode-neovim is repeatedly locking up in Insert Mode and not responding to ESC or Ctrl-[ to return to Normal Mode


fargly:
### Check the following:

- [X] I have tried setting `vscode-neovim.neovimClean` in VSCode settings and **restarting VSCode.**
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing (closed!) issues.

### Neovim version (nvim -v)

NVIM v0.10.1 (0.10.1+g7e194f0d0)

### Operating system/version

Windows 10

### Problem (including [log messages](https://github.com/vscode-neovim/vscode-neovim?tab=readme-ov-file#troubleshooting))

I'm repeatedly getting stuck in Insert Mode. During a typical dev session (6-8 hours) this occurs 1-3 times, nvim will refuse to leave insert mode. Both ESC and Ctrl-[ fail to work and, if I move to other tabs, they will also be stuck in insert mode. My only recourse is to Save via the menu and run 'Developer: Reload Window' which becomes rather annoying. I always run with 'vscode-neovim.neovimClean' set with no local config changes to nvim (as vanilla as you can get).

I have attached all of the log entries from today. I didn't see anything out of the ordinary, unfortunately. I'll update the issue when I check the logs while experiencing the issue again.
[vscode-neovim-excerpt.log](https://github.com/user-attachments/files/17155674/vscode-neovim-excerpt.log)

My Setup:
Windows 10 [vscode 1.93.1|nvim 0.10.1|vscode-neovim 1.18.12]---REMOTE--->Ubuntu 22.04 Server

### Steps To Reproduce

1. Install/Config Windows 10 [vscode 1.93.1|nvim 0.10.1|vscode-neovim 1.18.12]---REMOTE--->Ubuntu 22.04 Server
2. Edit python file moving between modes
3. Within 3 hours, Stuck in Insert Mode will occur
4. Reload Window, rinse and repeat

### Expected Behavior

nvim will allow normal editing, traversal between modes, and vim motions. Other than this issue, this is what I normally observe and expect. It is only recently that this lockup has started occurring. With documented exceptions, I expect it to behave similar to vanilla nvim in a terminal.

aFreimuthQby:
I am not only getting stuck in insert mode, but in any mode.

Versions:
OS: Debian 12
VS Code: 1.93.1-1726079302 amd64
[...]

PR #2282, opened 2024-10-12
fix: prevent loop on OptionSet

gjf7:
With the config below, entering and exiting Insert mode causes the `OptionSet` event to trigger repeatedly, leading to a loop.

```lua
vim.opt.number = true

vim.api.nvim_create_autocmd("InsertEnter", {
  pattern = "*",
  callback = function()
    vim.wo.relativenumber = true
  end,
})

vim.api.nvim_create_autocmd("InsertLeave", {
  pattern = "*",
  callback = function()
    vim.wo.relativenumber = false
  end,
})
```

Demonstration Videos:

## Before
[Link](https://1drv.ms/v/s!AkmEvT_Y9oQ6gcJBEggsXLzWsZJP_g?e=3l06EI)

The `OptionSet` keeps triggering.
## After
[Link](https://1drv.ms/v/s!AkmEvT_Y9oQ6gcJA16bZRkHnPX7qGQ?e=82s8Vb)

Note: This may not fully address #2267(not sure this is the root cause, in my case, vscode-neovim is not locking up in insert mode), but it does fix the looping OptionSet event in this scenario.

### DIFF
--- a/runtime/lua/vscode/sync-options.lua
+++ b/runtime/lua/vscode/sync-options.lua
@@ -44,45 +44,39 @@ end
 ---Handle changes from vscode
 ---Set nvim options
 ---@param buf number
----@param opts EditorOptions
-local function set_options(buf, opts)
-  -- Due to debounce of editor options change event, in some special cases, buffer
-  -- may be deleted when starting to set options.
+---@param new_opts EditorOptions
+---@param old_opts EditorOptions
+local function set_options(buf, new_opts, old_opts)
   if not api.nvim_buf_is_valid(buf) then
     return
   end
-  api.nvim_buf_set_var(buf, "vscode_editor_options", opts)
-  if vim.bo[buf].ts ~= opts.tabSize then
-    vim.bo[buf].ts = opts.tabSize
-    vim.bo[buf].sw = opts.tabSize
-  end
-  if vim.bo[buf].et ~= opts.insertSpaces then
-    vim.bo[buf].et = opts.insertSpaces
+
+  api.nvim_buf_set_var(buf, "vscode_editor_options", new_opts)
+
+  if new_opts.tabSize ~= old_opts.tabSize then
+    vim.bo[buf].ts = new_opts.tabSize
+    vim.bo[buf].sw = new_opts.tabSize
   end
 
-  local win
-  for _, w in ipairs(api.nvim_list_wins()) do
-    local win_buf = api.nvim_win_get_buf(w)
-    if win_buf == buf then
-      win = w
-      break
-    end
+  if new_opts.insertSpaces ~= old_opts.insertSpaces then
+    vim.bo[buf].et = new_opts.insertSpaces
   end
 
-  if win then
-    set_number(win, opts.lineNumbers)
+  local win = api.nvim_get_current_win()
+  if api.nvim_win_get_buf(win) == buf and new_opts.lineNumbers ~= old_opts.lineNumbers then
+    set_number(win, new_opts.lineNumbers)
   end
 end
 
 ---Check changes from nvim
 ---Set vscode options
-local function _check_options()
-  ---@type EditorOptions?
+local function check_options()
   local opts = vim.b.vscode_editor_options
   if not opts then -- should not happen
     return
   end
-  if not vim.b.vscode_editor_options_first_checked then -- load the defaults
+
+  if not vim.b.vscode_editor_options_first_checked then --load the defaults
     vim.b.vscode_editor_options_first_checked = true
     vim.bo.ts = opts.tabSize
     vim.bo.sw = opts.tabSize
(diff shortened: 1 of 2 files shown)


xiyaowong (changes requested):
Recommend wrapping generic functions to replace `vim.bo` and `vim.wo`.

```lua
-- Wrapper for nvim_set_option_value that sets the option only if the value differs
function M.set_option_value(name, value, opts)
  opts = opts or {}
  local current = api.nvim_get_option_value(name, opts)
  if current ~= value then
    api.nvim_set_option_value(name, value, opts)
  end
end

-- Wrapper for set_option_value to set a buffer option
function M.buf_set_option(buf, name, value)
  return M.set_option_value(name, value, { buf = buf })
end

-- Wrapper for set_option_value to set a window option
function M.win_set_option(win, name, value)
  return M.set_option_value(name, value, { win = win })
end
```

gjf7:
> Recommend wrapping generic functions to replace `vim.bo` and `vim.wo`.
> 
> ```lua
> -- Wrapper for nvim_set_option_value that sets the option only if the value differs
> function M.set_option_value(name, value, opts)
>   opts = opts or {}
>   local current = api.nvim_get_option_value(name, opts)
>   if current ~= value then
>     api.nvim_set_option_value(name, value, opts)
>   end
> end
> 
> -- Wrapper for set_option_value to set a buffer option
> function M.buf_set_option(buf, name, value)
>   return M.set_option_value(name, value, { buf = buf })
> end
> 
> -- Wrapper for set_option_value to set a window option
> function M.win_set_option(win, name, value)
>   return M.set_option_value(name, value, { win = win })
> end
> ```

Great suggestion, done!

theol0403 (approved):
Thank you! @xiyaowong merge when approved

runtime/lua/vscode/sync-options.lua
```diff
@@ -46,68 +65,67 @@ end
...
     return
   end
 
-  local ts, sw, et = vim.bo.ts, vim.bo.sw, vim.bo.et
-  local lineNumbers = get_number_style(vim.wo.nu, vim.wo.rnu)
+  local buf = api.nvim_get_current_buf()
+  local win = api.nvim_get_current_win()
+  local ts = api.nvim_get_option_value("tabstop", { buf = buf })
+  local sw = api.nvim_get_option_value("shiftwidth", { buf = buf })
+  local et = api.nvim_get_option_value("expandtab", { buf = buf })
```
xiyaowong, line 105: This part is still more concise with the original `vim.bo` and `vim.wo`. What do you think?

xiyaowong (changes requested):

runtime/lua/vscode/sync-options.lua
```diff
@@ -46,68 +65,67 @@ end
...
+  if vim.deep_equal(nvim_opts, vscode_opts) then
+    return
   end
+
+  vim.b.vscode_editor_options = nvim_opts
+  vscode.action("set_editor_options", { args = { api.nvim_get_current_buf(), nvim_opts } })
 end
 
-local check_options = util.debounce(_check_options, 20)
+local check_options_debounced = util.debounce(check_options, 20)
```
justinmk, line 128: could avoid some churn by not renaming this. the old name seemed fine

justinmk:
This is worth a try, but was anyone actually able to reproduce the OptionSet loop? 

This does not loop (with or without `nested`):

```lua
local augroup_id = vim.api.nvim_create_augroup('test', { clear = true })
vim.api.nvim_create_autocmd({ 'OptionSet' }, {
  -- nested = true,
  group = augroup_id,
  callback = function()
    vim.wo.relativenumber=not vim.wo.relativenumber
    vim.api.nvim_set_option_value('wrap', not vim.wo.relativenumber, {})
    vim.print(vim.fn.localtime())
  end,
  pattern = { 'tabstop', 'shiftwidth', 'expandtab', 'number', 'relativenumber' },
})

vim.wo.relativenumber=not vim.wo.relativenumber
vim.api.nvim_set_option_value('relativenumber', not vim.wo.relativenumber, {})
```

However, it could loop if there is a `RPC call => OptionSet => RPC call => OptionSet ...` loop somehwere.

runtime/lua/vscode/sync-options.lua
```diff
@@ -15,6 +15,25 @@ local util = require("vscode.util")
...
+
+-- Wrapper for set_option_value to set a buffer option
+function M.buf_set_option(buf, name, value)
+  return M.set_option_value(name, value, { buf = buf })
+end
+
+-- Wrapper for set_option_value to set a window option
+function M.win_set_option(win, name, value)
+  return M.set_option_value(name, value, { win = win })
+end
```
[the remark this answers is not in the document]
gjf7: Fixed

runtime/lua/vscode/sync-options.lua
gjf7: Yeah, original is simpler.

runtime/lua/vscode/sync-options.lua
gjf7: Fixed

xiyaowong:
> This is worth a try, but was anyone actually able to reproduce the OptionSet loop?
> 
> This does not loop (with or without `nested`):


We used debounce, so it wasn't executed directly when OptionSet was triggered, thus it is not affected by being nested.

xiyaowong (approved):

Issue #2234, opened 2024-09-09
Optimize startup error messages


xiyaowong:
- [x] Notify that `prerelease` differs from `release` #2229.
- [ ] Provide additional troubleshooting steps during remote development.
- [ ] For linux(?) mention `useWsl` and link to our WSL setup docs. https://github.com/vscode-neovim/vscode-neovim/issues/2247
    - [ ] also hint about `extensionKind` ?
- [x] Display all detected invalid Nvim versions when no valid version is found.
    - addressed by the `View Logs` button added in [#2287](https://github.com/vscode-neovim/vscode-neovim/pull/2287)
- [x] mention the exact path in the message, to prevent confusion in [remote scenarios](https://github.com/vscode-neovim/vscode-neovim/issues/2251#issuecomment-2353762071)
    - addressed by the `View Logs` button added in [#2287](https://github.com/vscode-neovim/vscode-neovim/pull/2287)

justinmk:
Do you think it's worth adding optional `paths: string[]` and `dirs: string[]` arguments to `findNvim()` ? Then we could use `findNvim()` to handle all cases, including when the user has set an hardcoded path in their settings.

justinmk:
added WSL item to the todo list.

justinmk:
`findNvim` now has `firstMatch` https://github.com/neovim/node-client/issues/370

xiyaowong:
> Do you think it's worth adding optional `paths: string[]` and `dirs: string[]` arguments to `findNvim()` 

I don't think it's needed. Most of the problems are:  
1. Version is too low.  
2. Wrong location when doing remote development.  
3. Permission issues (rare).

It's enough to search in `env:Path`.

justinmk:
But we have a redundant code path here that shows a different message and duplicates the logic of `findNvim`: https://github.com/vscode-neovim/vscode-neovim/blob/ccd89d3e457ee968c09eb13d711bde0e0097a3f9/runtime/vscode-neovim.vim#L24-L55

If `findNvim` accepted extra paths/dirs, then we could delete the above code.

xiyaowong:
> But we allow users to configure arbitrary locations in vscode settings. `findNvim` doesn't handle that (yet).
> 

If the users set the path themselves, then they should be responsible for it.

> And we have a redundant code path here that shows a different message and duplicates the logic of `findNvim`:
> 
> If `findNvim` accepted extra paths/dirs, then we could delete the above code.

How about

[...]

PR #2287, opened 2024-10-15
feat(startup): improve Nvim version check

justinmk:
## Problem:
Version check is spread out across two different parts of the codebase. https://github.com/vscode-neovim/vscode-neovim/issues/2234

## Solution:
- Use the new `paths` feature of `findNvim()` and use that to check the Nvim version, even when the user has configured a specific Nvim path.
- Improve the message wording.
- Show a `View Logs` button.

image

image

### DIFF
--- a/src/main_controller.ts
+++ b/src/main_controller.ts
@@ -106,7 +106,6 @@ export class MainController implements vscode.Disposable {
         this.setClientInfo();
         await this.setCurrentDir();
         await this.client.setVar("vscode_channel", await this.client.channelId);
-        await this.client.setVar("vscode_nvim_min_version", NVIM_MIN_VERSION);
 
         // This is an exception. Should avoid doing this.
         Object.defineProperty(actions, "client", { get: () => this.client, configurable: true });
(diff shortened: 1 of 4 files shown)


src/main_controller.ts
```diff
@@ -185,18 +184,20 @@ export class MainController implements vscode.Disposable {
             }
         }
 
-        let neovimPath = config.neovimPath;
-        // Only try to find nvim if the path is the default one
-        // And if we are not using WSL
-        if (neovimPath === "nvim" && !config.useWsl) {
```
justinmk, line 191: Why skip this for WSL?

src/main_controller.ts
```diff
@@ -185,18 +184,21 @@ export class MainController implements vscode.Disposable {
...
-            const nvimResult = findNvim({ minVersion: NVIM_MIN_VERSION });
-            logger.debug("Find nvim result: ", nvimResult);
-            const matched = nvimResult.matches.find((match) => !match.error);
-            if (!matched) {
-                throw new Error("Unable to find a suitable neovim executable. Please check your neovim installation.");
+        const isDefault = config.neovimPath === "nvim";
+        const nvimResult = findNvim({
+            minVersion: NVIM_MIN_VERSION,
+            paths: isDefault ? undefined : [config.neovimPath],
```
justinmk, line 190: Pass the user-configured path to `findNvim`.

src/main_controller.ts
xiyaowong: The purpose of our useWsl is to use nvim within WSL, which is different from remote development. findNvim is not functioning properly.

justinmk: Ok, I see that this is blocked by https://github.com/neovim/node-client/issues/430

justinmk:
continued in https://github.com/vscode-neovim/vscode-neovim/pull/2679

PR #2290, opened 2024-10-16
fix(startup): null reference in `onDidChangeWindowState` handler

justinmk:
## Problem:
vscode exthost logs may show error, when user changes windows during startup:

    2024-10-16 13:46:16.857 [error] Error: [ExtensionListenerError]
    Extension 'asvetliakov.vscode-neovim' FAILED to handle event:
    TypeError: Cannot read properties of undefined (reading 'command')

## Solution:
- Handle undefined `this.client`.
- TODO: where is `this.client` ever actually defined?

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -12,7 +12,7 @@ function getActionName(action: string) {
 class ActionManager implements Disposable {
     private disposables: Disposable[] = [];
     private actions: string[] = [];
-    private client!: NeovimClient;
+    private client?: NeovimClient;
 
     init() {
         this.initActions();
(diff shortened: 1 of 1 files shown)


src/actions.ts
```diff
@@ -12,7 +12,7 @@ function getActionName(action: string) {
 class ActionManager implements Disposable {
     private disposables: Disposable[] = [];
     private actions: string[] = [];
-    private client!: NeovimClient;
+    private client?: NeovimClient;
```
justinmk, line 15: This is an example of why `!` should rarely be used. It is used too often in this codebase.

P.S. where is the code that actually initializes this `client` ?

xiyaowong: In fact, within the nvim-related logic, the client is always instantiated. However, this is indeed a special case because this event is registered when the Action is instantiated.

xiyaowong (approved):

src/actions.ts
justinmk: > the client is always instantiated

where?

justinmk: Why are we doing that? The comment above it gives no insight.

Issue #2284, opened 2024-10-14
bug: "File exists (add ! to override)" on :write to file in symlink directory


ahelwer:
### Check the following:

- [X] I have tried setting `vscode-neovim.neovimClean` in VSCode settings and **restarting VSCode.**
- [X] I have read the vscode-neovim docs.
- [X] I have searched existing (closed!) issues.

### Neovim version (nvim -v)

NVIM v0.10.2
vscode-neovim 1.18.0
VS Codium 1.94.2

### Operating system/version

Arch Linux

### Problem (including [log messages](https://github.com/vscode-neovim/vscode-neovim?tab=readme-ov-file#troubleshooting))

When executing `:w` to save a file, the following error message will pop up in a notification box in the lower right corner of VS Codium:
```
File exists (add ! to override):
/path/to/file
Source: VSCode Neovim
```
Log:
```
2024-10-14 16:25:35.968 [info] MainController: Starting nvim: /usr/bin/nvim -N --embed --cmd source /home/ahelwer/.vscode-oss/extensions/asvetliakov.vscode-neovim-1.17.2-universal/runtime/vscode-neovim.vim --clean
2024-10-14 16:25:35.992 [info] MainController: Nvim info: {
  nvimVersion: {
    minor: 10,
    build: 'v0.10.2',
    api_level: 12,
    major: 0,
    api_prerelease: false,
    patch: 2,
    api_compatible: 0,
    prerelease: false
  },
  logFile: '/home/ahelwer/.local/state/nvim/log',
  configDir: '/home/ahelwer/.config/nvim'
}
2024-10-14 16:27:55.989 [info] MainController: Starting nvim: /usr/bin/nvim -N --embed --cmd source /home/ahelwer/.vscode-oss/extensions/asvetliakov.vscode-neovim-1.18.0-universal/runtime/vscode-neovim.vim --clean
2024-10-14 16:27:56.030 [info] MainController: Nvim info: {
  configDir: '/home/ahelwer/.config/nvim',
  nvimVersion: {
    prerelease: false,
    minor: 10,
    build: 'v0.10.2',
    patch: 2,
    major: 0,
    api_level: 12,
    api_prerelease: false,
    api_compatible: 0
  },
  logFile: '/home/ahelwer/.local/state/nvim/log'
}
```

### Steps To Reproduce

1. Install neovim (any version) and install vscode-neovim 1.18.0 or newer on Linux
2. Open a project in VS Codium
[...]

PR #2302, opened 2024-10-27
fix(buffers): Fix saving to files inside symlinked workspaces

2qar:
The existing behavior caused a mix of "symbolic" paths and physical paths to be used when saving buffers, which causes incorrect relative path resolution. Take this example structure:

  /home/me/coding/git/vscode-neovim
  /mnt/coding

where "/home/me/coding" is a symbolic link to "/mnt/coding".

The current behavior will mix up "/mnt/coding/git/vscode-neovim/README.md" and "/home/me/coding/git/vscode-neovim/README.md" when attempting to save the README, causing the incorrect destination path to be resolved and throwing an error when attempting to save.

The new behavior uses Neovim's default behavior of resolving "symbolic" paths to physical paths so only "/mnt/coding/git/vscode-neovim/README.md" will be used when determining destination file path.

Note that the test-case I've added doesn't 100% capture the use-case above because of a limitation with the VSCode API ([src](https://vscode-api.js.org/modules/vscode.workspace.html#updateWorkspaceFolders), updating the open workspaces can cause the extension to be terminated).

fix https://github.com/vscode-neovim/vscode-neovim/issues/2284

### DIFF
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -340,6 +340,9 @@ function M.init_document_buffer(data)
   force_filetype()
   -- Set bufname before setting lines so that filetype detection can work ???
   api.nvim_buf_set_name(buf, data.bufname)
+  -- Let nvim resolve the physical path of our file to avoid relative path issues
+  -- with symbolic links when saving the buffer. #2284
+  api.nvim_buf_set_name(buf, api.nvim_buf_get_name(buf))
   api.nvim_buf_set_lines(buf, 0, -1, false, data.lines)
   -- set vscode controlled flag so we can check it neovim
   api.nvim_buf_set_var(buf, "vscode_controlled", true)
--- a/src/test/integ/buffer-write-cmd.test.ts
+++ b/src/test/integ/buffer-write-cmd.test.ts
@@ -1,8 +1,9 @@
 import { strict as assert } from "assert";
 import path from "path";
+import { symlink } from "fs/promises";
 
 import { NeovimClient } from "neovim";
-import { Uri, commands, window, workspace } from "vscode";
+import { TextDocument, Uri, commands, window, workspace } from "vscode";
 
 import {
     attachTestNvimClient,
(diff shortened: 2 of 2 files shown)


xiyaowong (approved):
Wow!

src/test/integ/buffer-write-cmd.test.ts
```diff
@@ -15,6 +16,22 @@ import {
 
 describe("BufWriteCmd integration", () => {
     const testFiles: Uri[] = [];
+    const cleanupCallbacks: (() => Promise<void>)[] = [];
+
+    const cleanupFileCallback = async (uri: Uri) => {
+        try {
+            await workspace.fs.delete(uri);
+        } catch {
+            // ignore
```
justinmk, line 25: is this expected to happen frequently? if not, usually worth at least logging a warn-level message.

src/test/integ/buffer-write-cmd.test.ts
```diff
@@ -15,6 +16,22 @@ import {
 
 describe("BufWriteCmd integration", () => {
     const testFiles: Uri[] = [];
+    const cleanupCallbacks: (() => Promise<void>)[] = [];
+
+    const cleanupFileCallback = async (uri: Uri) => {
```
justinmk, line 21: nit: no reason to mention "callback" in the name.
```suggestion
    const cleanupFile = async (uri: Uri) => {
```

src/test/integ/buffer-write-cmd.test.ts
```diff
@@ -84,4 +127,18 @@ describe("BufWriteCmd integration", () => {
         assert.equal(doc.getText(), "aaa");
         assert.equal(await readFile(doc.uri), "hello world");
     });
+
+    it("Writing to a file in a symbolically linked workspace should succeed", async () => {
```
justinmk, line 131: nit: tests don't need to mention "should succeed" or other redundant phrases. it's assumed that tests are testing correct behavior, except for rare cases where it's not obvious.
```suggestion
    it("Writing to a file in a symlinked workspace", async () => {
```

justinmk (approved):
great solution!

runtime/lua/vscode/internal.lua
```diff
@@ -340,6 +340,9 @@ function M.init_document_buffer(data)
   force_filetype()
   -- Set bufname before setting lines so that filetype detection can work ???
   api.nvim_buf_set_name(buf, data.bufname)
+  -- Let nvim resolve the physical path of our file to avoid relative path issues
+  -- with symbolic links when saving the buffer
```
justinmk, line 344:
```suggestion
  -- with symbolic links when saving the buffer. #2284
```

2qar: :+1: did not realize there was an issue open for this, good catch

src/test/integ/buffer-write-cmd.test.ts
2qar: Ack, updated in the latest revision

src/test/integ/buffer-write-cmd.test.ts
2qar: Removed the redundant wording in the latest change.

theol0403 (approved):

PR #2308, opened 2024-11-01
Autofix: bug: visual mode remains after canceling vscode search

lattaai1:
I have updated the `handleModeChanged` method in the `ModeManager` class to properly handle the transition from visual mode to normal mode when the VSCode search is canceled. This change ensures that the Neovim visual mode is correctly exited when the user cancels the VSCode search. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission


ollien:
I'm reluctant to accept pull requests fully end to end generated by AI.

PR #2337, opened 2024-12-18
Disable escape key to leave insert mode

dlordi:
I find myself hitting Escape multiple times to dismiss an action (ie: search dialog in VSCode): it's bad habit but it assures me that I haven't missed the key press. While I like to use this extension to execute Neovim normal mode commands, I want to be able to explicitly enter normal mode with a dedicated key mapping that cannot be unintentionally used while working in insert mode.
Generally speaking, I like the idea that this extension can be adopted by people who are familiar with VSCode and kinda want to transition gradually into Neovim; so, in the same way it's actually possible to disable Ctrl keys via a config setting, I would suggest to do the same with Escape

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,6 +238,11 @@
                         "w"
                     ]
                 },
+                "vscode-neovim.leaveInsertModeOnEscape": {
+                    "type": "boolean",
+                    "default": true,
+                    "markdownDescription": "Escape key will leave insert mode to enter normal mode"
+                },
                 "vscode-neovim.editorLangIdExclusions": {
                     "type": "array",
                     "uniqueItems": true,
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -238,6 +238,11 @@
                         "w"
                     ]
                 },
+                "vscode-neovim.leaveInsertModeOnEscape": {
```
justinmk, line 241: Not in favor of another random option, especially one that now needs to be checked in various places.

 Is there not a way to achieve this with vscode keyboard shortcut settings? We could then document that.

dlordi: Not sure I can figure out which other places this option has to be checked.
Currently, dismissing a VSCode autocomplete/suggestion menu with Escape would enter normal mode: this doesn't seem right to me.
Trying to tweak VSCode keyboard shortcut settings wasn't successful and, even if possible, my feeling is that it would be too brittle

```json
  {
    "key": "escape",
    "command": "noop",
    "when": "textInputFocus && !inlineSuggestionVisible"
  },
```

theol0403:
Thanks for the PR. Too niche for a setting. Can be instead solved by opening the vscode keybindings editor, searching for the escape binding, right clicking, and clicking remove.

justinmk:
> Can be instead solved by opening the vscode keybindings editor, searching for the escape binding, right clicking, and clicking remove.

Documented here: https://github.com/vscode-neovim/vscode-neovim?tab=readme-ov-file#remove-other-vscode-or-passthrough-keybindings

Issue #2355, opened 2025-01-18
bug: treesitter language injections not working


wurli:
### Check the following:

- [x] I have tried setting `vscode-neovim.neovimClean` in VSCode settings and **restarting VSCode.**
- [x] I have read the vscode-neovim docs.
- [x] I have searched existing (closed!) issues.

### Neovim version (nvim -v)

0.10.2

### Operating system/version

MacOS 114.6.1

### Problem (including [log messages](https://github.com/vscode-neovim/vscode-neovim?tab=readme-ov-file#troubleshooting))

The language for code chunks in markdown files can usually be detected using treesitter. However, running VSCode, `vim.treesitter.get_parser(0, ""):language_for_range({ ... }):lang()` always returns `"markdown"`, whereas in Neovim it can be used to get the language of a given code chunk. This is because, when in VSCode, languages don't seem to be injected into the syntax tree in the same way they are in Neovim.

### Steps To Reproduce

1. Use the following `init.lua` to see the problem:
    <details>
    <summary>Expand for code:</summary>

    ``` lua
    vim.opt.shiftwidth = 4
    vim.opt.tabstop = 4
    vim.opt.softtabstop = 4
    vim.opt.expandtab = true
    vim.g.mapleader = " "
    
    -------------------------
    -- Bootstrap lazy.nvim --
    -------------------------
    local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
    if not (vim.uv or vim.loop).fs_stat(lazypath) then
        local lazyrepo = "https://github.com/folke/lazy.nvim.git"
        local out = vim.fn.system({ "git", "clone", "--filter=blob:none", "--branch=stable", lazyrepo, lazypath })
        if vim.v.shell_error ~= 0 then
            vim.api.nvim_echo({
                { "Failed to clone lazy.nvim:\n", "ErrorMsg" },
                { out, "WarningMsg" },
                { "\nPress any key to exit..." },
            }, true, {})
            vim.fn.getchar()
            os.exit(1)
        end
    end
    vim.opt.rtp:prepend(lazypath)
    
    -----------------------
    -- Set up treesitter --
    -----------------------
    require("lazy").setup({
        spec = {
            "nvim-treesitter/nvim-treesitter",
            config = function()
                require("nvim-treesitter.configs").setup({
                    ensure_installed = { "lua", "markdown" },
                    auto_install = true,
                    highlight = { enable = true }
                })
            end
        }
    })
    
[...]

PR #2356, opened 2025-01-18
fix: treesitter language injections

wurli:
This PR aims to re-enable treesitter language injections, which are currently disabled as a side-effect of turning off highlights using `vim.treesitter.stop()`, as described in #2355.

Simply removing the offending code didn't seem like a good option, since this would enable highlights, which would feed through to VSCode and result in visual differences which users might not appreciate.

I think this is really an upstream issue with Neovim, but I expect this could take some time to fix, so a workaround would be a good idea. A workaround for now would mean allowing the process which resolves and applies treesitter highlights to continue to run, but blocking the part of the process which applies the highlights themselves.

The best way I could find to do this was to shim `highlighter:for_each_highlight_state()` with an empty function, which seems to solve the problem.

Thanks!


theol0403:
Thanks for the PR! Can you check if enabling highlighting but neutering the highlight rendering as you've done also eliminates the need for https://github.com/vscode-neovim/vscode-neovim/blob/master/runtime/vscode/treesitter.lua ?

wurli:
> The original reproduction to the issue that solved was to use treesitter text objects like daf and then undo and observe that the text object doesn't work a second time.

I tried deleting `treesitter.lua` and testing as you suggest, by deleting a function definition using `daf`, undoing, then trying the same operation again. All seem to work as expected! So I think safe to say that `vim.treesitter.stop()` was the culprit :)

runtime/lua/vscode/highlight.lua
```diff
@@ -139,10 +139,17 @@ local function setup_win_hl_ns()
...
+local function disable_treesitter_highlights()
+  for _, buf in ipairs(vim.api.nvim_list_bufs()) do
+    pcall(function()
+      -- Hack to disable treesitter highlighting by monkey patching
+      -- `for_each_highlight_state()`. Could use `vim.treesitter.stop()`, but
+      -- this seems to have unintended side-effects as per
+      -- https://github.com/vscode-neovim/vscode-neovim/issues/2355
+      local highlighter = vim.treesitter.highlighter.active[buf]
+      ---@diagnostic disable-next-line: duplicate-set-field
+      function highlighter:for_each_highlight_state() end
```
xiyaowong, line 151: What if Neovim changes the method name or its functionality in the future?

xiyaowong (approved):
Tested, the textobject functionality works as expected. Neovim has likely already fixed the related potential bugs.

wurli:
Thanks for merging! This has also now been fixed on the nvim-treesitter side (https://github.com/nvim-treesitter/nvim-treesitter/pull/7571#event-15994292104), but I think still worth keeping the fix as `vim.treesitter.stop()` is deprecated and slated for removal anyway.

Innei:
@wurli @xiyaowong Hi, huys. I found maybe this pr will cause treesitter incrmenetal_selection not to work. 

This is my config.

```lua
return {
    {
        "nvim-treesitter/nvim-treesitter",
        vscode = true,
        opts = function(_, opts)
            -- use wildfire.nvim
            opts.incremental_selection = {
                enable = false,
                keymaps = {
                    init_selection = "<CR>",
                    node_incremental = "<CR>",
                    scope_incremental = "<Tab>",
                    node_decremental = "<BS>"
                }
            }

            opts.context_commentstring = {enable_autocmd = false}

            vim.list_extend(opts.ensure_installed,
                            {"prisma", "dockerfile", "svelte", "swift", "vue"})
        end
    },{
        "sustech-data/wildfire.nvim",
        event = "VeryLazy",
        vscode = true,
        dependencies = {"nvim-treesitter/nvim-treesitter"},
        config = function()
            require("wildfire").setup({
                keymaps = {
                    init_selection = "<CR>",
                    node_incremental = "<CR>",
                    node_decremental = "<BS>"
                }
            })
        end
    }
}

```
It works in 1.18.14 but not working in 1.18.15.

The `<CR>` key pressed no response.

xiyaowong:
@Innei Try adding the following code to init.lua:

```lua
local api = vim.api

api.nvim_create_autocmd({ "CursorHold", "InsertLeave" }, {
  group = api.nvim_create_augroup("vscode.treesitter", {}),
  callback = function()
    pcall(function()
      vim.treesitter.get_parser():parse()
    end)
  end,
})
```

xiyaowong:
> > @Innei Try adding the following code to init.lua:
> > ```lua
> > local api = vim.api
> > 
> > api.nvim_create_autocmd({ "CursorHold", "InsertLeave" }, {
> >   group = api.nvim_create_augroup("vscode.treesitter", {}),
> >   callback = function()
> >     pcall(function()
> >       vim.treesitter.get_parser():parse()
> >     end)
> >   end,
> > })
> > ```
> 
> Certainly can, thank you.

I also encountered this problem, and I think this autocommand still needs to be retained.

PR #2390, opened 2025-02-16
fix(defaults): keybindings for ctrl+[BracketLeft], ctrl+[BracketRight] as alternatives for ctrl+[ and ctrl+].

risa2000:
TLDR: This fixes an issue on macOS and non-US keyboard layout (e.g. Czech), where the corresponding Ctrl code is not correctly recognized. (discussed in #1875).

Long version:
When pressing `Ctrl+<key right to P>` on a keyboard with non-US layout (e.g. Czech), which would be `Ctrl+[` on US layout, this key maps to `<C-[>` in `vim` and `neovim` on Windows and macOS.

This was not working in `vscode-neovim` on macOS (but did work in Windows) for some reason which most likely stems from internal VSCode implementation and how VSCode does (or does not) handle the localized keyboard layouts in particular OS. The observed difference is noted [here](https://github.com/microsoft/vscode/issues/237533).

It appears that both VSCode versions (on macOS and Windows) maps the aforementioned keypress to `ctrl+[BracketLeft]` but only Windows version is then able to pass it further to the extension as `ctrl+[`.

This patch adds an explicit keybinding for `ctrl+[BracketLeft]` (and following the same logic for `ctrl+[BracketRight]`) to workaround the VSCode discrepancy in the two OSes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,11 @@ tab.
   this open Terminal and execute the following command:
   `defaults write com.microsoft.VSCode ApplePressAndHoldEnabled -bool false`.
 - To fix the remapped escape key not working in Linux, set `"keyboard.dispatch": "keyCode"`.
+- Two VSCode developer commands are useful for keybindings debugging:
+    - `Developer: Toggle Keyboard Shortcuts Troubleshooting` for tracing VSCode emitted keypresses and their processing
+      via defined keybindings.
+    - `Developer: Inspect Key Mapping` for getting the recognized mappings for the current keyboard layout inside
+      VSCode.
 
 ### Performance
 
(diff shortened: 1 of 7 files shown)


justinmk:
Since you [mentioned](https://github.com/vscode-neovim/vscode-neovim/discussions/1875#discussioncomment-12216888) that `ctrl+[BracketLeft]` should work for most systems, I'm in favor of removing the old `ctrl+[` mappings.

WDYT @xiyaowong @theol0403  ?

justinmk (approved):

risa2000:
@justinmk 
Looking at the CI test for windows-latest it looks like its failing has nothing to do with the patch but with some mismatch of `CRLF` vs `LF` interpretation. Could you confirm this is not an issue, or should I look into it?

PR #2394, opened 2025-02-18
Fix #2368: Exclude editors without view columns to avoid random nvim crashes

### LINKED ISSUE #2368: In relates to "Syncing layout: Syncing active editor" keeps running forever
Adding some more information related to bug #2184 

This always happens when GitHub copilot generates code in editors or it answers chat messages with source code blocks which can be applied to source editor. Either way after applying the generated code, if you move around the editor, suddenly things stops working with the following error messages

Error logs in Extension Host
```
2025-02-01 13:05:42.590 [error] TypeError: invalid msgpack-RPC message: expected array, got: 114
	at /home-path/.vscode/extensions/asvetliakov.vscode-neovim-1.18.15/dist/extension.js:2:162311
	at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Messages in neovim extension logs

```
2025-02-01 13:05:47.331 [warning] BufferManager: Cleanup took longer than 5s:  { windows: [ 1007 ], buffers: [] }
```

I also notice same happens if I work in editor UIs like Settings Editor or Diff Viewer for git, But the crash happening in both those scenarios are seldom.

gayanper:
### Description
This changes avoid starting nvim for editors which lacks a view column, for editors such as 

- Github Copilot code suggestions in chat and edits
- Diff editors

cause random crashes as described in #2368, It seems one common pattern in such editors are lack of a view column. So this change tries to avoid starting nvim for such code editors

### Testing
I have run this patch for a week in my daily driver machine and ensured that with this patch I have not come across any random nvim crashes as before.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -252,7 +252,7 @@ export class BufferManager implements Disposable {
     }
 
     public isExternalTextDocument(textDoc: TextDocument): boolean {
-        if (textDoc.uri.scheme !== "file") {
+        if (textDoc.uri.scheme === "output") {
             return true;
         }
         return this.externalTextDocuments.has(textDoc);
@@ -764,9 +764,10 @@ export class BufferManager implements Disposable {
         const bufname = await this.bufnameForTextDocument(document);
 
         const modifiable = !this.isExternalTextDocument(document);
-        const dirty = this.isExternalTextDocument(document) ? false : document.isDirty;
-        const uri_str = modifiable ? document.uri.toString() : undefined;
-        const uri_data = modifiable ? document.uri.toJSON() : undefined;
+        const isFile = document.uri.scheme === "file";
+        const dirty = isFile ? document.isDirty : false;
+        const uri_str = isFile ? document.uri.toString() : undefined;
+        const uri_data = isFile ? document.uri.toJSON() : undefined;
 
         await actions.lua("init_document_buffer", {
             buf: bufId,
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -252,7 +252,7 @@ export class BufferManager implements Disposable {
     }
 
     public isExternalTextDocument(textDoc: TextDocument): boolean {
-        if (textDoc.uri.scheme === "output") {
+        if (textDoc.uri.scheme !== "file") {
             return true;
         }
         return this.externalTextDocuments.has(textDoc);
@@ -576,7 +576,7 @@ export class BufferManager implements Disposable {
                 this.isLayoutOutdated = false;
                 const token = this.syncLayoutSource?.token;
 
-                const visibleEditors = [...window.visibleTextEditors].filter((e) => e.viewColumn !== undefined);
+                const visibleEditors = [...window.visibleTextEditors];
                 const activeEditor = window.activeTextEditor;
 
                 if (token?.isCancellationRequested) continue;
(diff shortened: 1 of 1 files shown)


xiyaowong:
Some people tried to block this kind of editor before, but it doesn’t really work.  

- `vscode-neovim` works at the editor level, so blocking it would affect everything, including typing.  
- VS Code uses editors in many places, such as the output panel, code blocks, and long text fields. `vscode-neovim` runs fine there and is actually useful.  
- `Editor.ViewColumn` is a layout-related property in VS Code, which we do not care about. It has nothing to do with Neovim issues.

gayanper:
@xiyaowong yes you have a point, I knew that the criteria I used is not directly relevant with the issue, I just found that in my use cases avoid the non editor view type editors will resolve the issues.

But I think what you mentioned by the editor types I think I can explore more on this.

justinmk (approved):
worth a try

src/buffer_manager.ts
```diff
@@ -252,7 +252,7 @@ export class BufferManager implements Disposable {
     }
 
     public isExternalTextDocument(textDoc: TextDocument): boolean {
-        if (textDoc.uri.scheme === "output") {
+        if (textDoc.uri.scheme !== "file") {
```
xiyaowong, line 255: For user data like `settings.json`, the scheme is `vscode-userdata`.
For all files in a remote development environment, the scheme is `remote`.
There are many similar cases. The judgment of `ExternalTextDocument` can only be based on a blacklist.

xiyaowong:
> @xiyaowong yes you have a point, I knew that the criteria I used is not directly relevant with the issue, I just found that in my use cases avoid the non editor view type editors will resolve the issues.

This is probably an illusion. The problem lies in RPC. Blocking these editors reduces certain operations, which in turn reduces the frequency of bugs, but it does not solve the problem.  

Think about it—these are all text editors, so what difference does it make? What if new files are created continuously?

src/buffer_manager.ts
gayanper: I agree, if this experiments work we can improve this to include more schema which are safer.

But at the same time does majority of users use vim feature out side their main editors ?

sf1tzp: Apologies in advance - I am not that familiar with the vscode or vscode-neovim internals. I encountered this kind of crash tonight and have been thinking about your discussion here. 

I wonder, Is it possible to identify these chat-type editors without relying on their scheme?
I typically develop on a remote machine as well, and based on your comment @xiyaowong I assume these chat-type editors use the `remote` scheme as well, making them indistinguishable from a 'normal' remote file.

I do not know if [this vscode ChatViewId](https://github.com/microsoft/vscode/blob/33df589a3a6a2e2ea6bcf016520ae04932b8104a/src/vs/workbench/contrib/chat/browser/chat.ts#L267) is discernible from here.. For instance if this `workbench.panel.chat.view` prefix is accessible on the Buffer or TextDocument, we could filter those editors by that instead of the document scheme?

¯\\_ (ツ)_/¯

Anyways, thank you for opening this issue @gayanper and I'm curious about how it could be resolved.

gayanper:
> > @xiyaowong yes you have a point, I knew that the criteria I used is not directly relevant with the issue, I just found that in my use cases avoid the non editor view type editors will resolve the issues.
> 
> This is probably an illusion. The problem lies in RPC. Blocking these editors reduces certain operations, reducing the frequency of bugs, but it does not solve the problem.
> 
> Think about it—these are all text editors, so what difference does it make? What if new files are created continuously?

So if this is RPC related, does that means if one open many editors with splits we should be able to reproduce the same issue ? because I normally run with 3 columns of editors in my wide screen, but I only get this issue if I use the GitHub chat which output code examples.

PR #2400, opened 2025-02-26
fix: we need to execute treesitter parsing manually

xiyaowong:
https://github.com/vscode-neovim/vscode-neovim/pull/2356#issuecomment-2684120268


justinmk:
Can you provide a short problem/solution explanation? Unraveling the purpose of this later will be even harder than it already is now. And the comment in the file in this PR doesn't help :)

xiyaowong:
@justinmk  Below are the issues and PRs related to this problem.
- #1700
- #1713
- #1717
- #2154

PR #2402, opened 2025-02-28
fix(test): normalize EOL in messages

xiyaowong:


src/test/integ/messages.test.ts
```diff
@@ -24,6 +24,12 @@ function findOutputChannel(): TextEditor | undefined {
     });
 }
 
+function assertOutputContent(expected: string) {
```
justinmk, line 27: Nice change 👍

src/messages_manager.ts
```diff
@@ -160,3 +159,27 @@ export class MessagesManager implements Disposable {
         return msg + "\n";
     }
 }
+
+class MessageBuffer {
```
justinmk, line 163: Defining a new class adds a layer of indirection. This seems a bit over-engineered. `handleFlush()` is where the logic lives, and `MessageBuffer` is essentially an indirect way of moving the logic out of `handleFlush()`. 

Is `MessageBuffer` a useful abstraction? `MessagesManager` itself is already an abstraction doing the dirty work. So moving some parts of it to a similar `MessageBuffer` is indirection without a clear semantic improvement.

xiyaowong:
Yes. This should be unnecessary. I initially thought it was due to changes in Neovim's output, so I first tried modifying the content in the message. Later, I realized it was actually a change in VSCode, where the document in the output uses `\r\n` for line breaks on Windows.

src/messages_manager.ts
xiyaowong: Yes. This should be unnecessary. I initially thought it was due to changes in Neovim's output, so I first tried modifying the content in the message. Later, I realized it was actually a change in VSCode, where the document in the output uses `\r\n` for line breaks on Windows.

theol0403 (approved):

PR #2414, opened 2025-03-14
fix(test): use custom `clipboardCopyAction` implementation

xiyaowong:
Through simple testing, it seems that this action often does not work in tests (VSCode 1.98.0 and later versions).


xiyaowong:
It may take some time for VSCode to fix the issue, and this only affects the test environment. So, I will merge the PR to complete the remaining PRs.

PR #2416, opened 2025-03-14
fix: do not open the file automatically after writing

### LINKED ISSUE #2384: bug: Command `:w name` opens the file after writing it
### Check the following:

### Neovim version (nvim -v)

NVIM v0.10.4 Build type: Release LuaJIT 2.1.1713484068

### Operating system/version

Windows 11 Build 22631

I was trying to save the content of the current buffer to a different file (as a backup) with something like `:w some_name` and while it works at creating a new file, it opens it on vscode right away which is not my intention.

From the readme:

> Since version 1.18.0, :w, :wa and :sav commands are supported and no longer alias to VSCode commands. You can use them as you would in Neovim.

I have v1.18.15 of the extension and I get from the readme that it should behave as it does in neovim which is not opening the new file right away.

If this is a limitation of the extension, feel free to close the ticket.

### Steps To Reproduce

1. Open vscode `code .`
2. Open any file.
3. Attempt to save to a new buffer `:w something`

Observe: New file `something` is created but it is immediately opened in vscode.

### Expected Behavior

Using the command `:w name` should not open the new file immediately.

xiyaowong:
Neovim doesn't auto-open saved files; providing a save confirmation would better meet most users' needs.

- close #2384

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -545,9 +545,12 @@ export class BufferManager implements Disposable {
 
         const text = document.getText();
         const bytes = new TextEncoder().encode(text);
-        await workspace.fs.writeFile(saveUri, bytes);
-        const doc = await workspace.openTextDocument(saveUri);
-        await window.showTextDocument(doc);
+        try {
+            await workspace.fs.writeFile(saveUri, bytes);
+            window.setStatusBarMessage(`Saved ${saveUri.fsPath}`, 3000);
+        } catch (error) {
+            window.showErrorMessage(`Failed to save ${saveUri.fsPath}: ${error}`);
+        }
     }
 
     // #region Sync layout


src/buffer_manager.ts
```diff
@@ -545,9 +545,12 @@ export class BufferManager implements Disposable {
 
         const text = document.getText();
         const bytes = new TextEncoder().encode(text);
-        await workspace.fs.writeFile(saveUri, bytes);
-        const doc = await workspace.openTextDocument(saveUri);
-        await window.showTextDocument(doc);
+        try {
+            await workspace.fs.writeFile(saveUri, bytes);
+            window.setStatusBarMessage(`Saved ${saveUri.fsPath}`, 3000);
```
justinmk, line 550: path can be very long, how about only showing the filename ("tail")?

src/buffer_manager.ts
```diff
@@ -545,9 +545,12 @@ export class BufferManager implements Disposable {
...
         const text = document.getText();
         const bytes = new TextEncoder().encode(text);
-        await workspace.fs.writeFile(saveUri, bytes);
-        const doc = await workspace.openTextDocument(saveUri);
-        await window.showTextDocument(doc);
+        try {
+            await workspace.fs.writeFile(saveUri, bytes);
+            window.setStatusBarMessage(`Saved "${path.basename(saveUri.fsPath)}"`, 3000);
+        } catch (error) {
+            window.showErrorMessage(`Failed to save ${saveUri.fsPath}: ${error}`);
```
justinmk, line 552:
```suggestion
            window.showErrorMessage(`Failed to save "${path.basename(saveUri.fsPath)}": ${error}`);
```

justinmk (approved):

src/buffer_manager.ts
xiyaowong: I think it's better to display the full path in the error message.

PR #2419, opened 2025-03-16
perf: avoid unnecessary sending of editor options to neovim

xiyaowong:
Only send updates when the options we care about change.

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -133,7 +133,14 @@ export class BufferManager implements Disposable {
      */
     private bufferProvider: BufferProvider;
 
+    /**
+     * Used to debounce the editor options change event
+     */
     private editorOptionsChangedTimers: WeakMap<TextEditor, NodeJS.Timeout> = new WeakMap();
+    /**
+     * Avoids unnecessary sending of options to neovim
+     */
+    private editorOptionsMap = new WeakMap<TextEditor, TextEditorOptions>();
 
     /**
      * Buffer event delegate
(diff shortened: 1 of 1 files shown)


src/buffer_manager.ts
```diff
@@ -133,7 +133,14 @@ export class BufferManager implements Disposable {
...
     private bufferProvider: BufferProvider;
 
+    /**
+     * Used to debounce the editor options change event
+     */
     private editorOptionsChangedTimers: WeakMap<TextEditor, NodeJS.Timeout> = new WeakMap();
+    /**
+     * Avoids unnecessary sending of options to neovim
+     */
+    private editorOptionsMap = new WeakMap<TextEditor, TextEditorOptions>();
```
justinmk, line 143: can we check some other existing state (i.e. whatever these settings are actually used for), instead of adding this cache?

src/buffer_manager.ts
```diff
@@ -739,9 +746,22 @@ export class BufferManager implements Disposable {
...
+        const newOptions = { ...editor.options };
+        this.editorOptionsMap.set(editor, newOptions);
+
+        if (
+            oldOptions &&
+            oldOptions.tabSize === newOptions.tabSize &&
+            oldOptions.insertSpaces === newOptions.insertSpaces &&
+            oldOptions.lineNumbers === newOptions.lineNumbers
+        ) {
+            return;
```
justinmk, line 762: Is this avoiding RPC calls, or some other expensive operation? 

That's the kind of thing that really helps explain why a change is made. That's why structuring each PR/commit message as a Problem/Solution pair is really helpful. Otherwise people just have to investigate or assume why a change was made.

xiyaowong: Currently, the main purpose is to ignore changes to the `cursorStyle` option.

src/buffer_manager.ts
xiyaowong: It seems not. The events provided by VSCode only include the new value after the change.

src/buffer_manager.ts
justinmk: > Currently, the main purpose is to ignore changes to the `cursorStyle` option.

how often does that happen? is it worth the extra code?

xiyaowong: The cursor style will change when switching modes.

xiyaowong: Although RPC calls are cheap, there are unnecessary RPC calls and redundant code for processing nvim options. The caching overhead is negligible and the implementation is straightforward. This logic was omitted previously to reduce the codebase when adding the module, but it is now deemed necessary, similar to #2282.

justinmk (approved):

PR #2455, opened 2025-04-18
fix: use the cursor style from VS Code as the style for insert mode

### LINKED ISSUE #2432: feature: Allow changing VS Code cursor for different vim modes
### Did you check the docs?

- [x] I have read all the vscode-neovim docs
- [x] I understand the fundamental limitations of the plugin and believe this request to be reasonable

Before installing this extension my "insert" mode cursor was set to `line-thin`, but after installing, when in "insert" mode, the cursor is set to `line`. I would like my "insert" mode cursor to remain as `line-thin`.

This function in `cursor_manager.ts` has the cursor's hard-coded: https://github.com/vscode-neovim/vscode-neovim/blob/041e540ab30e9dd6474292755ef2152ff8018d01/src/cursor_manager.ts#L177-L196

Adding options in the extension settings for which cursor to use for each mode (and defaulting to the current hard-coded values) would solve this issue. 

I'm not sure there's anything I could do in my nvim settings to accomplish this, as it seems to be a VS Code specific issue.

### Additional context

_No response_

xiyaowong:
- resolve #2432

### DIFF
--- a/src/cursor_manager.ts
+++ b/src/cursor_manager.ts
@@ -40,6 +40,10 @@ export class CursorManager implements Disposable {
      */
     private cursorModes: Map<string, CursorInfo> = new Map();
     /**
+     * Default cursor style for the editor, used as the cursor style in insert mode.
+     */
+    private defaultCursorStyle: TextEditorCursorStyle = TextEditorCursorStyle.Line;
+    /**
      * Cursor positions per editor in neovim
      * ! Note: we should track this because setting cursor as consequence of neovim event will trigger onDidChangeTextEditorSelection with Command kind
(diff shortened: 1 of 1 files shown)


justinmk (approved):

ian-h-chamberlain:
After updating to v.18.20, my cursor seems to be stuck as "block" style even when in insert mode (possibly due to this PR in combination with #2419 ?) I had previously never configured `editor.cursorStyle` so it was should have been using the default `"line"`.

Is there a workaround to force the cursor to a given style, or should I just downgrade to 1.18.19 for now?

PR #2497, opened 2025-06-19
feat: expose module `env` to access `vscode.env`

xiyaowong:


runtime/lua/vscode/api/env.lua
```diff
@@ -0,0 +1,60 @@
...
+---@field sessionId string
+---@field shell string
+---@field uiKind "Desktop"|"Web"
+---@field uriScheme string
+---@field clipboard Clipboard
+local M = {}
+
+M.clipboard = {
+  readText = function()
+    return code.eval("return await vscode.env.clipboard.readText()")
```
justinmk, line 26: This is duplicating nvim's own clipboard api.

Again, duplicating interfaces has these costs:

- confusing for users: the don't know where to look to get features. Since now we expose this part of vscode directly, they'll be less likely to discover other builtin parts of Nvim
- adds cost maintenance cost
- in the case of clipboard, it avoids Nvim's clipboard features/providers, which are expected by plugins and user configs

martin-braun:
I have to second @justinmk here. No doubt this is working, it introduces a lot of technical dept on solving things that aren't really problems. So far I'm the only one asking for access, and if I understand correctly, I don't need this extension to expose this for me, when I can simply do it in my configuration myself.

The implementation exceeds a general mapper, it's explicit, meaning when the VSCode API changes in the future, this project has more responsibilities to take care of adapting to it. I think this responsibility should only fall into those who want to access the API themselves and this exposing of the API is beyond this project's scope.

PR #2520, opened 2025-07-15
fix: change the default statusline to an empty string

xiyaowong:
PR #2518 did not resolve the issue.

### DIFF
--- a/runtime/lua/vscode/default-options.lua
+++ b/runtime/lua/vscode/default-options.lua
@@ -8,6 +8,9 @@ local function setup()
 
   -- When enable `ext_messages`, `cmdheight` will be set to 0 by default.
   vim.opt.cmdheight = 2
+
+  -- Change the default statusline to an empty string
+  vim.opt.statusline = "%##"
 end
 
 return { setup = setup }
--- a/runtime/vscode/statusline.lua
+++ b/runtime/vscode/statusline.lua
@@ -1,19 +1,26 @@
 local api = vim.api
 
 local ns = api.nvim_create_namespace("vscode.statusline")
+
+local function get_status()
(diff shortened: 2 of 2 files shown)


runtime/vscode/statusline.lua
```diff
@@ -3,10 +3,12 @@ local api = vim.api
 local ns = api.nvim_create_namespace("vscode.statusline")
 local curr_status = ""
 
+local DEFAULT_STATUSLINE = api.nvim_get_option_info2("statusline", {}).default
+
 local function refresh()
   local status = ""
 
-  if vim.o.laststatus == 0 or vim.o.statusline == "" then
+  if vim.o.laststatus == 0 or vim.o.statusline == "" or vim.o.statusline == DEFAULT_STATUSLINE then
```
justinmk, line 11: isn't there an early init area where vscode-neovim sets default options? then the user can override that if they want, or augment it with `+=`.

justinmk: somehow the latest push is even more complicated.

my suggestion was simply to set a default 'statusline' *early*, wherever vscode-neovim sets its defaults. we don't need to do backflips to override anything that happens after that.

runtime/lua/vscode/default-options.lua
```diff
@@ -8,6 +8,9 @@ local function setup()
 
   -- When enable `ext_messages`, `cmdheight` will be set to 0 by default.
   vim.opt.cmdheight = 2
+
+  -- Change the default statusline to an empty string
+  vim.opt.statusline = "%##"
```
justinmk, line 13: Seems like we could choose a better default than just "empty", but ok. Not a blocker...

PR #2531, opened 2025-08-11
fix(cmdline): handle async input hide race condition

### LINKED ISSUE #2510: bug: Search & replace with confirm (:%s/foo/bar/gc) not working properly
### Check the following:

### Neovim version (nvim -v)

v0.11.2 and v0.10.4

### Operating system/version

Ubuntu 24.04.2 LTS

Search & replace with confirmation (`:%s/foo/bar/gc`) doesn't work properly for me using either neovim v0.11.2 or v0.10.4 as backend (both run using appimages). But the failure mode is slightly different.

When using v0.11.2, the confirmation entry, shown in a dialog just like the command line, just disappears after a while (as if the answer given was `q`, regardless of the actual input), seemingly when the window needs to scroll to show the next match.

When using v0.10.4, the confirmation entry is shown in the status bar instead, and it doesn't quit after the first input, *but* the UI has issues. Sometimes the term is highlighted, sometimes you just get a cursor at the start of the match, and sometimes that cursor even stays on a previous match, making it very hard to see what replacement you're actually deciding on. Example of this is shown below, where the `barbar` on line 165 has just been replaced (from `foobar`), and the `foobar` on line 166 is the actual replacement currently in question. Not at all reflected by the cursor, and no highlight in sight.

Image

### Steps To Reproduce

1. Create a new file with a repeated pattern of two lines saying one thing, followed by many lines saying another (enough to not have two groups of the first pattern on the screen at once). This pattern should be repeated at least 4 times. E.g.
```
foobar
foobar
bar
bar
...
bar
bar
foobar
foobar
bar
bar
...
```
2. Issue the command `:%s/foo/bar/gc`.
3. Answer `y` a few times.
4. See it all fall apart.

### Expected Behavior

`:%s/foo/bar/gc` being just as usable in vscode-neovim as natively in neovim (or vim).

xiyaowong:
- fix #2510

- [x] Add test

### DIFF
--- a/src/cmdline_manager.ts
+++ b/src/cmdline_manager.ts
@@ -2,7 +2,7 @@ import { Disposable, QuickInputButton, QuickPick, QuickPickItem, ThemeIcon, comm
 
 import { EventBusData, eventBus } from "./eventBus";
 import { MainController } from "./main_controller";
-import { disposeAll } from "./utils";
+import { disposeAll, ManualPromise } from "./utils";
 import { createLogger } from "./logger";
 import { calculateInputAfterTextChange } from "./cmdline/cmdline_text";
 import { GlyphChars } from "./constants";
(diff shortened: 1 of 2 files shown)


justinmk (approved):

xiyaowong:
It suddenly clicked—I finally figured out why the tests kept failing.

PR #2539, opened 2025-09-01
add: word wrap files, pending maintainer direction

YouSame2:
## Overview
extension didnt function well with vscode when wordWrap was enabled. i spent a lot of time configuring my own neovim keymaps to work in vscode one-to-one to how neovim handles keymaps with wordWrap on. I believe I have it as best as it can get. 

looking over the contributing.md file, figured that this "Delegates as much as possible to nvim" and could potentially be a PR to this project rather than a plugin/wiki in itself. Hoping you agree.

This first draft is to get you opinion on where the files and code should go. as of now i just copied over my neovim files into runtime/vscode/word_wrap pending your review.

## My proposal
looking over the codebase i believe the best place to put this code is as follows:

utils.lua -> runtime/lua/vscode/util.lua (not sure if this is the right place for it though)
keymaps-vscode.lua & togglewrap.lua -> combine into 1 file into runtime/vscode/word_wrap.lua or replace existing motions override vim file.

let me know what you think and i can go from there.

## Features
ToggleWrap Command: Adds a :ToggleWrap user command to enable, disable, or toggle word wrap globally.
Keymap Integration: ~~Sets up <leader>uw to toggle wrap and adjust movement keymaps for wrapped lines.~~ just an example of how to use it. wont be included in final draft.
Macro Safety: Automatically disables wrap and related keymaps during macro recording to avoid conflicts.
VSCode Sync: Keeps VSCode’s editor.wordWrap setting in sync with Neovim’s wrap state.

### DIFF
--- a/runtime/vscode/word_wrap/keymaps-vscode.lua
+++ b/runtime/vscode/word_wrap/keymaps-vscode.lua
@@ -1,0 +1,302 @@
+--[[
+NOTES ABOUT KEYMAPS... it gets confusing
+  -- ignore "o" in j/k that gets complex usually u want to 3yj according to relative lines not screen lines
+  -- check vim.fn.mode() to avoid issues with visual block mode
+  -- we can add v:count == 0 ? check to keep 3j function the same as without wrap? need to ask manintainer about this. for now i just include it in my personal remapping of j.
+  -- is_macro_executing to fix keymap issues when executing macros. plays back keys as if theres no word wrap.
+  -- i learned after re-writing these keymaps 20 times that for proper functionality in macros we need to use expr in mappings
+  -- i prefer to remap I/A to be wrap aware. but these could be changed to something else to prevent user confusion.
+  -- same with D/C/Y
+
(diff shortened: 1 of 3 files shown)


runtime/vscode/word_wrap/keymaps-vscode.lua
```diff
@@ -0,0 +1,302 @@
...
+
+local utils = require("vscode.word_wrap.utils")
+local vscode = require("vscode")
+
+M = {
+  {
+    "n",
+    "gj",
+    function()
+      if utils.is_macro_executing() then
```
justinmk, line 24: No, this is a hack. Mapping over and handling individual keys is just not going to be maintainable.

YouSame2:
> It could be a standalone plugin rather than built into vscode-neovim, making updates and maintenance easier.

thats my next step. should have it out shortly. i just figured that behavior of default keys that are builtin to neovim should function out of the box for everyone without the need for a plugin.

PR #2560, opened 2025-09-27
feat: Add statusLineItems config option

aritsune:
Problem:
User can't disable display of msg_show, msg_showcmd and msg_showmode events.

Solution:
Adds a statusLineItems config option that lets the user choose which events (msg_show, msg_showcmd, msg_showmode, and statusline) are rendered, and in which order.

Wrote this tweak to satisfy my customization needs. Should work as before by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -334,6 +334,21 @@
                     "default": "|",
                     "description": "Separator used by statusline"
                 },
+                "vscode-neovim.statusLineItems": {
+                    "type": "array",
+                    "uniqueItems": true,
+                    "items": {
+                        "type": "string",
+                        "pattern": "(statusline|mode|cmd|msg)"
+                    },
+                    "default": [
+                        "statusline",
+                        "mode",
+                        "cmd",
+                        "msg"
+                    ],
+                    "description": "Order of status line items displayed"
+                },
                 "vscode-neovim.compositeTimeout": {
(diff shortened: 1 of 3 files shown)


aritsune:
far as I can tell what's referred to as "the statusline" in this extension doesn't just print the neovim `statusline` but also `msg_show`, `msg_showcmd` and `msg_showmode` events. I wanted to have just the neovim statusline and this seemed like the least disruptive way to have it do that in my own setup

justinmk:
That wasn't clear from your PR description. Always describe the actual problem you are solving. In this case:

```
Problem:
User can't disable display of msg_show, msg_showcmd and msg_showmode events.
```

package.json
```diff
@@ -334,6 +334,21 @@
...
+                        "type": "string",
+                        "pattern": "(statusline|mode|cmd|msg)"
+                    },
+                    "default": [
+                        "statusline",
+                        "mode",
+                        "cmd",
+                        "msg"
+                    ],
+                    "description": "Order of status line items displayed"
```
justinmk, line 350: vscode calls it the "statusbar" 

```suggestion
                    "description": "Control the info displayed in the statusbar"
```

justinmk:
Waiting for comments from other maintainers.

PR #2515, opened 2025-07-13
fix: improve message handling

xiyaowong:
fix https://github.com/vscode-neovim/vscode-neovim/issues/2507

**Key points:**

1. Some common and simple messages are shown only in the status line.
2. The `message_manager` and `statusline_manager` are merged into a single `message_manager`, making event handling clearer, since the status line was already handling part of the message events.
3. Messages shown through `msg_show` are accumulated — in other words, the output panel shows some message history. This is necessary because we need to handle the `replace_last` and `append` options, which require storing past messages. In contrast, `msg_history_show` directly replaces the content.
4. Once the user has viewed the messages and closed the output panel, the messages are considered expired, and the output is cleared immediately.

**TODO:**

* [ ] Require Neovim 0.12+
* [ ] Handle `cmdheight` option properly
* [ ] Add integration tests

Issue #2507, opened 2025-07-05
bug: searching and other actions auto-open the Output messages view


vdawg-git:
### Check the following:

- [x] I have tried setting `vscode-neovim.neovimClean` in VSCode settings and **restarting VSCode.**
- [x] I have read the vscode-neovim docs.
- [x] I have searched existing (closed!) issues.

### Neovim version (nvim -v)

NVIM v0.12.0-dev-736+gd9465e984b

### Operating system/version

Arch

### Problem (including [log messages](https://github.com/vscode-neovim/vscode-neovim?tab=readme-ov-file#troubleshooting))

When I search for something with /, and go through the results, it will open the messages output panel of Neovim in VSCode, spamming me with the output.
This is also happening sometimes when deleted lines etc.

### Steps To Reproduce

1. Type `/`
2. Press `n` multiple times
3. See the output tab pop up

### Expected Behavior

Output does not pop up.

justinmk:
What version of `nvim` are you using? This is shown in the vscode-neovim logs.

Show a screenshot. This only happens if Nvim outputs multiple lines, which of course requires a multiline way to display those messages. With `nvim --clean` this does not happen. https://github.com/vscode-neovim/vscode-neovim/issues/2099#issuecomment-2169085647

xiyaowong:
This is the same as the current reason for the CI failure — there is an issue with the current handling of the `msg_clear` event.

xiyaowong:
No time to fix it. Just temporarily set `cmdheight` to a larger value.

xiyaowong:
I think we can display some less important messages (such as searchcmd and search_count) only in the status bar, without putting them in the output panel. Other messages should always be shown in the output panel and cleared after they expire using a timer—after all, important messages are usually saved in the history, so this won't cause much impact. On another note, we’re still using cmdheight to determine whether to open the output panel actively. For now, there doesn’t seem to be a better solution.

xiyaowong:
It is recommended to use nvim 0.11.3. PR #2515 will only be merged after nvim 0.12 is released, as it introduces breaking changes.

vegerot:
@xiyaowong Thank!  For people running Neovim HEAD and VSCode Insiders, would you please release #2515 as a [Pre-Release version](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions)?

asweet-confluent:
[...]

PR #2587, opened 2025-11-26
feat: disable messages on output

fish895623:
image
image

i know this is important option, but when i using vscode-neovim, keep focus to OUTPUT tab. so i make this to disable it.


justinmk:
this will be fixed by https://github.com/vscode-neovim/vscode-neovim/pull/2515 

workarounds are mentioned in https://github.com/vscode-neovim/vscode-neovim/issues/2507

Issue #2606, opened 2026-02-23
bug: buffer sync sporadic failure


OldStarchy:
### Check the following:

- [x] I have tried setting `vscode-neovim.neovimClean` in VSCode settings and **restarting VSCode.**
- [x] I have read the vscode-neovim docs.
- [x] I have searched existing (closed!) issues.

### Neovim version (nvim -v)

NVIM v0.12.0-dev-1459+g0791594101

### VSCode version and flavor (Cursor, vscodium, …)

VSCode Version: 1.109.5 (user setup)

### Operating system/version

Windows 11

### Problem (including [log messages](https://github.com/vscode-neovim/vscode-neovim?tab=readme-ov-file#troubleshooting))

Buffer sync for a file starts failing at unpredictable times.

* All neovim driven operations (`dd`, `p` etc.) are not synced back to vscode
* Changes in vscode (insert mode) are not synced back to neovim
* Restarting neovim `Neovim: Restart Extension` does not resolve the issue
* Only happens in one editor
* Closing the editor and reopening it resolves it.

Evidence that neovim is out of sync is as follows

* Once you notice its out of sync (eg. `dd` doesn't delete a line)
* `ggVGd` to remove all lines
* vscode has all content, but cannot move cursor past L0:0 in normal mode
* you can `p` into the buffer and it doesn't show in vscode
* `ggVGy` and `p` into a different editor window puts the content of the out-of-sync buffer into the other editor

### Steps To Reproduce

I have not yet been able to identify what triggers the desync, but it happens frequently.

### Expected Behavior

No desync.

justinmk:
Does it happen if you disable other vscode extensions (and perhaps also Nvim plugins)?

Ninja260:
@OldStarchy `Neovim: Stop Extension` and `Neovim: Restart Extension` in case you don't want to close the IDE. 🎉. I have given shortcuts for these commands, as crushing out happen quite often sometimes.

theol0403:
I've noticed some desyncs over the years, but unfortunately without exact replication steps we can't really address it. The restart extension command is good for when it happens.

OldStarchy:
I have the "Restart Extension" already bound to `ctrl+k, ctrl+n` and I use it frequently, but as I mentioned before it doesn't fix this issue. Closing and reopening just the editor tab (not vscode) does.

[...]

PR #2610, opened 2026-04-11
fix: drain queued document changes after buffer init

zhoufanscut:
## Summary

Fixes a race condition where document changes fired during async buffer initialization were silently orphaned, causing the neovim buffer to diverge from VS Code's content.

I first hit this in the **VS Code search editor**. Repro:

1. In a normal editor, visual-select a keyword.
2. Trigger "Open Search Editor" to search for that keyword (the selection is pre-populated into the search editor's query area).
3. Repeat a few times.

After a few iterations, the search editor's cursor lands in the wrong position and subsequent keystrokes behave as if neovim's buffer state doesn't match what's on screen. It's intermittent because it depends on whether VS Code fires its content-population change *during* the async buffer init.

Debugging with an AI assistant, we traced it to this sequence:

1. `BufferManager.syncVisibleEditors` calls `await initBufferForDocument(doc, buf, editor)` — an async flow.
2. While that's running, VS Code fires `onDidChangeTextDocument` for the search editor as its content is populated.
3. `DocumentChangeManager.onChangeTextDocument` pushes the change into `documentChangeQueue`, then early-returns at `if (!this.documentContentInNeovim.has(doc)) return;` because `onBufferInit` hasn't fired yet.
4. When `onBufferInit` eventually fires, it sets `documentContentInNeovim` but never drains the queue. The queued changes are lost, and neovim's view of the buffer is now stale.

Any document type that fires changes during async init can hit this — the search editor is just a reliable reproducer.

## Fix

Three small changes in `src/buffer_manager.ts` and `src/document_change_manager.ts`:

1. **Set `textDocumentToBufferId` before `initBufferForDocument`.** Otherwise, when the drain runs, `processTextDocumentChange` → `getBufferIdForTextDocument(doc)` returns `undefined` and silently drops the change, defeating the drain.
2. **Drain queued changes in `onBufferInit`** after `documentContentInNeovim` is set, under `documentChangeLock` so the drain serializes correctly against any subsequent `onChangeTextDocument` calls.
3. **Make `onBufferInit` async and `await` it at both call sites** (`initBufferForDocument` and the external-buffer path). This ensures the drain completes before `buffer.listen("lines", ...)` subscribes, so neovim's own line events can't interleave with the drain.

The type signature of `onBufferInit` now returns `void | Promise<void>` to reflect this.

## Test plan

- [x] Reproducer: visual-select a keyword, open search editor to search for it, repeat — cursor and edits behave correctly after the fix
- [x] Regular file editing is unaffected (no queued changes → drain is a no-op)
- [x] External buffer path (`handleExtBufferOpen` → `onBufferInit`) still works
- [x] `npm run lint:check`, `npm run format:check`, `npm run test-compile` all pass via pre-commit hook

### DIFF
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -159,7 +159,12 @@ export class BufferManager implements Disposable {
         more: boolean,
     ) => void;
 
-    public onBufferInit?: (bufId: number, doc: TextDocument, initDocText: string, initDocVersion: number) => void;
+    public onBufferInit?: (
+        bufId: number,
+        doc: TextDocument,
+        initDocText: string,
+        initDocVersion: number,
+    ) => void | Promise<void>;
 
     private get client() {
         return this.main.client;
@@ -717,10 +722,9 @@ export class BufferManager implements Disposable {
                     logger.error(`Cannot create a buffer, code: ${buf}`);
                     continue;
                 }
-                await this.initBufferForDocument(doc, buf, editor);
-
-                logger.log(doc.uri, LogLevel.Debug, `Document: ${doc.uri}, BufId: ${buf.id}`);
                 this.textDocumentToBufferId.set(doc, buf.id);
+                logger.log(doc.uri, LogLevel.Debug, `Document: ${doc.uri}, BufId: ${buf.id}`);
+                await this.initBufferForDocument(doc, buf, editor);
             }
             if (this.textEditorToWinId.has(editor)) continue;
             const editorBufferId = this.textDocumentToBufferId.get(doc)!;
@@ -849,7 +853,7 @@ export class BufferManager implements Disposable {
         if (!this.isExternalTextDocument(document)) {
             await actions.lua("clear_undo", bufId);
         }
-        this.onBufferInit?.(bufId, document, text, version);
+        await this.onBufferInit?.(bufId, document, text, version);
         buffer.listen("lines", this.receivedBufferEvent);
         actions.fireNvimEvent("document_buffer_init", bufId);
     }
@@ -917,7 +921,7 @@ export class BufferManager implements Disposable {
 
         this.externalTextDocuments.add(doc);
         this.textDocumentToBufferId.set(doc, id);
-        this.onBufferInit?.(id, doc, doc.getText(), doc.version);
+        await this.onBufferInit?.(id, doc, doc.getText(), doc.version);
 
         const windows = await this.client.windows;
         let closeWinId = 0;
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -164,7 +164,7 @@ export class BufferManager implements Disposable {
         doc: TextDocument,
         initDocText: string,
         initDocVersion: number,
-    ) => void | Promise<void>;
+    ) => Promise<void>;
 
     private get client() {
         return this.main.client;
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -373,9 +373,19 @@ export class BufferManager implements Disposable {
                 const buffers = await this.client.buffers;
                 const buf = buffers.find((b) => b.id === bufnr);
                 if (buf) {
-                    await this.initBufferForDocument(doc, buf);
+                    // Set the mapping before init so that queued onChangeTextDocument events
+                    // drained in DocumentChangeManager.onBufferInit can resolve the buffer id.
+                    // Roll back on failure so the document isn't stuck in a half-initialized state.
+                    this.textDocumentToBufferId.set(doc, bufnr);
+                    try {
+                        await this.initBufferForDocument(doc, buf);
+                    } catch (e) {
+                        this.textDocumentToBufferId.delete(doc);
+                        throw e;
+                    }
+                } else {
+                    this.textDocumentToBufferId.set(doc, bufnr);
                 }
-                this.textDocumentToBufferId.set(doc, bufnr);
             }
             if (window.activeTextEditor?.document !== doc) {
                 const editor = await window.showTextDocument(doc, {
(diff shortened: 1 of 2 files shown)


src/buffer_manager.ts
```diff
@@ -712,10 +717,9 @@ export class BufferManager implements Disposable {
                     logger.error(`Cannot create a buffer, code: ${buf}`);
                     continue;
                 }
-                await this.initBufferForDocument(doc, buf, editor);
-
-                logger.log(doc.uri, LogLevel.Debug, `Document: ${doc.uri}, BufId: ${buf.id}`);
                 this.textDocumentToBufferId.set(doc, buf.id);
+                logger.log(doc.uri, LogLevel.Debug, `Document: ${doc.uri}, BufId: ${buf.id}`);
+                await this.initBufferForDocument(doc, buf, editor);
```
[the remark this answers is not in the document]
zhoufanscut: Good catch — addressed in 351ff78. `initBufferForDocument` is now wrapped in try/catch; on failure we delete the `textDocumentToBufferId` mapping and `continue`, so a future `syncVisibleEditors` can retry init instead of treating the doc as already initialized.

src/buffer_manager.ts
```diff
@@ -844,7 +848,7 @@ export class BufferManager implements Disposable {
         if (!this.isExternalTextDocument(document)) {
             await actions.lua("clear_undo", bufId);
         }
-        this.onBufferInit?.(bufId, document, text, version);
+        await this.onBufferInit?.(bufId, document, text, version);
```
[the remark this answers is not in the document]
zhoufanscut: Good catch — addressed in 351ff78. In `handleExternalBuffer` the mapping is now set *before* `initBufferForDocument`, so queued `onChangeTextDocument` events drained in `DocumentChangeManager.onBufferInit` can resolve the buffer id via `getBufferIdForTextDocument`. On failure the mapping is rolled back and the error is rethrown to the existing outer try/catch.

src/buffer_manager.ts
```diff
@@ -159,7 +159,12 @@ export class BufferManager implements Disposable {
         more: boolean,
     ) => void;
 
-    public onBufferInit?: (bufId: number, doc: TextDocument, initDocText: string, initDocVersion: number) => void;
+    public onBufferInit?: (
+        bufId: number,
+        doc: TextDocument,
+        initDocText: string,
+        initDocVersion: number,
+    ) => void | Promise<void>;
```
justinmk: why return both `void | Promise`, instead of just always a Promise? is the non-Promise variant required somewhere ?

justinmk (approved):
conceptually this sounds like a good change. will feel better if @xiyaowong can review/approve

xiyaowong:
Does issue #2606 relate to this PR?

src/buffer_manager.ts
zhoufanscut: Good point, there's only one implementation and it's async now. the void | half was defensive

zhoufanscut:
> Does issue #2606 relate to this PR?

before I created this PR, claude found #2606 and said:
This is highly likely the same bug — "sporadic buffer sync failure", "happens at unpredictable times", "only one editor affected". The symptoms match a
  race condition during buffer initialization perfectly. The user even notes that reopening the editor fixes it (fresh init, no race)."

but I was not very sure, so I didnt link it
Happy to link it as Relates to #2606 in the description if you think that's appropriate

justinmk:
Still waiting on at least one unaddressed comment. Can you rebase and address it?

src/buffer_manager.ts
zhoufanscut: Done in 357d22b — single impl and it's `async`, so the `void` half was unreachable. Narrowed it to `Promise<void>`. Also rebased onto latest master.

justinmk:
i haven't looked at this yet, but just for reference here's a claude review (it found many issues):


<details>
<summary>review details </summary>

> Publishing the doc→buffer mapping before init closes the orphaned-change window but opens
> several new ones.
> 
> ### Correctness
> 
> 1. **`document_change_manager.ts:188`** — the drain replays changes that predate the init
>    snapshot. A dirty-state-only event (`contentChanges === []`) skips the version guard and
>    rewinds `documentContentInNeovim` to pre-init text, so later changes compute byte offsets
>    against the wrong lines and `nvim_buf_set_text` corrupts the buffer. It also clears
>    `modified` on a buffer just marked dirty.
> 2. **`document_change_manager.ts:190`** — the queue is emptied before processing, so a throw
>    mid-loop discards the remaining changes: the exact bug this PR fixes. Shift from the live
>    queue, or restore the tail in a `finally`.
> 3. **`buffer_manager.ts:876`** — awaiting `onBufferInit` makes `buffer.listen("lines")`
>    conditional on the drain succeeding, so one RPC error permanently disables sync for that
>    buffer. `attachNeovimExternalBuffer` has no try/catch, so it also skips the cursor fix-up.
> 4. **`buffer_manager.ts:386`** — the new `else` sets the mapping without ever initializing, so
>    `has(doc)` skips init forever: a permanent no-sync document with an unbounded queue.
> 
> ### Races from publishing the mapping early
> 
> 5. **`:739`** — a concurrent `syncEditorLayout` sees `has(doc)`, skips init, and calls
>    `createNeovimWindow` / `nvim_win_set_cursor` on an empty, nameless buffer.
> 6. **`:684`** — `cleanupWindowsAndBuffers` can now `nvim_buf_delete` the buffer mid-init.
> 
> ### Behaviour
> 
> 7. **`:874`** — `clear_undo` runs before the drain, so replayed changes become undoable: in a
>    search editor `u` reverts the auto-populated query.
> 8. **`document_change_manager.ts:193`** — no `isExternalTextDocument` guard, so provider-echo
>    changes get written back into read-only `:help` buffers, defeating `modifiable=false`.
> 
> ### Design / performance
> 
> 9. **`:850`** — this replays a stale snapshot rather than removing the staleness; `modified` is
>    even read several awaits after `text`/`version`. Re-reading both immediately before
>    `init_document_buffer` shrinks the gap to one RPC and removes most of the need to replay —
>    and with it several findings above.
> 10. **`document_change_manager.ts:191`** — the drain holds the global lock across N sequential
>     round-trips, blocking typing and cursor sync. Batch into one `handle_changes`.
> 
> ### Robustness / nits
> 
> 11. **`:743`** — rollback drops the mapping but leaks the nvim buffer; cleanup only iterates
>     that map.
> 12. **`:745`** — a fatal layout error becomes a silent `continue`, logging `.message` without
>     the stack.
> 13. Duplicated `set()` in both if/else arms (`:386`); the same rationale written three times
>     (`:376`, `:735`, `document_change_manager.ts:183`); and `document_change_manager.ts:186`
>     describes the old behaviour rather than what the code does.

</details>

PR #2616, opened 2026-05-02
Handle Neovim buffer sync compatibility

tmwatchanan:
## Summary

- Fall back when Neovim rejects the `BufModifiedSet` autocmd event by watching text/write events and forwarding modified-state changes.
- Recover from `Vim:E95: Buffer with this name already exists` during document buffer initialization by deleting only a stale vscode-controlled duplicate for the same document and retrying the buffer name assignment.

## Root cause

Newer Neovim development builds can reject the `BufModifiedSet` autocmd event expected by the extension runtime. Separately, layout sync can create a new document buffer before stale cleanup has removed the previous vscode-controlled buffer for the same document, which makes `nvim_buf_set_name` fail with `E95`.

### DIFF
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -13,23 +13,21 @@ local function get_buf_var(buf, name)
   end
 end
 
+---Finds a hidden stale buffer for the same VS Code document that blocks naming during init.
 local function find_duplicate_document_buffer(buf, name, uri)
   for _, candidate in ipairs(api.nvim_list_bufs()) do
     if candidate ~= buf and api.nvim_buf_is_valid(candidate) then
       local candidate_name = api.nvim_buf_get_name(candidate)
       local candidate_uri = get_buf_var(candidate, "vscode_uri")
-      local candidate_controlled = get_buf_var(candidate, "vscode_controlled")
 
-      local name_matches = candidate_name == name or candidate_uri == uri
-      local is_vscode_document = candidate_controlled == true or candidate_uri == uri
-
-      if name_matches and is_vscode_document then
+      if candidate_name == name and candidate_uri == uri and vim.tbl_isempty(fn.win_findbuf(candidate)) then
         return candidate
       end
     end
   end
 end
 
+---Retries E95 buffer naming after removing a stale same-document duplicate.
 local function set_document_buffer_name(buf, name, uri)
   local ok, err = pcall(api.nvim_buf_set_name, buf, name)
   if ok then
@@ -39,10 +37,12 @@ local function set_document_buffer_name(buf, name, uri)
   if type(err) == "string" and err:find("E95:", 1, true) then
     local duplicate = find_duplicate_document_buffer(buf, name, uri)
     if duplicate then
-      pcall(api.nvim_buf_delete, duplicate, { force = true })
-      ok, err = pcall(api.nvim_buf_set_name, buf, name)
-      if ok then
-        return
+      local delete_ok = pcall(api.nvim_buf_delete, duplicate, { force = true })
+      if delete_ok then
+        ok, err = pcall(api.nvim_buf_set_name, buf, name)
+        if ok then
+          return
+        end
       end
     end
   end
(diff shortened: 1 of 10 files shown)


src/test/integ/actions.test.ts
```diff
@@ -67,7 +67,7 @@ describe("Eval VSCode", () => {
 
         await assert.rejects(
             () => eval_from_nvim(client, "const a = {}; a.a = a; return a"),
-            /Error executing lua Return value of eval not JSON serializable: TypeError: Converting circular structure to JSON.*/,
+            /Return value of eval not JSON serializable: TypeError: Converting circular structure to JSON/,
         );
```
justinmk, line 71: Nvim messages changed so this change is needed 👍

src/buffer_manager.ts
```diff
@@ -546,6 +546,11 @@ export class BufferManager implements Disposable {
         const currentPath = normalize(current_name);
         const targetPath = normalize(target_name);
 
+        // :w !cmd — filter/shell write; not a filesystem path. Avoid saveAs/save dialogs that hang tests/CI.
+        if (targetPath.startsWith("!")) {
+            return;
```
justinmk, line 551: this sure looks like a hack...

runtime/lua/vscode/internal.lua
```diff
@@ -6,6 +6,50 @@ local util = require("vscode.util")
...
 local M = {}
 
+local function get_buf_var(buf, name)
+  local ok, value = pcall(api.nvim_buf_get_var, buf, name)
+  if ok then
+    return value
+  end
+end
+
+local function find_duplicate_document_buffer(buf, name, uri)
```
justinmk, line 17: needs a concise docstring that gives insight into why it's needed, what is the key idea(s) of this function.

(NOT a long, verbose docstring explaining the implementation.)

runtime/lua/vscode/internal.lua
```diff
@@ -6,6 +6,50 @@ local util = require("vscode.util")
...
+      local is_vscode_document = candidate_controlled == true or candidate_uri == uri
+
+      if name_matches and is_vscode_document then
+        return candidate
+      end
+    end
+  end
+end
+
+local function set_document_buffer_name(buf, name, uri)
```
justinmk, line 31: needs a concise docstring that gives insight into why it's needed, what is the key idea(s) of this function.

(NOT a long, verbose docstring explaining the implementation.)

runtime/lua/vscode/internal.lua
```diff
@@ -6,6 +6,50 @@ local util = require("vscode.util")
...
+    return value
+  end
+end
+
+local function find_duplicate_document_buffer(buf, name, uri)
+  for _, candidate in ipairs(api.nvim_list_bufs()) do
+    if candidate ~= buf and api.nvim_buf_is_valid(candidate) then
+      local candidate_name = api.nvim_buf_get_name(candidate)
+      local candidate_uri = get_buf_var(candidate, "vscode_uri")
+      local candidate_controlled = get_buf_var(candidate, "vscode_controlled")
```
justinmk: this is also a hack, but apparently we already depend on this var...

justinmk (approved):
Looks like a much needed improvement, thanks

runtime/lua/vscode/internal.lua
```diff
@@ -302,13 +354,27 @@ local function set_buffer_autocmd(buf)
       vscode.action("save_buffer", { args = { data } })
     end,
   })
-  api.nvim_create_autocmd({ "BufModifiedSet" }, {
+  local ok, create_err = pcall(api.nvim_create_autocmd, { "BufModifiedSet" }, {
```
justinmk, line 357: Note: Nvim 0.12 removed `BufModifiedSet`, and we will need to use `OptionSet` with `pattern=modified` instead. That can be a followup PR though.

justinmk:
Thanks! This looks like a reasonable improvement, I'm hoping you will keep an eye for further improvements.

PR #2618, opened 2026-06-01
Fix failed integration tests

tmwatchanan:
## Summary

Fixes the failing `BufWriteCmd` integration tests seen in CI by handling `:write !cmd` as an external shell write rather than a VS Code file save, and by making the test runner disable VS Code’s built-in Git extensions explicitly.

## Changes

- Skip `save_buffer` handling for `BufWriteCmd` targets that start with `!`.
- Remove `--disable-extensions` from integration-test launch args so `--disable-extension=vscode.git` and `--disable-extension=vscode.git-base` are honored.
- Update the `:write !cmd` integration test to drive the interactive Ex command through Neovim input and dismiss the hit-enter prompt intentionally.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,6 @@
             "smartStep": true,
             "runtimeExecutable": "${execPath}",
             "args": [
-                "--disable-extensions",
                 "--disable-extension=vscode.git",
                 "--disable-extension=vscode.git-base",
                 "--extensionDevelopmentPath=${workspaceFolder}",
--- a/runtime/lua/vscode/internal.lua
+++ b/runtime/lua/vscode/internal.lua
@@ -344,6 +344,9 @@ local function set_buffer_autocmd(buf)
     callback = function(ev)
       local current_name = api.nvim_buf_get_name(ev.buf)
       local target_name = ev.match
+      if vim.startswith(target_name, "!") then
+        return
+      end
 
       local data = {
         buf = ev.buf,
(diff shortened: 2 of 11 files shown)


.github/workflows/build_test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             - name: Setup Node
               uses: actions/setup-node@v6
               with:
-                  node-version: "18"
+                  node-version-file: package.json
```
justinmk, line 24: oh, that's a nice approach. https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

would be nice to put this in a separate PR so we can merge it quickly

justinmk:
how about focusing on a few changes at a time? that will make reviews faster

.github/workflows/build_test.yml
tmwatchanan: Sure, I'll create a separate PR for this

xiyaowong:
Looks good overall. But do we really need to test nightly? The latest stable is 0.12.5, yet our tests already cover 0.13. Are we supposed to keep up with every nightly change?

justinmk:
> Are we supposed to keep up with every nightly change?

they generally shouldn't be disruptive, and it gives early warning...

xiyaowong (approved):
Let's merge first and iterate later.

justinmk:
I have a PR on the way which actually fixes the bugs instead of whatever this is doing.

PR #2626, opened 2026-07-21
fix: scope list fold keybindings to listFocus

wenfangdu:
## What

Append `&& listFocus` to the `when` clause of the list fold/expand keybindings so they only fire when a list widget is focused, rather than in any non-editor, non-input context.

## Changes

- `package.json` — 8 keybindings updated
- `scripts/keybindings/5_widgets.cjs` — 8 keybindings updated (source generator)
- `.gitignore` — ignore the `.claude/` directory

Affected keys: `z o`, `z shift+o`, `z c`, `z shift+c`, `z a`, `z shift+a`, `z m`, `z shift+m`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 Session.vim
 *.vsix
 *.zip
+.claude/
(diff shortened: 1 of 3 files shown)


wenfangdu:
@justinmk Please review, should be safe to merge.

PR #2636, opened 2026-09-07
fix(visual): wrong selections on nvim 0.13

justinmk:
Problem:
Linewise visual selections and (VScode) multicursor insert-points are
off-by-one on Nvim 0.13: `V` on a 14-char line yields (0,0)-(1,0)
instead of (0,0)-(0,14), and `V`+`A` inserts at the start of the next
line.

Analysis:
Nvim 0.13 `vim.lsp.util.make_given_range_params` was [reimplemented](https://github.com/neovim/neovim/commit/ff43f1950e1fd3b5ae37924bc8f9c6c3e9d54908) on
top of `vim.range.mark():to_lsp()`. The old version blindly did
`B[2] = B[2] + 1` to convert Nvim's inclusive end to LSP's exclusive end,
which VSCode clamped back to the line-end (thus it happened to work).

Solution:
- Convert byte columns to UTF-16, stopping the exclusive end at EOL.
- Drop the deprecated `make_given_range_params`, `make_position_params` and
  legacy `str_utfindex` calls.

Tests:
- Poll for the help buffer instead of waiting a fixed 1s.
- `Ctrl-u`: wait for the typed text to reach Nvim before sending it. Still
  fails on 0.13, but deterministically: Nvim has the text and does not
  treat it as entered in the current insert session.

### DIFF
--- a/runtime/lua/vscode/api.lua
+++ b/runtime/lua/vscode/api.lua
@@ -1,6 +1,8 @@
 local api = vim.api
 local fn = vim.fn
 
+local util = require("vscode.util")
+
 local M = {}
 
 ------------------------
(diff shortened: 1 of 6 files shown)


justinmk:
This reduces the failing tests from 6 => 2

PR #2672, opened 2026-09-11
fix: "Right-hand side of 'instanceof' is not an object"

### LINKED ISSUE #2671: Can't start neovim: `Error: Right-hand side of 'instanceof' is not an object`
### Check the following:

### Neovim version (nvim -v)

0.12.5

### VSCode version and flavor (Cursor, vscodium, …)

1.137.0

### Operating system/version

Windows 11 25H2

Can't start neovim with the message:

> Failed to set current dir: Error: Right-hand side of 'instanceof' is not an object

log:

```log
2026-09-10 00:08:21.749 [info] MainController: Starting nvim: C:/Dev/BasicTools/neovim/bin/nvim.exe -N --embed --cmd execute 'source' fnameescape('c:/Users/spindensity/.vscode/extensions/asvetliakov.vscode-neovim-1.19.2/runtime/vscode-neovim.vim') --clean
2026-09-10 00:08:21.749 [error] MainController: Failed to set current dir: Error: Right-hand side of 'instanceof' is not an object
```

vscode-neovim 1.19.1 is OK, only 1.19.2 has this problem.

### Steps To Reproduce

1. Set `vscode-neovim.neovimClean` to `true`;
2. Install vscode-neovim 1.19.2 from vsix;
2. Restart vscode.

### Expected Behavior

Neovim starts successfully..

justinmk:
Problem:
Request to Nvim fails in 1.19.2:

    Failed to set current dir: Error: Right-hand side of 'instanceof' is not an object

node-client's `api/Buffer` and `api/types` require each other, and
`types.js` collects the ext-type classes at load time. Entering the package
through `neovim/lib/api/Buffer` (for `ATTACH`) leaves `Buffer` undefined
there, so the msgpack encoder's `instanceof` throws. The production
(webpack) bundle evaluates that import first; CI only tests the
development bundle.

Solution:
Drop the import. `buf[ATTACH](true)` was a no-op anyway: the `listen()`
before it already attaches the buffer, so `ATTACH` returned without sending
`nvim_buf_attach`.

fix https://github.com/vscode-neovim/vscode-neovim/issues/2671

I also have a fix for node-client itself, which would remove the need for this PR, but this PR is still a worthwhile cleanup.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -1,5 +1,5 @@
 import { NeovimClient } from "neovim";
-import { VimValue } from "neovim/lib/types/VimValue";
+import type { VimValue } from "neovim/lib/types/VimValue";
 import { ConfigurationTarget, Disposable, Range, commands, window, workspace } from "vscode";
 
 import { eval_for_client } from "./actions_eval";
--- a/src/buffer_manager.ts
+++ b/src/buffer_manager.ts
@@ -2,7 +2,6 @@ import path from "path";
 
 import { debounce } from "lodash";
 import { Buffer, NeovimClient } from "neovim";
-import { ATTACH } from "neovim/lib/api/Buffer";
 import {
     CancellationToken,
     CancellationTokenSource,
@@ -1017,7 +1016,6 @@ class BufferProvider implements TextDocumentContentProvider {
         }
 
         buf.listen("lines", this.receivedBufferEvent);
-        await buf[ATTACH](true);
 
         return lines.join("\n");
     }


xiyaowong:
Confirmed, the issue is fixed. Just wondering, how did you debug it? Or did you already have an idea of where the problem was?

I also tried digging into `node-client`, but I’m not very familiar with it, so I didn’t really have a clue where to start.

xiyaowong (approved):

justinmk:
> Just wondering, how did you debug it?

AI 😂

